PR #134, opened 2020-08-19
Rearrange content in main readme

### LINKED ISSUE #126: Readme adjustments
- [ ] Split Quickstart in Buidler & Truffle subsections

frangio:
Fixes #126

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,35 +7,49 @@
 - Manage proxy admin rights.
 - Easily use in tests.
 
-## Quickstart
+## Installation
 
-Install the plugin for your favorite toolchain:
+### Buidler
+
 ```
-npm install --save-dev @openzeppelin/buidler-upgrades
+npm install --save-dev @openzeppelin/buidler-upgrades @nomiclabs/buidler-ethers ethers
 ```
+
+This installs our Buidler plugin along with the necessary peer dependencies.
+
+### Truffle
+
 ```
 npm install --save-dev @openzeppelin/truffle-upgrades
 ```
 
-And set up a [Buidler script](https://buidler.dev/guides/scripts.html) to deploy and upgrade, after adding `usePlugin('@openzeppelin/buidler-upgrades')` to your `buidler.config.js`:
+## Usage
+
+See the documentation for each plugin, or take a look at the sample code snippets below.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,35 +7,49 @@
...
 
-## Quickstart
+## Installation
+
+### Buidler
 
-Install the plugin for your favorite toolchain:
 ```
-npm install --save-dev @openzeppelin/buidler-upgrades
+npm install --save-dev @openzeppelin/buidler-upgrades @nomiclabs/buidler-ethers ethers
```
martriay, line 15: Rather than assuming that the user is already using Buidler, we should assume they *want* to use Buidler. Me suggestion then would be to include it in the installation command, same for Truffle.

martriay: This is to guarantee successful tryouts, so I'd reconsider the `installation` title above in favor of something self contained, although I don't have strong feelings about it.

README.md
```diff
@@ -7,35 +7,49 @@
...
 
-And set up a [Buidler script](https://buidler.dev/guides/scripts.html) to deploy and upgrade, after adding `usePlugin('@openzeppelin/buidler-upgrades')` to your `buidler.config.js`:
+## Usage
+
+See the documentation for each plugin, or take a look at the sample code snippets below.
+
+| [<img src="assets/buidler.svg" height="20px" width="30px" alt="">Buidler](./packages/plugin-buidler/README.md)| [<img src="assets/truffle.svg" height="20px" width="30px" alt="">Truffle](./packages/plugin-truffle/README.md) |
+|-|-|
+
+Buidler users will be able to write [scripts](https://buidler.dev/guides/scripts.html) that use the plugin to deploy or upgrade a contract.
```
martriay, line 33: Although it's not the idea to showcase the whole API, I think we're rendering the `admin` module invisible. What do you think of using at least one its functions?

README.md
frangio: > Rather than assuming that the user is already using Buidler, we should assume they want to use Buidler

I disagree with this. I see this section as "I'm using Buidler, I want to use this, what should I do".

The criteria about "guaranteeing successful tryouts" is interesting though. Let's keep the discussion going and keep iterating on this section.

README.md
frangio: I extended this sentence to mention admin, but didn't add it to the snippet to keep it smaller. The admin module is documented in detail in each plugin.

PR #142, opened 2020-08-26
Initial docs

### LINKED ISSUE #141: Prepare content for publishing on docs site
Should mostly be based on the existing content we have in README files.

Structure:

- Overview
- Using with Truffle
- Using with Buidler
- Network Files
- FAQ
- API Reference
  - Truffle
  - Buidler

We should also include the general upgradeabilty content that is currently found in Upgrades, and some from CLI.

- [Writing Upgradeable Contracts](https://docs.openzeppelin.com/upgrades/2.8/writing-upgradeable)
- [Proxies](https://docs.openzeppelin.com/upgrades/2.8/proxies)
- [Why Am I Getting the Error "Cannot Call Fallback Function From the Proxy Admin"?](https://docs.openzeppelin.com/cli/2.8/faq#why-am-i-getting-the-error-cannot-call-fallback-function-from-the-proxy-admin)

There is also [Creating Upgradeable Contracts From Solidity](https://docs.openzeppelin.com/upgrades/2.8/creating-upgradeable-from-solidity) but I think that should go in Contracts docs once we have a factory contract.

abcoathup:
Initial docs based on READMEs and CLI and Lib docs.

Reference to `import "@openzeppelin/upgrades/contracts/Initializable.sol";` will need to be updated once https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2335 merged and released.

`network-files.adoc` may need reworking.

----

Resolves #141


docs/modules/ROOT/pages/network-files.adoc
```diff
@@ -0,0 +1,64 @@
...
+    }
+  },
+  "admin": {
+    "address": "...",
+    "txHash": "..."
+  }
+}
+----
+
+For every logic contract, besides the deployment address, the following info is also tracked:
```
frangio, line 43: I think we use "implementation" and "logic" kind of inconsistently. I also don't think either of those terms is good.

Marto has suggested we could use the term "version", and talk about "contract versions". In this sentence we would say "For every contract version, ...".

What do you think?

docs/modules/ROOT/pages/network-files.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+* `types` keeps track of all the types used in the contract or its ancestors, from basic types like `uint256` to custom `struct` types
+* `storage` tracks the storage layout of the linearized contract, referencing the types defined in the `types` section, and is used for verifying that any storage layout changes between subsequent versions are compatible
```
frangio, line 46: Here we could have a link to the FAQ entry about storage compatibility.

docs/modules/ROOT/pages/network-files.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+The naming of the file will be `<network_name>.json`, but note that `<network_name>` is not taken from the name of the network's entry in the Truffle or Buidler configuration file, but is instead inferred from the canonical network id associated to the entry.
```
frangio, line 48: Technically we use the chain id, which is not necessarily the same as the network id.

I also don't think "canonical" means anything in this context so I removed it, but if I was mistaken let me know.

```suggestion
The naming of the file will be `<network_name>.json`, but note that `<network_name>` is not taken from the name of the network's entry in the Truffle or Buidler configuration file, but is instead inferred from the chain id associated to the entry. 
```

docs/modules/ROOT/pages/network-files.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+* `types` keeps track of all the types used in the contract or its ancestors, from basic types like `uint256` to custom `struct` types
+* `storage` tracks the storage layout of the linearized contract, referencing the types defined in the `types` section, and is used for verifying that any storage layout changes between subsequent versions are compatible
+
+The naming of the file will be `<network_name>.json`, but note that `<network_name>` is not taken from the name of the network's entry in the Truffle or Buidler configuration file, but is instead inferred from the canonical network id associated to the entry. 
+
+[[configuration-files-in-version-control]]
+== Configuration Files in Version Control
+
+Public network files like `mainnet.json` or `ropsten.json` should be tracked in version control. These contain valuable information about your project's status in the corresponding network; the addresses of the contract implementations that have been deployed, the addresses of the proxies that have been deployed, etc. Such files should be identical for all the contributors of a project.
```
frangio, line 53: Proxy addresses are not contained in these files in the plugins.

```suggestion
Public network files like `mainnet.json` or `ropsten.json` should be tracked in version control. These contain valuable information about your project's status in the corresponding network, like the addresses of the contract implementations that have been deployed. Such files should be identical for all the contributors of a project.
```

docs/modules/ROOT/pages/network-files.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+However, local network files like `unknown-<network_id>.json` only represent a project's deployment in a temporary local network such as `ganache-cli` that are only relevant to a single contributor of the project and should not be tracked in version control.
```
frangio, line 55:
```suggestion
However, local network files like `unknown-<chain_id>.json` only represent a project's deployment in a temporary local network such as `ganache-cli` that are only relevant to a single contributor of the project and should not be tracked in version control.
```

frangio: This is generally true but the "unknown" files could also correspond to other public chains that we're unaware of, like the Ethereum Classic chain. I think we should explain that.

docs/modules/ROOT/pages/writing-upgradeable.adoc
```diff
@@ -0,0 +1,394 @@
...
+    function __ERC20_init_unchained(string memory name, string memory symbol) internal initializer {
+        _name = name;
+        _symbol = symbol;
+        _decimals = 18;
+    }
+  ...
+}
+----
+
+Whether using OpenZeppelin Contracts or another Ethereum Package, always make sure that the package is set up to handle upgradeable contracts.
```
frangio, line 158:
```suggestion
Whether using OpenZeppelin Contracts or another smart contract library, always make sure that the package is set up to handle upgradeable contracts.
```

docs/modules/ROOT/pages/writing-upgradeable.adoc
```diff
@@ -0,0 +1,394 @@
...
+contract MyContract is Initializable {
+    uint256 public hasInitialValue;
+
+    function initialize() public initializer {
+        hasInitialValue = 42;
+    }
+}
+----
+
+Note that it still is fine to set constants here, because the compiler https://solidity.readthedocs.io/en/latest/contracts.html#constant-state-variables[does not reserve a storage slot for these variables], and every occurrence is replaced by the respective constant expression. So the following still works with OpenZeppelin Upgrades:
```
frangio, line 185:
```suggestion
Note that it is still fine to define _constant_ state variables in this way, because the compiler https://solidity.readthedocs.io/en/latest/contracts.html#constant-state-variables[does not reserve a storage slot for these variables], and every occurrence is replaced by the respective constant expression. So the following still works with OpenZeppelin Upgrades:
```

docs/modules/ROOT/pages/writing-upgradeable.adoc
```diff
@@ -0,0 +1,394 @@
...
+contract MyContract is Initializable {
+    ERC20 public token;
+
+    function initialize() public initializer {
+        token = new ERC20("Test", "TST"); // This contract will not be upgradeable
+    }
+}
+----
+
+The easiest way around this issue is to simply accept an instance of that contract as a parameter, and inject it after creating it:
```
frangio, line 219:
```suggestion
If you would like the `ERC20` instance to be upgradeable, the easiest way to achieve that is to simply accept an instance of that contract as a parameter, and inject it after creating it:
```

docs/modules/ROOT/pages/writing-upgradeable.adoc
```diff
@@ -0,0 +1,394 @@
...
+
+There is, however, an exception. If the direct call to the logic contract triggers a `selfdestruct` operation, then the logic contract will be destroyed, and all your contract instances will end up delegating all calls to an address without any code. This would effectively break all contract instances in your project.
+
+A similar effect can be achieved if the logic contract contains a `delegatecall` operation. If the contract can be made to `delegatecall` into a malicious contract that contains a `selfdestruct`, then the calling contract will be destroyed.
+
+As such, it is strongly recommended to avoid any usage of either `selfdestruct` or `delegatecall` in your contracts. If you need to include them, make absolutely sure they cannot be called by an attacker on an uninitialized logic contract.
```
frangio, line 248: I think the CLI allowed force-overriding this but we don't.

```suggestion
As such, it is not allowed to use either `selfdestruct` or `delegatecall` in your contracts.
```

abcoathup:
> Can you delete the API docs from the readmes, and also the FAQ, so we avoid that duplication?

I removed API & FAQ.  
Also removed How do the plugins work? (as this would have lots of links to docs FAQ) and Managing Ownership to avoid duplication.

docs/modules/ROOT/pages/writing-upgradeable.adoc
abcoathup: Resolved in: https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/142/commits/2111130ed7681ce04989a5b15a87ed4173511454

docs/modules/ROOT/pages/network-files.adoc
abcoathup: Added explanation for limited set of public chains
https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/142/commits/3c56793eafe3fd699e4cb6629af098df824f6684

docs/modules/ROOT/pages/network-files.adoc
abcoathup: Sounds good. Updated in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/142/commits/a05349dca998d49951c5adc19c20863e5d059223

docs/modules/ROOT/pages/network-files.adoc
abcoathup: Added link: https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/142/commits/c00466b4ce08e97f8149af712e1d409165297958

README.md
```diff
@@ -75,123 +75,6 @@ it('works before and after upgrading', async function () {
 });
 ```
 
-## Usage
-
-
-## How do the plugins work?
```
frangio, line 81: I think we want to keep some content like this on the readme. The links can go to the FAQ on the docs site. But feel free to also just remove the links for now until we know what the actual links will be when we publish.

abcoathup: I have added How to the plugins work and Managements back in but without links to the FAQ.

abcoathup:
@frangio what will the name of the docs URL be?  Is it https://docs.openzeppelin.com/upgradesplugins or something else?

README.md
abcoathup: If we know what the URL will be, I can add the links to the API and to the FAQ.

frangio:
According to the current config the URL will be `docs.openzeppelin.com/upgrades-plugins`.

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/8f142a4827472e12caddbb5562f2bffc86165d0e/docs/antora.yml#L1

README.md
abcoathup: I added links to the FAQ, and for each plugin a link to the API and the main docs.

PR #143, opened 2020-08-27
Add manifest migration script

### LINKED ISSUE #15: Implement migration from CLI manifests to the new format
I started building the manifest format from scratch because it was easier for me, ~but we have to decide whether we'll reuse the format that exists in OpenZeppelin CLI~. It turns out that the existing CLI manifest contains a lot of information that is irrelevant for the plugins so we will define a new one. Thus we have to add the `manifestVersion` field (3.0?) and implementing automatic migration from older to newer manifest version for users who decide to migrate to the plugins.

martriay:
Fixes #15


packages/core/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,243 @@
...
+interface MigrationOutput {
+  migrations: Record<string, ManifestData>;
+  migrationData: unknown;
+}
+
+interface AddressWrapper {
+  address?: string;
+}
+
+interface LegacyManifest {
```
frangio, line 170: Feel free to simplify this type by removing any properties we don't use, or using `unknown` in places where we don't use the value. Same for all the other types.

packages/core/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,243 @@
+import path from 'path';
+import { promises as fs } from 'fs';
+import type { ManifestData, ImplDeployment } from '../manifest';
+import type { StorageItem, StorageLayout, TypeItem } from '../storage';
+
+const OPEN_ZEPPELIN_FOLDER = '.openzeppelin';
+const MIGRATION_FILE_LOCATION = 'openzeppelin-cli-migration.json';
+
+export async function migrateLegacyProject(): Promise<void> {
```
frangio, line 9: This is missing the deletion of those files that are no longer needed, for example all of the `dev-*` that are renamed to `unknown-*`, although we might want to delete them entirely as I mentioned in another comment.

And there's also the `project.json` which we might want to delete or do something about. This idea goes beyond the requirements that we had initially defined so maybe we want to explore it in a second iteration, but we might want to export the compiler version and options somehow, maybe offering the user to fill in or create their Truffle or Buidler config.

packages/plugin-buidler/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node
+import { migrateLegacyProject } from '@openzeppelin/upgrades-core';
+
+migrateLegacyProject();
```
frangio, line 4: This is an async call so we need to catch a potential rejection, otherwise node fails with `UnhandledPromiseRejection`.

```suggestion
migrateLegacyProject().catch(e => {
  console.error(e);
  process.exit(1);
});
```

packages/plugin-truffle/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node
+import { migrateLegacyProject } from '@openzeppelin/upgrades-core';
+
+migrateLegacyProject();
```
frangio, line 4:
```suggestion
migrateLegacyProject().catch(e => {
  console.error(e);
  process.exit(1);
});
```

packages/core/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,243 @@
...
+async function migrateManifestFiles(manifestList: string[]): Promise<MigrationOutput> {
+  const migrationData: Record<string, unknown> = {};
+  const migrations: Record<string, ManifestData> = {};
+
+  for (const manifestFile of manifestList) {
+    const oldManifest = JSON.parse(await fs.readFile(manifestFile, 'utf8'));
+    const { manifestVersion } = oldManifest;
+
+    if (manifestVersion === undefined) {
+      continue;
```
frangio, line 30: This looks like it should be an error. If there is no `manifestVersion` it's a really old CLI that didn't version its manifest files. We should fail if that's the case.

packages/core/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    const [major] = manifestVersion.split('.');
+
+    if (major !== '2') {
+      continue;
```
frangio, line 36: Same here, we should validate that the manifest we're migrating is the latest version, otherwise we should fail, the user should use the CLI to obtain an updated manifest, and then we can work with that.

packages/core/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,243 @@
...
+  const network = getNetworkName(fileName);
+  return isTestnet(network) || isDevnet(network);
+}
+
+function getNetworkName(fileName: string): string {
+  return path.basename(fileName, '.json');
+}
+
+function isDevnet(network: string) {
+  return network.includes('dev');
```
frangio, line 69: This is too permissive IMO. We should use a regular expression
```suggestion
  return /^dev-\d+$/.test(network);
```

packages/core/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,243 @@
...
+    },
+  };
+}
+
+function transformImplementations(contracts: LegacyContracts): Record<string, ImplDeployment> {
+  const impls: Record<string, ImplDeployment> = {};
+  for (const contractName in contracts) {
+    const contract = contracts[contractName];
+    if (contract.deployedBytecodeHash === undefined) {
+      continue;
```
frangio, line 119: This happens when it's an abstract contract, right? Can you add a comment explaining that? It took me a little while to remember.

packages/core/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,243 @@
...
+    } else {
+      impls[contract.deployedBytecodeHash] = transformImplementationItem(contract);
+    }
+  }
+  return impls;
+}
+
+function transformImplementationItem(contract: LegacyContract): ImplDeployment {
+  return {
+    address: contract.address || 'NO_ADDRESS_FOUND',
```
frangio, line 129: Shouldn't there always be an address?

packages/core/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,243 @@
...
+    storage: contract.storage?.map(transformStorageItem) || [],
+    types: transformTypes(contract.types || {}),
+  };
+}
+
+function transformStorageItem(storageItem: LegacyStorageItem): StorageItem {
+  return {
+    contract: storageItem.contract,
+    label: storageItem.label,
+    type: storageItem.type,
```
frangio, line 145: Have you looked into the differences between the CLI `typeIdentifier` processing and ours? If not this is just a reminder. I am pretty sure we do very different stuff there and I think it's worth stopping everything and taking a good look at how we can handle that.

packages/core/src/scripts/migrate-legacy-project.ts
martriay: The field is optional in the migrated type

packages/core/src/manifest.ts
```diff
@@ -8,7 +8,7 @@ import { compare as compareVersions } from 'compare-versions';
 import type { Deployment } from './deployment';
 import { StorageLayout } from './storage';
 
-const manifestVersion = '3.0';
+const manifestVersion = '3.1';
```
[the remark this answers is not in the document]
martriay: Mh maybe but why would you run this script if you're already using the plugins?

packages/core/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,243 @@
+import path from 'path';
+import { promises as fs } from 'fs';
+import type { ManifestData, ImplDeployment } from '../manifest';
+import type { StorageItem, StorageLayout, TypeItem } from '../storage';
+
+const OPEN_ZEPPELIN_FOLDER = '.openzeppelin';
+const MIGRATION_FILE_LOCATION = 'openzeppelin-cli-migration.json';
```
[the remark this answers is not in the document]
martriay: No reason, I haven't properly addressed that file (content and name) yet.

packages/core/src/manifest.ts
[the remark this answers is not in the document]
frangio: This is in `manifest.ts`!

packages/core/src/scripts/migrate-legacy-project.ts
martriay: This is a copypaste of the original types in the SDK.

packages/core/src/scripts/migrate-legacy-project.ts
martriay: Yes that part is yet to be done. Remember this is a draft.

packages/core/src/scripts/migrate-legacy-project.ts
martriay: I thought of this behavior to skip already migrated files in the same spirit of idempotence of the plugins.

packages/core/src/scripts/migrate-legacy-project.ts
frangio: I'd personally follow YAGNI in this case, but it's not a strong opinion.

packages/core/src/scripts/migrate-legacy-project.ts
frangio: In that case we should throw an error if it's not defined. It would be good to know in what scenarios the field is not present. But we need the address otherwise the plugins won't work.

packages/core/src/scripts/migrate-legacy-project.ts
martriay: I did and it does different stuff, this is another issue yet to be solved.

packages/core/src/scripts/migrate-legacy-project.ts
frangio: We should be careful with that, not everything has to be idempotent. Idempotence is important for the deployment and impl store modules, and it can have interesting properties for the user, but it's not a goal in general.

That said, I think it's a good idea but we should only ignore files with `manifestVersion >= 3.0`.

packages/core/src/scripts/migrate-legacy-project.ts
```diff
@@ -0,0 +1,251 @@
...
+
+    const [major, minor] = oldManifest.manifestVersion.split('.');
+
+    if (major === '3') {
+      continue;
+    }
+
+    // 2.2 latest version
+    if (major !== '2' || minor !== '2') {
+      throw new Error(`Migration failed: expected manifest version 2.2, got ${major}.${minor} instead`);
```
martriay, line 40: Maybe we could point them somewhere to update it, otherwise this is a deadend.

packages/core/src/deployment.ts
```diff
@@ -17,7 +17,7 @@ export async function resumeOrDeploy<T extends Deployment>(
 ): Promise<T> {
   // If there is a deployment stored, we look its transaction up. If the
   // transaction is found, the deployment is reused.
-  if (cached !== undefined) {
+  if (cached?.txHash !== undefined) {
```
frangio, line 20: The result of this is that if `txHash` is undefined the contract is redeployed. But if `txHash` is undefined it's because the contract was migrated from a CLI project, and we can probably reuse it.

The body of this `if` will check that the transaction has been mined. We can't do that without `txHash`. We can probably assume that the tx has been mined, or we could try to check that there is code at the address.

packages/core/src/scripts/migrate-oz-sdk-project.ts
```diff
@@ -0,0 +1,351 @@
...
+  const valueType = transformTypeName(getValueType(typeName));
+  return `t_array(${valueType})${size}_storage`;
+}
+
+function getValueType(typeName: string): string {
+  return optimisticMatch(typeName, /<(.*)>/)[1];
+}
+
+function stripContractName(s: string): string {
+  return optimisticMatch(s, /(.*)\.(.*)/)[2];
```
frangio, line 239: The type of this function is wrong, it can return `undefined` if there is no match. The same error is in the previous two functions.

I'm not sure I understand the point of the function `optimisticMatch`. As far as I can tell you can achieve the same effect using the `?.` operator:

```
s.match(/(.*)\.(.*)/)?.[2]
```

packages/core/src/scripts/migrate-oz-sdk-project.ts
```diff
@@ -0,0 +1,351 @@
...
+
+function transformStructTypeName(typeName: string): string {
+  const valueType = stripContractName(getValueType(typeName));
+  return `t_struct(${valueType})_storage`;
+}
+
+function transformEnumTypeName(typeName: string): string {
+  const valueType = stripContractName(getValueType(typeName));
+  return `t_enum(${valueType})`;
+}
```
frangio, line 216: Enums and structs also have the AST id as part of their name in the plugins. For example `t_enum(Direction)92` and `t_struct(Foo)89_storage`.

I thought we had this information but I think I was mistaken. The SDK stores the AST ids for the state variables, but not for the struct definition.

In any case, it's not critical, everything should work without the id. We should respect the same format, though, so we have to make up a number and put it in. This number we make up should be unique in the same way that AST ids would be unique, so that future code can rely on that. (By this I mean we can't just hardcode the number 42 everywhere.)

This should be doable but it may require some changes to the code because the made up number has to be propagated for each type... unless it can be generated deterministically from the name alone. What do you think?

packages/core/src/scripts/migrate-oz-sdk-project.ts
```diff
@@ -0,0 +1,351 @@
...
+  return `t_mapping(${valueType})`;
+}
+
+function transformDynArrayTypeName(typeName: string): string {
+  const valueType = transformTypeName(getValueType(typeName));
+  return `t_array(${valueType})dyn_storage`;
+}
+
+function transformStaticArrayTypeName(typeName: string): string {
+  const size = optimisticMatch(typeName, /:(\d*)/)[1];
```
frangio, line 229: I would use `+` instead of `*` in the regular expression because we expect there will be at least one number. This way if that expectation is broken we will find out.

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+
+function getNetworkName(fileName: string): string {
+  return path.basename(fileName, '.json');
+}
+
+function isDevelopmentNetwork(network: string): boolean {
+  // 31337      => buidler evm
+  // 1337       => ganache
+  // 13+ digits => ganache also uses timestamps
+  return /^dev-(3?1337|\d{13,})$/.test(network);
```
frangio, line 112: I was thinking we should match a timestamp since Ganache was released... What do you think?

```suggestion
  // 142 + 10 digits => ganache timestamp (some time since 2015)
  return /^dev-(3?1337|142\d{10})$/.test(network);
```

Ganache (then ethereumjs-testrpc) was released 5 years ago.

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+function isManifestFile(fileName: string): boolean {
+  const network = getNetworkName(fileName);
+  return isPublicNetwork(network) || isDevelopmentNetwork(network) || isUnknownNetwork(network);
+}
+
+function getNetworkName(fileName: string): string {
+  return path.basename(fileName, '.json');
+}
+
+function isDevelopmentNetwork(network: string): boolean {
```
frangio, line 108: 1337 should be removed. I was mistaken before, I thought 1337 would show up with `geth --dev`, but its default network id is 1. Ganache would never have a network id of 1337 unless it was manually set (and we wouldn't support that case).

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+import { ManifestData, ImplDeployment } from '../manifest';
+import type { StorageItem, StorageLayout, TypeItem, TypeMembers, StructMember } from '../storage';
+
+const OPEN_ZEPPELIN_FOLDER = '.openzeppelin';
+const EXPORT_FILE = 'openzeppelin-cli-export.json';
+const PROJECT_FILE = path.join(OPEN_ZEPPELIN_FOLDER, 'project.json');
+const SUCESS_CHECK = chalk.keyword('green')('✔') + ' ';
+
+export async function migrateLegacyProject(): Promise<void> {
+  const manifestsMigrationData = await migrateManifestFiles();
```
frangio, line 13: Can you rename the variable as well as the property in `MigrationOutput` to something containing "export data"? Perhaps `networkExportData`. I'm still confused by the term "migration data".

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+  return isUnknownNetwork(oldName) ? oldName.replace('dev', 'unknown') : oldName;
+}
+
+async function writeJSONFile(location: string, data: unknown): Promise<void> {
+  await fs.writeFile(location, JSON.stringify(data, null, 2));
+}
+
+function getExportData(oldManifest: LegacyManifest): ExportData {
+  // we flexibilize the typing to allow required fields deletion
+  const manifestsMigrationData: Partial<LegacyManifest> = oldManifest;
```
frangio, line 133: We need to create a copy to avoid modifying the argument. (In this sense, TypeScript is right to disallow deletion!)

```suggestion
  const manifestsMigrationData: Partial<LegacyManifest> = Object.assign({}, oldManifest);
```

Additionally, here's how to properly type this variable to be able to delete properties and return it without the unsafe `as` cast. (Not a suggestion because the cast should be removed too.)

```typescript
  const manifestsMigrationData: ExportData & Partial<LegacyManifest> = ...;
```

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+interface MigrationOutput {
+  newManifests: Record<string, ManifestData>;
+  manifestsMigrationData: unknown;
+}
+
+interface AddressWrapper {
+  address?: string;
+}
+
+interface ExportData {
```
frangio, line 319: I was surprised to see so many things are exported, but I think it does make sense to export those that are addresses.

But we should probably delete `manifestVersion`, `zosversion`, `version`, `frozen`, and `dependencies`.

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+    const newFileName = getNewManifestLocation(manifestFile);
+    await writeJSONFile(newFileName, newManifest);
+    console.log(SUCESS_CHECK + `Successfully migrated ${newFileName}`);
+  }
+
+  return manifestsMigrationData;
+}
+
+async function deleteLegacyFiles(): Promise<void> {
+  const oldManifests = await getManifests();
```
frangio, line 45: Can we make it so `getManifests` is only called once?

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+
+  await writeJSONFile(EXPORT_FILE, exportData);
+  console.log(SUCESS_CHECK + `Migration data exported to ${EXPORT_FILE}`);
+  await deleteLegacyFiles();
+
+  console.log("\nThese were your project's compiler options:");
+  console.log(compiler);
+}
+
+async function migrateManifestFiles() {
```
frangio, line 29: I want to propose a refactor here. First of all, the names `migrateManifestFiles` and `migrateManifests` are too similar. Secondly, I find the separation of concerns to be kind of weird, since one function reads from the file system and the other writes to it; I would expect those two things to happen in the same function/abstraction layer.

With that in mind I would rename the second function to `migrateManifestsData` and make it so it receives the legacy manifest objects already read from disk. But I would like to know if you have any other ideas for refactoring.

I find there are too many layers of "migration" and it becomes hard to follow.

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+  const newManifests: Record<string, ManifestData> = {};
+
+  for (const manifestFile of oldManifests) {
+    const oldManifest = JSON.parse(await fs.readFile(manifestFile, 'utf8'));
+
+    if (oldManifest.manifestVersion === undefined) {
+      throw new Error('Migration failed: manifest version too old');
+    }
+
+    const [major, minor] = oldManifest.manifestVersion.split('.');
```
frangio, line 68: I think this should use the same version comparison library we use in `manifest.ts`.

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+    throw new Error('Could not find implementation address');
+  }
+
+  return {
+    address: contract.address,
+    layout: transformLayout(contract),
+  };
+}
+
+function transformLayout(contract: LegacyContract): StorageLayout {
```
frangio, line 184: If either `storage` or `types` are undefined, I think it should be an error, rather than silently ignored.

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+
+function transformTypes(oldTypes: LegacyTypes): Record<string, TypeItem> {
+  const newTypes: Record<string, TypeItem> = {};
+  for (const typeName in oldTypes) {
+    newTypes[transformTypeName(typeName)] = transformType(getTypeKind(typeName), oldTypes[typeName]);
+  }
+  return newTypes;
+}
+
+function transformType(typeKind: string, oldType: LegacyType): TypeItem {
```
frangio, line 209:
```suggestion
function transformType(typeKind: TypeKind, oldType: LegacyType): TypeItem {
```

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+
+function getValueType(typeName: string): string {
+  return optimisticMatch(typeName, /<(.+)>/)[1];
+}
+
+function stripContractName(s: string): string {
+  return optimisticMatch(s, /(.+)\.(.+)/)[2];
+}
+
+function optimisticMatch(s: string, rg: RegExp): string[] {
```
frangio, line 284: Can you remove this function and use the non-null assertion operator `!`?

e.g. `s.match(/.../)![2]`

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+  return `t_array(${valueType})dyn_storage`;
+}
+
+function transformStaticArrayTypeName(typeName: string): string {
+  const size = optimisticMatch(typeName, /:(\d+)/)[1];
+  const valueType = transformTypeName(getValueType(typeName));
+  return `t_array(${valueType})${size}_storage`;
+}
+
+function getValueType(typeName: string): string {
```
frangio, line 276: I think this would be more accurately named `getArgument`.

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+      return transformStaticArrayTypeName(typeName);
+    case 'Elementary':
+      return typeName;
+    default:
+      throw new Error(`Unknown type: ${typeName}`);
+  }
+}
+
+function transformStructTypeName(typeName: string): string {
+  const valueType = stripContractName(getValueType(typeName));
```
frangio, line 251: For structs and enums this is `name` rather than `valueType`.

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,426 @@
...
+}
+
+function transformEnumTypeName(typeName: string): string {
+  const valueType = stripContractName(getValueType(typeName));
+  return `t_enum(${valueType})`;
+}
+
+function transformMappingTypeName(typeName: string): string {
+  const valueType = transformTypeName(getValueType(typeName));
+  return `t_mapping(${valueType})`;
```
frangio, line 262: This would be missing the key type. Did we settle on using `unknown`?

```suggestion
  return `t_mapping(unknown,${valueType})`;
```

packages/core/src/scripts/migrate-legacy-project.ts
frangio: I couldn't find documentation for this. The recommendation should be to first update to the latest version of OpenZeppelin CLI, then migrate to the plugins.

packages/core/src/scripts/migrate-oz-cli-project.ts
martriay: The current script doesn't allow `solidityLibs`. The fields in this type are actually the ones in `LegacyManifest` minus the ones we delete from it.

packages/core/src/scripts/migrate-oz-cli-project.ts
martriay: We could, I did it this way because I thought making it a parameter of this function might be confusing/obscure since it's deleting more than legacy manifests.

packages/core/src/scripts/migrate-oz-cli-project.ts
frangio: Good point, I forgot about that.

packages/core/src/scripts/migrate-oz-cli-project.ts
martriay: I separated the read + migration functions from the write ones to make sure the migration is successful before rewriting any file. The idea is to fail before touching anything.

The migration of the files depends on the successful migration of the abstract manifests.

packages/core/src/scripts/migrate-oz-cli-project.ts
martriay: Why? can't a contract exist without any variables defined in it? They're also optional members of the types.

packages/core/src/scripts/migrate-oz-cli-project.ts
martriay: No, `optimisticMatch` returns an array with the same element if no match is found.

packages/core/src/scripts/migrate-oz-cli-project.ts
frangio: In that case it should already be an empty array in the CLI layout, isn't it?

packages/core/src/scripts/migrate-oz-cli-project.ts
frangio: But that isn't used anywhere.

packages/core/src/scripts/migrate-oz-cli-project.ts
martriay: What if it's run in a system with an outdated clock?

frangio: In that case the timestamp could have less than 13 digits too.

packages/core/src/scripts/migrate-oz-cli-project.ts
martriay: Not sure, but I think we should be compliant with the types.

packages/core/src/scripts/migrate-oz-cli-project.ts
frangio: I think it's best to read it once only.

packages/core/src/scripts/migrate-oz-cli-project.ts
martriay: Yes, in `stripContractName` for example

packages/core/src/scripts/migrate-oz-cli-project.ts
martriay: Shouldn't we then deep copy instead of `Object.assign()`?

frangio: By "avoid modifying the argument" I was referring to the `delete` operations that are also contained in this function. I think we can establish the convention that objects aren't modified unless they are meant to be, and if so it's not necessary to deep copy.

packages/core/src/scripts/migrate-oz-cli-project.ts
frangio: My suggestion was embarrassingly mistaken, so let's go back to just checking the number of digits.

packages/core/src/scripts/migrate-oz-cli-project.ts
frangio: Being compliant with the types means not ignoring the different possible types. Throwing an error is not ignoring.

The SDK types this as an optional property but it never emits it as undefined. See [`getStorageLayout`](https://github.com/OpenZeppelin/openzeppelin-sdk/blob/6c040b36052519f09569d6c40493707f2149b374/packages/lib/src/validations/Storage.ts#L10) and [`StorageLayoutInfo`](https://github.com/OpenZeppelin/openzeppelin-sdk/blob/6c040b36052519f09569d6c40493707f2149b374/packages/lib/src/validations/Storage.ts#L46-L49). If we find `undefined`, it's an error and we should abort the migration.

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,434 @@
...
+function getArgument(typeName: string): string {
+  // here we assume the regex has been already validated
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  return typeName.match(/<(.+)>/)![1];
+}
+
+function stripContractName(s: string): string {
+  const match = s.match(/(.+)\.(.+)/);
+  // input might not contain a contract name, so we fallback to it
+  return match ? match[2] : s;
```
frangio, line 298: I think this is better expressed by an optional regex group.

```suggestion
  // regex always matches
  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
  return s.match(/(.+\.)?(.+)/)![2];
```

packages/core/src/deployment.ts
```diff
@@ -38,31 +49,42 @@ export async function resumeOrDeploy<T extends Deployment>(
...
+      }
+      debug('verifying deployment tx mined', txHash);
+      const tx = await getTransactionByHash(provider, txHash);
+      if (tx === null) {
         throw new InvalidDeployment(deployment);
-      } else {
-        return;
       }
+      if (tx.blockHash !== null) {
+        debug('verifying deployment tx mined succeeded', txHash);
```
frangio, line 74: I understand the intent of adding "succeeded" at the end, but it ends up being a confusing sentence.

```suggestion
        debug('succeeded verifying deployment tx mined', txHash);
```

packages/core/src/deployment.ts
```diff
@@ -38,31 +49,42 @@ export async function resumeOrDeploy<T extends Deployment>(
...
+      await sleep(pollInterval);
     }
-    debug('waiting for deployment tx mined', deployment.txHash);
-    await sleep(5e3);
   }
 
-  // A timeout is NOT an InvalidDeployment
-  throw new TransactionMinedTimeout(deployment);
+  if (await hasCode(provider, address)) {
+    debug('verifying deployment succeeded', txHash);
```
frangio, line 83: This debug log was before the call to `getCode`, so it should go before `hasCode`. Also, the same comment I had above about the placement of "succeeded".

```suggestion
  debug('succeeded verifying deployment', txHash);
  if (await hasCode(provider, address)) {
```

packages/core/src/deployment.ts
```diff
@@ -73,6 +95,6 @@ export class TransactionMinedTimeout extends Error {
 
 export class InvalidDeployment extends Error {
   constructor(readonly deployment: Deployment) {
-    super(`Invalid deployment ${deployment.txHash}`);
+    super(`Invalid deployment with address ${deployment.address} and txHash ${deployment.txHash}`);
```
frangio, line 98:
```suggestion
    super(`Invalid deployment with address ${deployment.address} and txHash ${deployment.txHash ?? 'unknown'}`);
```

Does this trigger the line length limit...?

packages/core/src/deployment.ts
```diff
@@ -1,34 +1,45 @@
...
-      debug('resuming previous deployment', cached.txHash);
-      return cached;
-    } else if (!(await isDevelopmentNetwork(provider))) {
-      // If the transaction is not found we throw an error, except if we're in
-      // a development network then we simply silently redeploy.
-      throw new InvalidDeployment(cached);
+    const { txHash, address } = cached;
+    if (txHash === undefined) {
+      // Deployments can have address but no txHash if they were migrated from OpenZeppelin CLI
+      if (await hasCode(provider, address)) {
```
frangio, line 22: I don't think this should be here. If there is no `txHash`, I would just return the deployment without checking anything. `resumeOrDeploy` makes no guarantee about the contents of the address (equivalently, it makes no guarantee about the success of the transaction, only that it has been mined). `waitAndValidateDeployment` will then check the contents of the address to see if it has code.

packages/core/src/scripts/migrate-oz-cli-project.ts
```diff
@@ -0,0 +1,434 @@
...
+}
+
+function getNetworkName(filename: string): string {
+  return path.basename(filename, '.json');
+}
+
+function isDevelopmentNetwork(network: string): boolean {
+  // 142 + 10 digits => ganache timestamp (some time since 2015)
+  // 31337           => buidler evm
+  return /^dev-(31337|142\d{10})$/.test(network);
```
frangio, line 119:
```suggestion
  // 13+ digits      => ganache timestamp
  // 31337           => buidler evm
  return /^dev-(31337|\d{13,})$/.test(network);
```

frangio (approved):
Great work! Thanks!

PR #155, opened 2020-09-06
fix: links to main README

julianmrodri:
* Fixes links to the main `README.md` file from within the Plugin folders (both `/plugin-buidler` and `/plugin-truffle`.


frangio (approved):
Good catch. Thank you!

PR #162, opened 2020-09-10
Added view modifier to Admin only calls

radek1st:


frangio:
Thank you for the suggestion @radek1st. This doesn't compile because Solidity sees that the `ifAdmin` modifier potentially modifies storage via `delegatecall`. This is related to a family of usability issues because of which we recommed using this contract together with a `ProxyAdmin` contract, which is the setup these plugins create.

Can you share how you ran into this issue and what motivates you to make this change? The use case might be useful for us to know, and we may also be able to suggest some alternatives.

PR #171, opened 2020-09-15
Consider AST values can be undefined in solc >=0.7.1

### LINKED ISSUE #169: "Variable ... assigned an initial value" incorrectly reported with Solidity 0.7.1
Following the [Truffle Upgrades guide](https://forum.openzeppelin.com/t/openzeppelin-truffle-upgrades/3579) using Solidity 0.7.1 incorrectly reports "Variable ... assigned an initial value"

```console
$ npx truffle test

Compiling your contracts...
===========================
> Compiling ./contracts/Box.sol
> Compiling ./contracts/Migrations.sol
> Artifacts written to /tmp/test--1719-7SgrG1tvBhGC
> Compiled successfully using:
   - solc: 0.7.1+commit.f4a555be.Emscripten.clang



  Contract: Box (proxy)
    1) "before each" hook for "retrieve returns a value previously initialized"


  0 passing (85ms)
  1 failing

  1) Contract: Box (proxy)
       "before each" hook for "retrieve returns a value previously initialized":
     Error: Contract `Box` is not upgrade safe

Box.sol:6: Variable `value` is assigned an initial value
    Move the assignment to the initializer
    https://zpl.in/upgrades/error-004
      at Object.assertUpgradeSafe (node_modules/@openzeppelin/upgrades-core/src/validate.ts:160:11)
      at deployProxy (node_modules/@openzeppelin/truffle-upgrades/src/deploy-proxy.ts:31:3)
```

Changing the compiler to 0.7.0 and removing the build directory and the test passes successfully.
```console
$ npx truffle test

Compiling your contracts...
===========================
> Compiling ./contracts/Box.sol
> Compiling ./contracts/Migrations.sol
> Artifacts written to /tmp/test--1745-fxB10p9tJ53z
> Compiled successfully using:
   - solc: 0.7.0+commit.9e61f92b.Emscripten.clang



  Contract: Box (proxy)
    ✓ retrieve returns a value previously initialized


  1 passing (3s)
```

The same issue occurs with the Buidler plugin
```console
$ npx buidler test
Compiling...
Downloading compiler version 0.7.1
Compiled 1 contract successfully


  Box (proxy)
    1) "before each" hook for "retrieve returns a value previously initialized"


  0 passing (657ms)
  1 failing

  1) Box (proxy)
       "before each" hook for "retrieve returns a value previously initialized":
     Error: Contract `Box` is not upgrade safe

contracts/Box.sol:6: Variable `value` is assigned an initial value
    Move the assignment to the initializer
    https://zpl.in/upgrades/error-004
      at Object.assertUpgradeSafe (node_modules/@openzeppelin/upgrades-core/src/validate.ts:160:11)
      at Proxy.deployProxy (node_modules/@openzeppelin/buidler-upgrades/src/deploy-proxy.ts:33:5)
```
---

## Box.sol
```solidity
// contracts/Box.sol
pragma solidity ^0.7.0;
 
contract Box {
    uint256 private value;
 
    // Emitted when the stored value changes
    event ValueChanged(uint256 newValue);
 
    // Stores a new value in the contract
    function store(uint256 newValue) public {
        value = newValue;
        emit ValueChanged(newValue);
    }
 
    // Reads the last stored value
    function retrieve() public view returns (uint256) {
        return value;
    }
}
```

## Box.proxy.test.js
```js
// test/Box.proxy.test.js
// Load dependencies
const { expect } = require('chai');
const { deployProxy } = require('@openzeppelin/truffle-upgrades');
 
// Load compiled artifacts
const Box = artifacts.require('Box');
 
// Start test block
contract('Box (proxy)', function () {
  beforeEach(async function () {
    // Deploy a new Box contract for each test
    this.box = await deployProxy(Box, [42], {initializer: 'store'});
  });
 
  // Test case
  it('retrieve returns a value previously initialized', async function () {
    // Test if the returned value is the same one
    // Note that we need to use strings to compare the 256 bit integers
    expect((await this.box.retrieve()).toString()).to.equal('42');
  });
});
```

## truffle-config.js
```js
...
  compilers: {
    solc: {
      version: "0.7.1", 
...
```

frangio:
Fixes #169 

Nullable properties in the AST are now optional, so they can now have both `null` and `undefined` as a value (depending on the compiler version used). This was causing the state variable initialization check to produce false positive errors.

I also split the changelog so there is a separate one for each subpackage, because the versioning is different.

Also updates solidity-ast to the latest version which includes `undefined` in the types.


martriay (approved):

martriay:
Looks good, thanks!

PR #182, opened 2020-09-21
Add override flag to enable linked libraries

### LINKED ISSUE #144: Offer an unsafeAllowLibraries option
In the same way we have `unsafeAllowCustomTypes` we could offer `unsafeAllowLibraries` to work around the lack of support for external libraries until we implement it (#52).

Note that this is probably not trivial as simply ignoring the error, because linking libraries changes the bytecode hash, and this may break other assumptions we have. We may want to store in the network file the bytecode positions where linked library addresses are stored.

julianmrodri:
###  Override option for External Libraries 

**Description**

Provides a new `unsafeAllowLinkedLibraries` override option which allows the user to deploy implementation contracts linked to External Libraries.  When this flag is set to `true`, it allows the deployment/upgrade to proceed but showing a warning message to the user.

**Important Notes**
* Libraries should be **deployed and linked** prior to calling `deployProxy` and `upgradeProxy`.
* In the case of **Buidler** the plugin handles the "linked" bytecode by restoring the original placeholders and enforcing all validations work as expected (linked libraries are stored based on the `linkReferences` section of  `solc-output`). We are assuming, as the **Buidler** team recommends, that the libraries will be linked directly **in the bytecode** before calling `deployProxy` and `upgradeProxy`. This forces us to rebuild the **unlinked** bytecode to apply all validations.

* In the case of **Truffle** an additional step was required to recreate the `linkReferences` section by parsing directly the bytecode, as this section is not provided by Truffle in the Artifacts. Also it was required to retrieve the **linked** addresses from the `networks[networkId].links` attribute. **Truffle** handles linked libraries in a different way than the Buidler approach. It always manages the **unlinked** bytecode and only links the actual library at the time of **deployment**. This required us to manually **link** the libraries into the bytecode, to get the correct identifier for the `manifest`.

* We had to include a new field with the `linkedBytecodeWithoutMetadata` hash in the `version` object. This is required because for all validations the plugins need the **unlinked** versions (as it uses today), but for the `impl-store` to properly **identify changes on deployed implementations**,  we need to work with the **fully linked** version as key in the `manifest`. 

* Testing included for both Plugins, with the following edge cases:
   
1.  Deploying a **Proxy** with an implementation linked to a Library.  Then deploying **again** but using a different library address.  The plugins now **REDEPLOYS** the implementation and a new proxy as expected.
2. Upgrading a **Proxy** with an implementation in which the only change is a link to a new library address. The plugin **REDEPLOYS** the implementation as expected, using the same Proxy.

Also ran all tests successfully for `solc` versions `0.5.1`, `0.6.1`, `0.7.1`.

* Documentation updated with instructions on how to use the new **override** flag.

**Out of Scope:** This change only contemplates the **override** function, but it is still up to the users responsibility to ensure the external libraries being used are **upgrade safe**. This change does not check the linked libraries source code.

Closes #144


packages/core/src/link-refs.ts
```diff
@@ -0,0 +1,38 @@
+import { SolcBytecode } from './solc-api';
+
+export interface LinkReference {
+  src: string;
+  name: string;
+  length: number;
+  start: number;
+  placeHolder: string;
```
frangio, line 8: This can be capitalized as `placeholder`.

packages/core/src/link-refs.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+  return linkRefs;
+}
+
+export function replaceLinkReferences(bytecode: string, linkReferences: LinkReference[]): string {
+  let unlinkedBytecode: string = bytecode;
+  for (const linkRef of linkReferences) {
+    const { length, start, placeHolder } = linkRef;
+    unlinkedBytecode =
+      unlinkedBytecode.substr(0, 2 + start * 2) + placeHolder + unlinkedBytecode.substr(2 + (start + length) * 2);
```
frangio, line 35: I feel uneasy about the `2 +` part, especially given that `extractLinkReferences` doesn't have that (presumably because Solidity returns the hex encoding without the `0x` prefix).

I think I would prefer if we removed the `0x` prefix at the beginning of the function ([as here](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/4f2a046804774cc10e15c989afdfc8a3d4bc176a/packages/core/src/version.ts#L21)), and then re-adding it in the return statement.

packages/core/src/validate.ts
```diff
@@ -111,10 +128,25 @@ export function getContractVersion(validation: Validation, contractName: string)
...
+
+function isValidVersion(validation: Validation, version: Version): boolean {
+  return findContractByVersion(validation, version) !== undefined;
+}
+
+export function getValidVersion(validation: Validation, bytecode: string): Version {
+  let version: Version = getVersion(bytecode);
+  if (!isValidVersion(validation, version)) {
+    const unlinkedBytecode: string = getUnlinkedBytecode(validation, bytecode);
+    version = getVersion(unlinkedBytecode);
```
frangio, line 143: In this case I would favor multiple return statements rather than the mutable `version` variable.

packages/core/src/validate.ts
```diff
@@ -134,15 +166,51 @@ export function getStorageLayout(validation: Validation, version: Version): Stor
   return layout;
 }
 
-export function assertUpgradeSafe(validation: Validation, version: Version, unsafeAllowCustomTypes = false): void {
+function getUnlinkedBytecode(validation: Validation, bytecode: string): string {
+  let unlinkedBytecode: string = bytecode;
+  Object.keys(validation)
+    .filter(name => validation[name].linkReferences.length > 0)
+    .find(name => {
```
frangio, line 173: When the loop is used for the side effects only, we should use a `for` loop.

packages/core/src/validate.ts
```diff
@@ -134,15 +166,51 @@ export function getStorageLayout(validation: Validation, version: Version): Stor
...
+      const linkReferences: LinkReference[] = getLinkReferences(validation, name);
+      unlinkedBytecode = replaceLinkReferences(bytecode, linkReferences);
+      const version = getVersion(unlinkedBytecode);
+      validation[name].version?.withMetadata === version.withMetadata;
```
frangio, line 177: Plus, the `return` is missing here!

packages/core/src/validate.ts
```diff
@@ -134,15 +166,51 @@ export function getStorageLayout(validation: Validation, version: Version): Stor
...
+  return c.linkReferences
+    .concat(...c.inherit.map(name => validation[name].linkReferences))
+    .concat(...c.libraries.map(name => validation[name].linkReferences));
+}
+
+export function assertUpgradeSafe(validation: Validation, version: Version, opts: ValidationOptions): void {
   const contractName = getContractName(validation, version);
   let errors = getErrors(validation, version);
   let thereAreExceptions = false;
+  let exceptionMessage = '';
```
frangio, line 194: Instead of concatenating all warning messages into a single warning, I would like this to print the multiple warnings separately, each with its own "Warning" label. This is to make it stand out as much as possible.

packages/core/src/validate.ts
```diff
@@ -196,7 +263,7 @@ const errorInfo: ErrorDescriptions<ValidationError> = {
   },
   'external-library-linking': {
     msg: e => `Linking external libraries like \`${e.name}\` is not yet supported`,
-    hint: `Stick to libraries with internal functions only`,
+    hint: `If you have manually checked that the libraries are upgrade safe, you can skip this check with the \`unsafeAllowLinkedLibraries\` flag`,
```
frangio, line 266: I would like to keep the previous recommendation, because I suspect some people may not be aware that they could be using internal functions.

```suggestion
    hint: `Use libraries with internal functions only, or skip this check with the \`unsafeAllowLinkedLibraries\` flag if you have manually checked that the libraries are upgrade safe`,
```

packages/core/src/link-refs.ts
```diff
@@ -0,0 +1,38 @@
...
+        length,
+        start,
+        placeHolder,
+      });
+    }
+  }
+  return linkRefs;
+}
+
+export function replaceLinkReferences(bytecode: string, linkReferences: LinkReference[]): string {
```
frangio, line 30: I feel like this should be called "unlink bytecode". What do you think?

packages/core/src/validate.ts
```diff
@@ -134,15 +166,51 @@ export function getStorageLayout(validation: Validation, version: Version): Stor
   return layout;
 }
 
-export function assertUpgradeSafe(validation: Validation, version: Version, unsafeAllowCustomTypes = false): void {
+function getUnlinkedBytecode(validation: Validation, bytecode: string): string {
```
frangio, line 169: If a match isn't found, this function should throw an error.

frangio (commented):
Some initial comments. I still haven't looked at everything!

packages/core/src/link-refs.ts
julianmrodri: Agree. Will do.

packages/core/src/link-refs.ts
julianmrodri: Yes will do that. I agree its not very intuitive.

packages/core/src/validate.ts
julianmrodri: Will take care of it. Do you think we should proceed with the refactor and handle this in `version.ts` ? It would be neater if you can simply query `getVersion` and get the correct one. (we also remove the `validVersion` naming which we know is not nice).

packages/core/src/link-refs.ts
julianmrodri: Agree.. actually works much better when this is moved to `version.ts` as it is very related about handling bytecode

packages/core/src/validate.ts
julianmrodri: Done.

packages/core/src/validate.ts
julianmrodri: Done.

packages/core/src/validate.ts
julianmrodri: This does not apply anymore based on the new design.

packages/core/src/validate.ts
```diff
@@ -196,7 +247,7 @@ const errorInfo: ErrorDescriptions<ValidationError> = {
   },
   'external-library-linking': {
     msg: e => `Linking external libraries like \`${e.name}\` is not yet supported`,
-    hint: `Stick to libraries with internal functions only`,
+    hint: `Use libraries with internal functions only, or skip this check with the \`unsafeAllowLinkedLibraries\` flag if you have manually checked that the libraries are upgrade safe`,
```
frangio, line 250: This hint is rather long and it doesn't look very good in the console. Options: 1) insert <code>\n____</code> in the middle to wrap it manually (underscores should be spaces for indentation), 2) handle it in `describeError`.

packages/core/src/version.ts
```diff
@@ -4,13 +4,15 @@ import cbor from 'cbor';
...
+  linkedWithoutMetadata: string;
 }
 
-export function getVersion(bytecode: string): Version {
+export function getVersion(bytecode: string, linkedBytecode?: string): Version {
   if (bytecode !== '') {
     return {
       withMetadata: hashBytecode(bytecode),
       withoutMetadata: hashBytecodeWithoutMetadata(bytecode),
+      linkedWithoutMetadata: hashBytecodeWithoutMetadata(linkedBytecode ?? bytecode),
```
frangio, line 15: I think it would be better to not default to `bytecode` and only set this if the `linkedBytecode` argument is present.

frangio (commented):
Great work @julianmrodri!

packages/plugin-buidler/test/linked-libraries.js
```diff
@@ -0,0 +1,128 @@
...
+  assert.strictEqual('V1', await token3.getLibraryVersion());
+
+  console.log('Calling transferPercent');
+  assert.strictEqual('10000', (await token3.balanceOf(ownerAddress)).toString());
+  await token3.transferPercent(testAddress, 10);
+  assert.strictEqual('1000', (await token3.balanceOf(testAddress)).toString());
+  assert.strictEqual('9000', (await token3.balanceOf(ownerAddress)).toString());
+}
+
+// linkBytecode: performs linking by replacing placeholders with deployed addresses
```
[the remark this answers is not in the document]
julianmrodri: I was pointed to this issue by the Buidler team which has that code. I can include it. https://github.com/nomiclabs/buidler/issues/611#issuecomment-638891597

packages/plugin-buidler/test/linked-libraries.js
```diff
@@ -0,0 +1,128 @@
...
+  const token = await upgrades.deployProxy(Token, ['TKN', 10000], { unsafeAllowLinkedLibraries: true });
+
+  assert.strictEqual('10000', (await token.totalSupply()).toString());
+  assert.strictEqual('V1', await token.getLibraryVersion());
+
+  console.log('Deploying Token with different Library...');
+  const TokenNew = await getLinkedContractFactory('Token', { SafeMath: safeMathLib2.address });
+  const tokenNew = await upgrades.deployProxy(TokenNew, ['TKN', 5000], { unsafeAllowLinkedLibraries: true });
+
+  assert.notEqual(token.address, tokenNew.address);
```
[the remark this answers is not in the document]
julianmrodri: No, I can remove.

packages/core/src/solc-api.ts
```diff
@@ -26,11 +26,18 @@ export interface SolcInput {
...
-      [library in string]: unknown;
-    };
+export type SolcLinkReferences = {
+  [file in string]: {
+    [library in string]: [
+      {
+        length: number;
+        start: number;
+      },
+    ];
```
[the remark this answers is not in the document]
julianmrodri: Good catch Thanks!

packages/plugin-truffle/src/validate.ts
```diff
@@ -69,3 +80,44 @@ function reconstructSolcInputOutput(artifacts: TruffleArtifact[]): { input: Solc
...
+    let linkedBytecode: string = Contract.bytecode;
+    Object.keys(networkInfo.links).forEach((name: string) => {
+      const address = networkInfo.links[name].replace(/^0x/, '');
+      const regex = new RegExp(`__${name}_+`, 'g');
+      linkedBytecode = linkedBytecode.replace(regex, address);
+    });
+    return linkedBytecode;
+  } else {
+    return Contract.bytecode;
+  }
```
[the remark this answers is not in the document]
julianmrodri: Done. Had to add a default value for `address` in case `networkInfo` is `undefined` to be able to use it in `replace`.

packages/core/src/validate.ts
julianmrodri: Done. Implemented option 1. This is the approach also used in the Warning messages which are quite long.

packages/plugin-truffle/src/truffle.ts
```diff
@@ -16,6 +16,17 @@ export interface ContractClass {
   class_defaults: ContractClassDefaults;
   contractName: string;
   address?: string;
+  networks?: {
+    [id: string]: NetworkObject;
```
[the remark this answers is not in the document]
julianmrodri: For this one I used the exact same definitions that Truffle has, and which are generated based on the schemas defined [here](https://github.com/trufflesuite/truffle/blob/develop/packages/contract-schema/spec/contract-object.spec.json).   We can try to implement the change if you prefer (its not that trivial though, requires quite some changes), but I think its also valuable to use the shared definition as this is only used for the `Truffle` contract.

packages/plugin-truffle/src/options.ts
```diff
@@ -1,14 +1,16 @@
 import { Deployer, ContractClass, ContractInstance, getTruffleConfig } from './truffle';
+import { ValidationOptions, withValidationDefaults } from '@openzeppelin/upgrades-core';
 
-export interface Options {
+export interface Options extends ValidationOptions {
   deployer?: Deployer;
-  unsafeAllowCustomTypes?: boolean;
 }
 
 export function withDefaults(opts: Options): Required<Options> {
```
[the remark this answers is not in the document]
julianmrodri: Done. I had to keep anyway a `withDefaults` function which returns the results of `withValidationDefaults` and `withDeployDefaults` combined, as this is needed in in `upgrade-proxy`. But I think now it looks clearer by having `Options = DeployOptions & ValidationOptions`.

packages/core/src/version.ts
julianmrodri: The issue if we make `linkedWithoutMetadata` optional is that then we need to do additional validations when calling `fetchOrDeploy` as this is the parameter that is passed and it is clearly mandatory. 

One workaround would bet o do something like this:

```javascript
  return fetchOrDeployGeneric(implLens(version.linkedWithoutMetadata ?? version.withoutMetadata), provider, deploy);
```
What do you think?

julianmrodri: If we proceed this is how `Version` and `getVersion` look like:

```typescript
export interface Version {
  withMetadata: string;
  withoutMetadata: string;
  linkedWithoutMetadata?: string;
}

export function getVersion(bytecode: string, linkedBytecode?: string): Version {
  if (bytecode !== '') {
    const version: Version = {
      withMetadata: hashBytecode(bytecode),
      withoutMetadata: hashBytecodeWithoutMetadata(bytecode),
    };
    if (linkedBytecode) {
      version.linkedWithoutMetadata = hashBytecodeWithoutMetadata(linkedBytecode);
    }

    return version;
  } else {
    throw new Error('Abstract contract not allowed here');
  }
}
```

julianmrodri:
@frangio  Please review again the  changes and let me know. 

The ones I set as `Resolved` are all implemented. The other 2-3 items I think require still a bit of discussion, specially the last one which I have the change done locally but wanted to get your input before proceeding. Thanks!

packages/core/src/version.ts
frangio: Hm, that's true. Let's keep it then!

packages/plugin-truffle/src/truffle.ts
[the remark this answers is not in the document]
frangio: I see. Let's keep it then!

frangio (approved):
The changes look great. Everything is looking good to me. Thanks a lot @julianmrodri!

packages/core/src/version.ts
julianmrodri: Another option is to return it empty (but defined) unless the parameter is specified and in that case it uses the hash.

Issue #136, opened 2020-08-20
Let users ask for a new ProxyAdmin to be created or detect when ownership of it was transferred


mverzilli:
I don't want to be prescriptive with this, so I'll do my best to explain the use case instead of suggesting solutions as in the title :).

1. I deploy an upgradeable contract (C1) using `deployProxy`.
2. I then `transferOwnership` of the ProxyAdmin to a multisig (let's call it M1). 
3. I now deploy another contract (C2) using `deployProxy`.
4. I want another multisig (M2) to control this contract, but the underlying ProxyAdmin is now owned by M1. 

So what I'd need in that case is to deploy C2 using a brand new ProxyAdmin that the deployer account controls. In practice, I solved the issue by deleting the `.openzeppelin` folder, which made the plugin create a new ProxyAdmin for me. But it's a hacky solution. 

It'd be nice to be able to ask for a new ProxyAdmin to be created, for the plugin to detect that I'm no longer owner of the current one and/or to get a warning that I'm using a ProxyAdmin that I no longer control.

frangio:
Interesting!

>In practice, I solved the issue by deleting the .openzeppelin folder

That's definitely quite extreme, because you're also losing the C1 deployment which you may want to reuse later. A more specific workaround would be to open the network file and delete the `admin` field so that a new one is deployed.

---

There's several ways to approach this.

1. Letting the user "forget" the current admin (an `admin.forget()` function).
2. Managing multiple admins at the same time, letting the user request a new one.
3. Not reusing an admin that is not owned by an available account?

I feel that in some (most?) scenarios you would want to reuse the admin even if you don't own it, and in that case 3 is out of the question.

Managing multiple admins is a feature we've thought about before, but at this point I don't know if the use case would be common enough.

Adding a forget function to the `admin` module would be very simple to implement and it has the potential to be very useful, so I think it makes a lot of sense to implement.

arijoon:
An additional use case is having multiple user envs (e.g. lets say `dev` and `staging`) within the same network but with different accounts controlling them and therefore requiring multiple `ProxyAdmin` requirements. In this scenario we would still want implementations shared to minimise deployments but not `ProxyAdmin`.

[...]

PR #212, opened 2020-11-04
add: get unknown network name from env

sirpy:
Sometimes you may want to have dev/staging/prod contracts on same network
if you set NETWORK in env this allows for multiple deployment to same network ie:
export NETWORK='ropsten-staging'

### DIFF
--- a/packages/core/src/manifest.ts
+++ b/packages/core/src/manifest.ts
@@ -40,8 +40,7 @@ export class Manifest {
   }
 
   constructor(chainId: number) {
-    const networkName = process.env.NETWORK || networkNames[chainId] || 'unknown';
-    const name = `${networkName}-${chainId}`;
+    const name = process.env.NETWORK ? `${process.env.NETWORK}-${chainId}` : networkNames[chainId] ?? `unknown-${chainId}`;
     this.file = path.join(manifestDir, `${name}.json`);
   }
 
--- a/packages/core/src/manifest.ts
+++ b/packages/core/src/manifest.ts
@@ -40,7 +40,8 @@ export class Manifest {
   }
 
   constructor(chainId: number) {
-    const name = networkNames[chainId] ?? `unknown-${chainId}`;
+    const networkName = process.env.NETWORK || networkNames[chainId] || 'unknown';
+    const name =  `${networkName}-${chainId}`;
     this.file = path.join(manifestDir, `${name}.json`);
   }
 
--- a/packages/core/src/manifest.ts
+++ b/packages/core/src/manifest.ts
@@ -41,7 +41,7 @@ export class Manifest {
 
   constructor(chainId: number) {
     const networkName = process.env.NETWORK || networkNames[chainId] || 'unknown';
-    const name =  `${networkName}-${chainId}`;
+    const name = `${networkName}-${chainId}`;
     this.file = path.join(manifestDir, `${name}.json`);
   }
 


packages/core/src/manifest.ts
```diff
@@ -40,7 +40,8 @@ export class Manifest {
   }
 
   constructor(chainId: number) {
-    const name = networkNames[chainId] ?? `unknown-${chainId}`;
+    const networkName = process.env.NETWORK || networkNames[chainId] || 'unknown';
+    const name =  `${networkName}-${chainId}`;
```
sirpy, line 44:
```suggestion
    const name = `${networkName}-${chainId}`;
```

packages/core/src/manifest.ts
```diff
@@ -40,7 +40,8 @@ export class Manifest {
   }
 
   constructor(chainId: number) {
-    const name = networkNames[chainId] ?? `unknown-${chainId}`;
+    const networkName = process.env.NETWORK || networkNames[chainId] || 'unknown';
+    const name = `${networkName}-${chainId}`;
```
sirpy, line 44:
```suggestion
    const networkName = process.env.NETWORK ? `${process.env.NETWORK}-${chainId}` : networkNames[chainId] ?? `unknown-${chainId}`;
```

sirpy, line 44:
```suggestion
    const name = process.env.NETWORK ? `${process.env.NETWORK}-${chainId}` : networkNames[chainId] ?? `unknown-${chainId}`;
```

frangio:
Hi @sirpy! I think the intention here is interesting but I don't like the idea of using an environment variable for it...

I want to think about a better way to achieve the same goal. Do you use Buidler/Hardhat or Truffle?

sirpy:
truffle.
other options i can think off are:
1. passing down as variable
2. allowing to change networkNames object which is currently frozen. so a user could do setNetworkName(3,'ropsten-staging') so for current run chainid 3 would resolve to new network name.

frangio:
Can you explain to me the precise reasons why you want to have separate network files for dev/staging/prod?

I don't see an issue in reusing implementation contracts, since they don't hold any state. There may be an issue with reusing the admin contract, but that can be solved in a different way by allowing you to have multiple admin contracts, as in https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/136.

sirpy:
> Can you explain to me the precise reasons why you want to have separate network files for dev/staging/prod?
> 
> I don't see an issue in reusing implementation contracts, since they don't hold any state. There may be an issue with reusing the admin contract, but that can be solved in a different way by allowing you to have multiple admin contracts, as in #136.

i'm not familiar with the internals but i want a different admin + state for each env, if you can suggest a way to do that which  could be integrated with the truffle plugin i would be happy to learn.
currently the deployProxy/upgradeProxy command reads the info it needs from the network file and based on it decided what to do

sirpy:
@frangio any updates on this?

frangio:
@sirpy Sorry, I don't think using an environment variable for this purpose is a good idea, and I don't think the alternatives are good either. This is more of a fundamental issue of how we're thinking of network state, and although I will definitely keep the use case you describe in mind going forward, I don't want to rush to providing a solution for it.

I'd encourage you to check out #136 and see if having multiple admins in the same network data file would be useful to you. However, given that you said you wanted different _state_ for each evironment, it might not be sufficient.

I can also suggest a couple of workarounds for you to use:

1) use different directories for each environment (`.openzepelin` is stored and read in the process working directory),
2) use different git branches.

sirpy:
@frangio why is allowing the user to set the network name is not a good solution? (setNetworkName(3,'ropsten-staging'))

this also allows custom support for all those sidechains like XDAI, POA, FUSE which are currently simply "unknown"
ie setNetworkName(122,'fuse')

frangio:
It's a different use case. We should add those networks as known in 

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/0761c2224464e36795e8e96da232f5e85904d6a6/packages/core/src/provider.ts#L52-L59

sirpy:
@frangio it makes no sense to make it hard coded.
I can probably give your a few more networks and there are probably a dozen i'm not aware of.
every time a developer is working on a new sidechain he should submit a PR?

frangio:
We could also use a shared list like [chainid.network](https://chainid.network). I've created an issue for this at https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/229.

sirpy:
@frangio it still makes no sense
who exactly updates this chainid.network? it is no different then making PR. i already see it doesnt have a network i know.
networks are popping like mushrooms after the rain.
allowing the developer to set the network names makes it compatible with whatever, plus solves the issue for having multiple envs for same network
what is your objection for letting the developer set the network name? it is a trivial config and you can still have your sane defaults

frangio:
@sirpy I appreciate the feedback but I will ask you to avoid comments such as "it makes no sense".

I rejected this PR as a way to use the same network in different environments. You're now discussing the separate issue of configuring names for "uknown" networks. They shouldn't necessarily be tackled in the same way, but I can see how solving the second issue may provide a solution for the first.

Using an environment variable will be a last resort. The project considers this is not the right tool for the job. The solution has to be either explicit in the use of the plugin API or integrated with the tooling (Truffle and Hardhat).

We could potentially use the network name as defined in the Truffle/Hardhat config. Hardhat already allows this, the network name is present in the environment `network` object. In Truffle I couldn't find it, but we can ask them to add it. What do you think of this idea?

sirpy:
that was always the purpose how to handle unknown networks. see title.
in any case to set network from truffle you would probably have some helper function, I simply suggest to expose it to developers as first step and meanwhile ask truffle to Also add access to it

sirpy:
@frangio can you reply to my last comment?

frangio:
I'm sorry, this is not a priority for us at the moment. I encourage you to use the workarounds I described above:

> I can also suggest a couple of workarounds for you to use:
> 
> -  use different directories for each environment (.openzepelin is stored and read in the process working directory),
> -  use different git branches.

Hronom:
@frangio I got in this situation when deploying to `Arbitrum sepolia` and getting `unknown-421614.json`.
I have in `hardhat.config.ts`:
```
        arbitrumSepolia: {
            url: process.env.ARBITRUM_SEPOLIA_NETWORK_RPC_URL || "",
            accounts: [process.env.PRIVATE_KEY || ""],
            chainId: 421614,
            gasPrice: 1,
        },
```

How to fix this and make name of file something reasonable? Can I change name somewhere inside `hardhat.config.ts`?

Issue #85, opened 2020-07-31
Override transaction options (from, gas, gasPrice) in create and upgrade proxy


spalladino:
Add `from`, `gas`, and `gasPrice` as optional values to the options parameter of `createProxy` and `upgradeProxy`, so the user can change the defaults for a specific transaction. For example:
```js
await upgradeProxy(existing.address, BoxV2, { deployer, from: '0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0', gasPrice: 10e9, gas: 1e6 });
```

(found in truffle plugin, unsure if applies to builder as well)

frangio:
We potentially want these settings to be granular since each function can correspond to multiple transactions.

pedrobranco:
> We potentially want these settings to be granular since each function can correspond to multiple transactions.

@frangio Agree, but at least the `from` option would be great to support.

frangio:
Note that the plugin will use the signer that comes attached to the `ImplFactory` in the Hardhat plugin, and I believe it also uses the same `from` default that was set for the contract class passed in the Truffle plugin.

This means that you can customize `from` by setting it in the contract abstraction you give the plugin (e.g. a `MyContractV2` class).

blabno:
I had a nightmare deploying to mainnet using `deployProxy`. I was unable to set `maxPriorityFeePerGas` which made me wait 36 hours to get the implementation deployed. After that I resumed the process, got the Admin contract deployed quickly, but then got stuck on Proxy deployment for another 2 days.
Overriding `maxPriorityFeePerGas` and `maxFeePerGas` is a must.
Ideally we should be able to resume the process and bump the fees.

frangio:
Here is a possible workaround for ethers.js:

```js
const FEE_DATA = {
    maxFeePerGas:         ethers.utils.parseUnits('100', 'gwei'),
    maxPriorityFeePerGas: ethers.utils.parseUnits('5',   'gwei'),
};

// Wrap the provider so we can override fee data.
const provider = new ethers.providers.FallbackProvider([ethers.provider], 1);
provider.getFeeData = async () => FEE_DATA;

// Create the signer for the mnemonic, connected to the provider with hardcoded fee data
const signer = ethers.Wallet.fromMnemonic(/* mnemonic */).connect(provider);

// Get the contract factory connected to signer so it uses hardcoded fee data
const MyContract = await ethers.getContractFactory('MyContract', signer);

[...]

PR #213, opened 2020-11-04
feature: Add deployer options as an argument

pedrobranco:
Draft PR for an initial proposal for fixing #85. 

I've only implemented `from` argument since other options doesn't seem so relevant as a custom deploy address.

### DIFF
--- a/packages/plugin-truffle/src/deploy-proxy.ts
+++ b/packages/plugin-truffle/src/deploy-proxy.ts
@@ -12,6 +12,7 @@ import { deploy } from './utils/deploy';
 import { getProxyFactory, getProxyAdminFactory } from './factories';
 import { wrapProvider } from './wrap-provider';
 import { Options, withDeployDefaults } from './options';
+import type { DeployerOptions } from './utils/deploy';
 
 interface InitializerOptions {
   initializer?: string | false;
@@ -21,6 +22,7 @@ export async function deployProxy(
   Contract: ContractClass,
   args: unknown[] = [],
   opts: Options & InitializerOptions = {},
+  deployerOpts: DeployerOptions = {},
 ): Promise<ContractInstance> {
   const { deployer } = withDeployDefaults(opts);
   const provider = wrapProvider(deployer.provider);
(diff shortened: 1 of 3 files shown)


frangio:
Thanks for giving this a shot @pedrobranco!

Let's see, there's several transactions for each `deployProxy` or `upgradeProxy` function call:

1. Implementation contract deployment
2. Admin contract deployment
3. Proxy contract deployment
4. Upgrade function call

In your PR you seem to only have used the `from` parameter for the first two. Why not also for the proxy?

For the function call it doesn't make as much sense because we should infer the sender from the admin contract (though we don't do this yet).

pedrobranco:
> In your PR you seem to only have used the from parameter for the first two. Why not also for the proxy?

True, didn't noticed it. The main goal here is to replace the default account when deploying, so for all of those transactions still seems fine.

> For the function call it doesn't make as much sense because we should infer the sender from the admin contract (though we don't do this yet).

Should we do it in this PR?

pedrobranco:
BTW, tests are failing with:
```
Error: while migrating AdminUpgradeabilityProxy: Invalid number of parameters for "undefined". Got 4 expected 3!
```
And by reverting the corresponding [this change](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/213/files#diff-5b7495dacd28e6c5dadbc547df0dcfeb1de99e63039fed72f19ecd949b3eddd2R48) also fails in `upgradeProxy`:
```
Error: Invalid number of parameters for "upgrade". Got 3 expected 2!
    at upgradeProxy (<REDACTED>/openzeppelin-upgrades/packages/plugin-truffle/src/upgrade-proxy.ts:84:15)

```
Seems that isn't allowing to pass the Web3 options. Correct me if I'm wrong but it seems like an issue already reported here: https://github.com/trufflesuite/truffle/issues/3506

pedrobranco:
@frangio can you help clarify these issues?

frangio:
@pedrobranco Sorry about the delay!

There does seem to be an underlying issue with Truffle. It's closer to https://github.com/trufflesuite/truffle/issues/2843 than the one you shared. But our situation is also a little bit different, see my comment https://github.com/trufflesuite/truffle/issues/2843#issuecomment-739083982.

The issue is we are sending an empty object and Truffle doesn't recognize that it's meant to be a params object and so sends it as part of the constructor arguments.

We will have to avoid sending the params argument when we don't have any actual params to send.

frangio:
I'm not sure when it will be released (later this week?) but Truffle has now fixed the problem you ran into in this PR.

https://github.com/trufflesuite/truffle/pull/3586

frangio:
This branch is outdated now and would likely need to be rewritten.

This issue is being tracked in https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/85.

PR #228, opened 2020-11-18
Add a function to silence warnings

### LINKED ISSUE #226: Add mechanism to reduce number of warnings printed
Users setting `unsafeAllowCustomTypes` flag get verbose warnings which is shown every time a contract is deployed which is extremely noisy when running tests.
```
You are using the `unsafeAllowCustomTypes` flag to skip storage checks for structs and enums.
Make sure you have manually checked the storage layout for incompatibilities.
```

Ideally there would be a mechanism to show a single line warning once when using `unsafeAllowCustomTypes` flag.

Requested in community forum: https://forum.openzeppelin.com/t/logging-verbose-during-local-tests-with-upgradeable-contracts/4633

frangio:
Fixes #226 

Requested by a user in the forum [[1]](https://forum.openzeppelin.com/t/logging-verbose-during-local-tests-with-upgradeable-contracts/4633). I think it makes sense.

### DIFF
--- a/packages/core/src/validate.ts
+++ b/packages/core/src/validate.ts
@@ -454,7 +454,7 @@ export function silenceWarnings(): void {
   console.error(
     '\n' +
       chalk.keyword('orange').bold('Warning: ') +
-      `All subsequent warnings will be silenced.\n\n` +
+      `All subsequent Upgrades warnings will be silenced.\n\n` +
       `    Make sure you have manually checked all uses of unsafe flags.\n`,
   );
   silenced = true;
--- a/packages/core/src/validate.ts
+++ b/packages/core/src/validate.ts
@@ -451,11 +451,13 @@ function* getEnumErrors(contractDef: ContractDefinition, decodeSrc: SrcDecoder):
 let silenced = false;
 
 export function silenceWarnings(): void {
-  console.error(
-    '\n' +
-      chalk.keyword('orange').bold('Warning: ') +
-      `All subsequent Upgrades warnings will be silenced.\n\n` +
-      `    Make sure you have manually checked all uses of unsafe flags.\n`,
-  );
-  silenced = true;
+  if (!silenced) {
+    console.error(
+      '\n' +
+        chalk.keyword('orange').bold('Warning: ') +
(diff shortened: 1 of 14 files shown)


packages/core/src/validate.ts
```diff
@@ -447,3 +447,15 @@ function* getEnumErrors(contractDef: ContractDefinition, decodeSrc: SrcDecoder):
     };
   }
 }
+
+let silenced = false;
+
+export function silenceWarnings(): void {
```
martriay, line 453: I think that this should forbid subsequent calls to prevent multiple logs, in line with the feature.

martriay (changes requested):

packages/core/src/validate.ts
```diff
@@ -447,3 +447,15 @@ function* getEnumErrors(contractDef: ContractDefinition, decodeSrc: SrcDecoder):
...
+  console.error(
+    '\n' +
+      chalk.keyword('orange').bold('Warning: ') +
+      `All subsequent Upgrades warnings will be silenced.\n\n` +
```
martriay, line 457: Maybe warning in here that this message shouldn't be seen outside testing envs?

martriay (changes requested):

packages/core/src/validate.ts
```diff
@@ -447,3 +447,15 @@ function* getEnumErrors(contractDef: ContractDefinition, decodeSrc: SrcDecoder):
     };
   }
 }
+
+let silenced = false;
```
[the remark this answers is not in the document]
frangio: I prefer to keep the variable next to the function that uses it.

packages/core/src/validate.ts
frangio: I changed this so that subsequent calls will not print the warning. I thought forbidding it with an error was too strong, and some people may want to use the function in different files.

packages/core/src/validate.ts
frangio: This was not a bad idea, but I decided not to add it in the message because I don't feel strongly about only using this in testing environments. I see it just as a recommendation.

PR #236, opened 2020-11-24
Clarify constant usage instead of Initial Values in Field Declarations

abcoathup:
Example in documentation to not use initial values in field declarations was actually a constant, causing confusion.

Based on feedback in community forum: https://forum.openzeppelin.com/t/initial-value-in-upgrade-contract/4688/5?u=abcoathup


Skyge:
Good job, the documentation is now much easier to understand.

frangio (approved):
Thanks!

PR #244, opened 2020-11-28
Fix #243

### LINKED ISSUE #243: waitAndValidateDeployment is broken on Hyperledger Besu
I use **openzeppelin-upgrades** with a Hyperledger Besu node in version `besu/v20.10.1/linux-x86_64/oracle_openjdk-java-11` running in a Docker container. Node configured in IBFT2.

The functions `deployProxy` and `upgradeProxy` do not work: they end with an error: `InvalidDeployment [Error]: Invalid deployment with address`.

The problem comes from this part of the code in `waitAndValidateDeployment` function:

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/56651304998c3b4a761fb832dc04f20e702b62d2/packages/core/src/deployment.ts#L68-L71

It seems that the Besu node does not return the `tx.blockHash` property before the transaction is included in a block. `tx.blockHash` is `undefined` before tx inclusion, not `null`.

By changing by 
```
if (tx.blockHash !== null && tx.blockHash !== undefined) {
...
}
```

It works properly.

adetante:
Fixes #243

### DIFF
--- a/packages/core/src/deployment.ts
+++ b/packages/core/src/deployment.ts
@@ -65,7 +65,7 @@ export async function waitAndValidateDeployment(provider: EthereumProvider, depl
       if (tx === null) {
         throw new InvalidDeployment(deployment);
       }
-      if (tx.blockHash !== null) {
+      if (tx.blockHash !== null && tx.blockHash !== undefined) {
         debug('succeeded verifying deployment tx mined', txHash);
         break;
       }


frangio (approved):
Thanks for debugging and fixing this!

frangio:
I've published a new version of `@openzeppelin/upgrades-core` including this fix.

adetante:
Thanks for the release!

PR #261, opened 2020-12-16
Automatic checking of storage compatibility of structs and enums

### LINKED ISSUE #95: Support structs & enums in storage layout checks
Checking storage layout that includes structs and enums is not yet supported. The initial version will contain an error that can be turned into a warning.

### LINKED ISSUE #152: Store struct and enum members in storage layout info
We are not storing struct and enum members in the storage layout for deployed impls. We should extract this information and store it as well, since it will be a prerequisite for https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/95.

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/608f133bca3500e8721c0867ed9a534e7353aa20/packages/core/src/storage.ts#L17-L19

### LINKED ISSUE #199: Check that enum representation doesn't change across upgrades
Enums are represented as the smallest integer type that can hold its numeric values. This not only affects internal representation (i.e. storage layout), but also ABI encoding, and can thus change function selectors if the enum is an argument.

As part of enum support (#95) we should check that an enum does not change its size when upgraded.

### LINKED ISSUE #248: Ignore struct errors originating in OpenZeppelin Contracts Upgradeable
Given that we ensure storage compatibility for OpenZeppelin Contracts Upgradeable (only within the same major version though!), we could ignore the struct errors that originate in it. In order to check that the major version is the same we would have to retrieve and store the package version in the network file, since we can't tell otherwise.

An example is ERC20Snapshot, as reported by a user [in the forum](https://forum.openzeppelin.com/t/error-deploying-erc20snapshotupgradeable/4864).

### LINKED ISSUE #249: Document how to check storage compatibility of structs
There are some additional rules that should be documented in the "Writing Upgradeable Contracts" guide, related to when it is safe o add fields to a struct. Note that this is different if the struct is in a mapping.

This content should also be linked from [the relevant FAQ entry](https://zpl.in/upgrades/error-007).

frangio:
Fixes #152 
Fixes #95 
Fixes #199
Resolves #248 
Resolves #249

- [x] ~Deprecation message for `unsafeAllowCustomTypes` flag.~ Kept for when struct members are missing from previous versions.
- [x] Deal with files (network files or validations.json) generated by previous plugin versions, since they are missing some new data we need.
- [x] Better diagnostic messages: explain why two structs are not considered equal, i.e. what members were removed/inserted/etc.
- [x] More tests.

Should this be considered a breaking change? I've currently made it so that an existing `unsafeAllowCustomTypes` flag will be ignored (and I'm additionally planning to make it print a deprecation message). If we assume that users were using this flag with truly compatible types (as intended) this it not a breaking change.


packages/core/src/utils/parse-type-id.ts
```diff
@@ -0,0 +1,104 @@
+import assert from 'assert';
+
+// This module parses type identifiers.
+
+export interface ParsedTypeId {
+  id: string;
+  head: string;
```
spalladino, line 7: Do you think we could type `head`, so we make sure that any checks that depend on this can be made exhaustive?

packages/core/src/utils/parse-type-id.test.ts
```diff
@@ -0,0 +1,20 @@
...
+import { parseTypeId } from './parse-type-id';
+
+const fixtures = [
+  't_uint256',
+  't_mapping(t_uint256,t_uint256)',
+  't_mapping(unknown,t_uint256)',
+  't_array(t_uint256)3_storage',
+  't_mapping(t_uint256,t_array(t_bool)dyn_storage)',
+  't_mapping(t_uint256,t_mapping(t_string_memory_ptr,t_address))',
+  't_function_internal_nonpayable(t_uint256)returns()',
```
spalladino, line 12: I opened up an old artifact (compiled with solc 0.5 probably) and I see they use $_ and _$ instead of ( and ). Are we translating that somewhere else?
```
"typeIdentifier": "t_function_internal_nonpayable$__$returns$__$",
```

packages/core/src/storage.ts
```diff
@@ -136,28 +168,171 @@ export function describeError(e: StorageOperation): string {
...
+  }
+
+  if (head === 't_enum') {
+    const originalMembers = original.item.members;
+    const updatedMembers = updated.item.members;
+    assert(originalMembers && isEnumMembers(originalMembers));
+    assert(updatedMembers && isEnumMembers(updatedMembers));
+    const ops = levenshtein(originalMembers, updatedMembers, (a, b) => (a === b ? 'equal' : 'replace'));
+    // it is only allowed to append new enum members
+    return ops.every(o => o.kind === 'append');
```
spalladino, line 301: Wasn't there an issue with a different underlying type being used for enums depending on how many possible values it has? Eg enums are backed by uint8s as long as they fit, but if you have too many members it may switch to a uint16. If that's the case, we should guard against appends that change the type of the backing field.

packages/core/src/utils/parse-type-id.test.ts
```diff
@@ -0,0 +1,20 @@
+import test from 'ava';
+
+import { parseTypeId } from './parse-type-id';
+
+const fixtures = [
```
spalladino, line 5: Can you add a test for enums?

packages/core/src/storage.ts
```diff
@@ -136,28 +168,171 @@ export function describeError(e: StorageOperation): string {
...
+  updated: ParsedTypeDetailed,
+  { canGrow }: { canGrow: boolean },
+): boolean {
+  if (original.head !== updated.head) {
+    return false;
+  }
+
+  if (original.args === undefined || updated.args === undefined) {
+    // both should be undefined at the same time
+    assert(original.args === updated.args);
```
spalladino, line 274: Can we add a message to each assertion? It will help debug user errors in the future.

packages/core/src/storage.ts
```diff
@@ -136,28 +168,171 @@ export function describeError(e: StorageOperation): string {
...
+
+function isEnumMembers<T>(members: TypeItemMembers<T>): members is EnumMember[] {
+  return members.length === 0 || typeof members[0] === 'string';
+}
+
+function isStructMembers<T>(members: TypeItemMembers<T>): members is StructMember<T>[] {
+  return members.length === 0 || typeof members[0] === 'object';
+}
+
+function matchStorageField(original: StorageField, updated: StorageField, { canGrow }: { canGrow: boolean }) {
```
spalladino, line 245: Is this ever called with `canGrow: true`?

packages/core/src/storage.ts
```diff
@@ -136,28 +168,171 @@ export function describeError(e: StorageOperation): string {
...
+  type: ParsedTypeDetailed;
+}
+
+// This function is generic because it can compare top level storage layout as well as struct layout.
+function getStorageUpgradeErrorsGeneric<T extends StorageField>(
+  original: T[],
+  updated: T[],
+  { canGrow }: { canGrow: boolean },
+): StorageOperation<T>[] {
+  const ops = levenshtein(original, updated, (a, b) => matchStorageField(a, b, { canGrow: false }));
```
spalladino, line 203: Shouldn't canGrow be true for the last item in the storage? Though I'm not sure if that can open the door to other kind of problems, and it's best to be conservative.

packages/core/src/utils/parse-type-id.test.ts
frangio: Yes! [`decodeTypeIdentifier`](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/a5fa60a691846080fd7ba84e9a5d6601d304902c/packages/core/src/storage.ts#L171)

packages/core/src/ast-dereferencer.ts
```diff
@@ -0,0 +1,42 @@
...
+        return cached as NodeTypeMap[T];
+      }
+    }
+
+    for (const ast of asts) {
+      for (const node of findAll(nodeTypes, ast)) {
+        if (node.id === id) {
+          cache.set(id, node);
+          return node;
+        }
```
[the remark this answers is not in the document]
frangio: It only happens in Truffle because we don't have the full solc output and we have to reconstruct it. But we try to detect partial compilation at the moment we reconstruct the solc output, and at this point the ASTs should be consistent.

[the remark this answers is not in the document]
frangio: The check is "best effort" though. [`checkImportIdConsistency`](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/a5fa60a691846080fd7ba84e9a5d6601d304902c/packages/plugin-truffle/src/validate.ts#L73)

Since the ASTDereferencer will validate the `nodeType` is as expected I'm not too worried about it.

packages/core/src/storage.ts
frangio: Yes, in fact newer Solidity versions will limit an enum to 256 elements to avoid this issue. But we should add this check for older versions. I would limit to 256 maximum elements to ensure compatbility the newer Solidity versions.

packages/core/src/storage.ts
frangio: Yeah technically the last item in storage can grow, I think it's safe and I'd love to support it. The issue is that `levenshtein` calls the match function without any indication that an element is the last one in the layout. We would have to change the levenshtein function to do that and I decided to go with the simpler conservative choice.

If people run into this limitation we will reconsider!

packages/core/src/storage.ts
frangio: No, good catch!

packages/core/src/utils/parse-type-id.ts
frangio: I prefer to keep `string` because Solidity could add new types in the future. Any sort of exhaustive check would be incorrect in that scenario, so it's better to force handling unknown types. In `compatibleTypes` any unknown types are conservatively considered incompatible.

packages/core/src/validate/error.ts
```diff
@@ -0,0 +1,72 @@
...
+    link: 'https://zpl.in/upgrades/error-005',
+  },
+  'external-library-linking': {
+    msg: e => `Linking external libraries like \`${e.name}\` is not yet supported`,
+    hint: () =>
+      `Use libraries with internal functions only, or skip this check with the \`unsafeAllowLinkedLibraries\` flag \n` +
+      `    if you have manually checked that the libraries are upgrade safe`,
+    link: 'https://zpl.in/upgrades/error-006',
+  },
+  'struct-definition': {
```
Amxx, line 45: Are these (struct-definition & enum-definition) still needed ? Should they be removed ?

frangio: These are needed because this will be a backwards compatible release, and in theory someone could be using an older Hardhat plugin with the newer Core. This could result in older cached validations to be passed to the core, and they can contain these errors. Those older versions of the Hardhat plugin will not invalidate the cache if it's an older version.

We could still detect it and throw an error, but I preferred to go this route for some reason. What do you think?

packages/core/src/utils/curry.ts
```diff
@@ -0,0 +1,17 @@
+export interface Curried<A, B, T> {
+  (a: A): (b: B) => T;
+  (a: A, b: B): T;
+}
+
+export function curry2<A, B, T>(fn: (a: A, b: B) => T): Curried<A, B, T> {
+  function curried(a: A): (b: B) => T;
+  function curried(a: A, b: B): T;
+  function curried(a: A, ...b: [] | [B]): T | ((b: B) => T) {
```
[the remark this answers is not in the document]
Amxx: So did I :)

Amxx (approved):

Amxx (approved):
I don't have a broad enough view to spot any subtle issues / inconsistency in the logic ... but overall, everything looks good at a local level.

davidlucid:
Thank you so much! @frangio

Issue #173, opened 2020-09-17
Provide an instance of ProxyAdmin in the admin module


frangio:
We allow people to transfer ownership of the `ProxyAdmin` but don't provide access to the admin address, except through the network manifest file.

We can provide an instance of a `ProxyAdmin` contract abstraction attached to the right address. Furthermore, we can configure this instance so that the default sender is the owner of the contract. This will be very useful for people.

This sounds like it should clearly go in the `admin` module, but I don't know what name we can give it.

abcoathup:
There could be an `admin.owner` or `admin.proxyAdminOwner` to expose the admin address (as there is `admin.transferProxyAdminOwnership`)
Then a `ProxyAdmin` could be used at that address to perform the upgrade.

frangio:
I think it will be better for users if we provide the `ProxyAdmin` instance directly. The address, and even the owner address, can be obtained from there.

frangio:
Fixed in #274.

PR #274, opened 2021-01-07
Add admin.getInstance

pinglamb:
This PR is for kicking off #173.

An `deployedProxyAdmin` method is added to `admin` module to get `ProxyAdmin` instance (feel free to suggest names). My use case is simply to get the address programmatically without `fs.readFileSync` the manifest.

### DIFF
--- a/packages/plugin-hardhat/CHANGELOG.md
+++ b/packages/plugin-hardhat/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Add `upgrades.admin.getInstance()` to retrieve the instance of `ProxyAdmin` that is in use. ([#274](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/274))
+
 ## 1.4.3 (2020-12-16)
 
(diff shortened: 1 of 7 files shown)


pinglamb:
@frangio Please feel free to suggest the proper naming and code structure. Thanks!

frangio:
Thanks @pinglamb! What is the reason for exposing it as a function (i.e. `upgrades.admin.deployedProxyAdmin()`) rather than just a getter (`upgrades.admin.instance`)?

pinglamb:
It seems that if it is an instance, once `hre.upgrades` is called it will try to load the `ProxyAdmin` object. For example, if you are using it for the first time, it will throw the `No ProxyAdmin ...` error. Also I am not sure if there will be other issues if we are caching the contract object with `hre.upgrades`. I guess the safest is loading it on demand? What do you think?

frangio:
Yes that makes sense. I think the function name `getInstance` sounds better to me. It's used as `upgrades.admin.getInstance()`.

We have to add the same function in the Truffle plugin. The code is very similar. Can you do these changes?

frangio (approved):
Thank you @pinglamb! I've changed the tests a little to give us more confidence that the return value is correct.

PR #286, opened 2021-01-20
Add new Defender Upgrades plugin for Hardhat

spalladino:
Adds a new `defender-upgrades` package that includes a hardhat plugin for proposing upgrades to Defender using the `defender-admin-client`. While still missing documentation and tests, I wanted to ask for an initial review, and discuss:

- By bundling the hardhat and truffle plugins into a single package, users would need to require `@openzeppelin/defender-upgrades/dist/hardhat` (and the equivalent for truffle). Are we ok with this, or should we set up separate packages for each? Or package the package (heh) differently so at least `dist` is not required?

- Do we have a preferred library for mocks that plays nice with Ava? I want to mock out `defender-admin-client` for writing the tests for `proposeUpgrade`, but it seems that [Ava doesn't ship with anything built-in for mocks](https://github.com/avajs/ava/issues/1829#issuecomment-578403538).


frangio:
For your first question, we can have a handwritten`hardhat.js` file at the top of the package with:

```javascript
module.exports = require('./dist/hardhat');
```

And the same for Truffle.

I always write mocks manually, so I don't have a preferred library. If it's too complex to write manually, using a simple library would be fine. Anything that is test framework agnostic would work. Sinon is the only one I know.

packages/core/src/manifest.ts
```diff
@@ -41,6 +42,7 @@ export class Manifest {
 
   constructor(chainId: number) {
```
frangio, line 43: The changes you made in this file can replaced by this change only:

```suggestion
  constructor(readonly chainId: number) {
```

(This is the way I'd do it.)

packages/plugin-defender/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "composite": true,
+    "outDir": "dist",
+    "rootDir": "src",
+    "resolveJsonModule": true,
+  },
+  "include": [
+    "src/**/*"
+  ],
+  "references": [
+    { "path": "../core" }
```
frangio, line 13: I think we need this because it's a dependency too.

```suggestion
    { "path": "../core" },
    { "path": "../plugin-hardhat" }
```

packages/plugin-defender/src/hardhat/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { ProposalResponse } from 'defender-admin-client';
+
+const BaseUrl = 'https://defender.openzeppelin.com';
+
+export function getProposalUrl(proposal: ProposalResponse): string {
+  return `${BaseUrl}/#/admin/contracts/${proposal.contractId}/proposals/${proposal.proposalId}`;
```
frangio, line 6: I don't remember if I mentioned this already. I would've expected the server to send this URL as part of the response, rather than us building it based on the parameters.

If the URL came in the response, I'd imagine this gives Defender more flexibility to change things independently of this package.

Another alternative would be to build the URL in the client library, although if I remember correctly the client is quite thin.

What do you think?

packages/plugin-defender/src/hardhat/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export interface HardhatDefenderConfig {
+  apiKey: string;
+  apiSecret: string;
+}
+
+extendConfig((config: HardhatConfig, userConfig: Readonly<HardhatUserConfig>) => {
+  if (!userConfig.defender) {
+    console.warn(
+      `Defender API key and secret are not set. Please add the following to your hardhat.config.js exported configuration:\n\n{\n  defender: {\n    apiKey: 'YOUR_API_KEY',\n    apiSecret: 'YOUR_API_SECRET'\n  }\n}\n`,
```
frangio, line 22: This can be a little nicer if we use `JSON.stringify(sampleConfig, null, 2)`.

Also, this is a matter of style but I've been avoiding "Please" in this kind of message, jumping straight to the indication.

packages/plugin-defender/package.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "dependencies": {
+    "defender-admin-client": "^1.3.0"
+  },
+  "peerDependencies": {
+    "@nomiclabs/hardhat-ethers": "^2.0.0",
+    "@openzeppelin/hardhat-upgrades": "^1.4.2",
+    "@openzeppelin/upgrades-core": "^1.4.2",
+    "hardhat": "^2.0.2"
+  }
```
frangio, line 39: As we discussed, we want to make `@openzeppelin/hardhat-upgrades` a full dependency of this package, and eventually have a different package for Truffle-Defender integration.

The other dependencies here don't need to be listed.

packages/plugin-defender/src/hardhat/propose-upgrade.ts
```diff
@@ -0,0 +1,34 @@
...
+  proxyAddress: string,
+  ImplFactory: ContractFactory,
+  opts?: ValidationOptions & { title?: string; description?: string },
+) => Promise<ProposalResponse & { url: string }>;
+
+export function makeProposeUpgrade(hre: HardhatRuntimeEnvironment): ProposeUpgradeFunction {
+  return async function proposeUpgrade(proxyAddress, ImplFactory, opts = {}) {
+    if (!hre.config.defender) {
+      throw new Error(`Missing Defender API key and secret in hardhat config`);
+    }
```
frangio, line 20: Isn't this already checked in `extendConfig`?

packages/plugin-defender/src/hardhat/propose-upgrade.ts
```diff
@@ -0,0 +1,34 @@
...
+    const client = new AdminClient(hre.config.defender);
+
+    const chainId = await getChainId(hre.network.provider);
+    const network = fromChainId(chainId);
+    if (!network) {
```
frangio, line 25:
```suggestion
    if (network === undefined) {
```

I generally avoid pure `!` because it only checks falsyness and can miss falsy values that are not `undefined` (e.g. empty string). I accept it for objects because all objects are truthy, but when dealing with strings I do the full comparison.

packages/plugin-defender/src/hardhat/propose-upgrade.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { ContractFactory } from 'ethers';
+import { FormatTypes } from 'ethers/lib/utils';
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+import { fromChainId } from './network';
+import { getProposalUrl } from './utils';
+
+export type ProposeUpgradeFunction = (
+  proxyAddress: string,
+  ImplFactory: ContractFactory,
+  opts?: ValidationOptions & { title?: string; description?: string },
```
frangio, line 13: Can you define an interface `ProposalOptions` with title and description?

frangio (commented):
Looks good!

I'm not sure if I want us to have an examples directory. I haven't had good experiences with examples in-repo in the past. I think it should just be a test, and examples should be in a guide.

packages/plugin-defender/src/hardhat/utils.ts
spalladino: Moved to client in https://github.com/OpenZeppelin/defender-client/pull/29. Will remove from here once merged.

packages/plugin-defender/src/hardhat/propose-upgrade.ts
spalladino: No, extendConfig just warns if the api key is not set. This way the user does not need to enter api key and secret if they run a buidler script that does not interact with defender.

spalladino:
@frangio I've addressed all comments, added tests, and docs. Ready for re-review!

spalladino:
I'm getting codecov errors though, for a module that no longer exists (`openzeppelin/defender-upgrades/dist/hardhat). Is there a codecov cache I need to clean?

frangio:
I will review this tomorrow, but the Codecov failure I'm seeing is just that the lines added by this PR are only 79% covered.

You can see the red lines [in this link](https://app.codecov.io/gh/OpenZeppelin/openzeppelin-upgrades/compare/286/diff). They are error conditions that we're not triggering.

It's sufficient coverage though, so don't worry about it. (We could change the failure thresholed for Codecov but it's sometimes useful to see it fail to look at the blind spots in our tests.)

packages/plugin-defender-hardhat/src/propose-upgrade.ts
```diff
@@ -0,0 +1,37 @@
+import '@openzeppelin/hardhat-upgrades/dist/type-extensions';
+import { getChainId, ValidationOptions } from '@openzeppelin/upgrades-core';
+import { AdminClient, ProposalResponse } from 'defender-admin-client';
+import type { ContractFactory } from 'ethers';
+import { FormatTypes } from 'ethers/lib/utils';
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+import { fromChainId } from 'defender-base-client';
```
frangio, line 7: Can you add `defender-base-client` as a dependency? I know it's a transitive dependency from the admin client, but (I think) this could fail depending on how npm or yarn (or whatever) lays out the dependencies on the file system if we don't declare the dependency explicitly.

packages/plugin-defender-hardhat/src/propose-upgrade.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    const chainId = await getChainId(hre.network.provider);
+    const network = fromChainId(chainId);
+    if (network === undefined) {
+      throw new Error(`Network ${chainId} is not supported in Defender Admin`);
+    }
+
+    const nextImpl = await hre.upgrades.prepareUpgrade(proxyAddress, ImplFactory);
+    const contract = { address: proxyAddress, network, abi: ImplFactory.interface.format(FormatTypes.json) as string };
+    return client.proposeUpgrade({ newImplementation: nextImpl, ...opts }, contract);
```
frangio, line 35: I'd avoid forwarding unrelated options to the client.

```suggestion
    const { title, description } = opts;
    const newImplementation = await hre.upgrades.prepareUpgrade(proxyAddress, ImplFactory);
    const contract = { address: proxyAddress, network, abi: ImplFactory.interface.format(FormatTypes.json) as string };
    return client.proposeUpgrade({ newImplementation, title, description }, contract);
```

frangio (commented):
Looks great! These are only nitpicks and small things.

I'm still not sold on the examples directory. Let's discuss this offline.

packages/plugin-defender-hardhat/test/propose-upgrade.js
```diff
@@ -0,0 +1,92 @@
...
+      fromChainId: () => t.context.fakeChainId,
+    },
+  }).makeProposeUpgrade(hardhat);
+
+  t.context.Greeter = await ethers.getContractFactory('Greeter');
+  t.context.GreeterV2 = await ethers.getContractFactory('GreeterV2');
+  t.context.greeter = await upgrades.deployProxy(t.context.Greeter);
+});
+
+test.afterEach(() => {
```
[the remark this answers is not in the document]
spalladino: Interesting, didn't know this about ava!

frangio (approved):
Looks good to me. Thank you!

PR #291, opened 2021-01-26
Include DeployOption skipAll for validation

### LINKED ISSUE #280: Add a manual override for every check
Requested in https://forum.openzeppelin.com/t/force-override-upgrade-validation/5368:

>When attempting to use openzeppelin-upgrades we are able to override only a particular subset of unsafe actions (namely unsafeAllowCustomTypes and unsafeAllowLinkedLibraries). However in some cases, one may want to override things like “Use of delegatecall is not allowed” (e.g. when delegate calls are only used to read contract storage).
>
>Would it be within the realm of possibility to include a manual override for all validations, or simply a forceSkipValidation feature?

bh2smith:
Addresses/closes #280

This PR introduces a new deployment option that allows the user to override validation entirely with a boolean flag `skipAll`. Both truffle and hardhat plugin have been changed accordingly and unit tests are introduced in both. 

Note that the tests in truffle plugin include new contracts that are explicitly tested (probably not as on point and irrelvant to the change) while the hard hat test is much more concise and directly to the point. Willing to do either, both are probably not necessary.

### DIFF
--- a/packages/core/src/validate.ts
+++ b/packages/core/src/validate.ts
@@ -58,12 +58,14 @@ interface ValidationErrorOpcode extends ValidationErrorBase {
 export interface ValidationOptions {
   unsafeAllowCustomTypes?: boolean;
   unsafeAllowLinkedLibraries?: boolean;
+  skipAll?: boolean;
 }
 
 export function withValidationDefaults(opts: ValidationOptions): Required<ValidationOptions> {
   return {
     unsafeAllowCustomTypes: opts.unsafeAllowCustomTypes ?? false,
     unsafeAllowLinkedLibraries: opts.unsafeAllowLinkedLibraries ?? false,
+    skipAll: opts.skipAll ?? false,
   };
 }
 
@@ -199,7 +201,7 @@ export function assertUpgradeSafe(validations: Validations, version: Version, op
   let errors = getErrors(validations, version);
   errors = processExceptions(contractName, errors, opts);
 
-  if (errors.length > 0) {
+  if (errors.length > 0 && !opts.skipAll) {
     throw new ValidationErrors(contractName, errors);
   }
 }
--- a/packages/plugin-hardhat/test/deploy-validation.js
+++ b/packages/plugin-hardhat/test/deploy-validation.js
@@ -10,3 +10,8 @@ test('invalid deployment', async t => {
   const { Invalid } = t.context;
   await t.throwsAsync(() => upgrades.deployProxy(Invalid), undefined, 'Contract `Invalid` is not upgrade safe');
 });
+
+test('deploys invalid deployment with skipAll = true', async t => {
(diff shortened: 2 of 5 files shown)


packages/plugin-truffle/test/contracts/InvalidUpgradable/Calculator.sol
```diff
@@ -0,0 +1,20 @@
...
+        calculatedResult = a + b;
+        assert(calculatedResult >= a);
+        
+        emit Add(a, b);
+        user = msg.sender;
+        
+        return calculatedResult;
+    }
+
+}
```
bh2smith, line 20:
```suggestion
}

```

packages/plugin-truffle/test/contracts/InvalidUpgradable/Machine.sol
```diff
@@ -0,0 +1,40 @@
...
+    }
+    
+    function addValuesWithCall(address calculator, uint256 a, uint256 b) public returns (uint256) {
+        (bool success, bytes memory result) = calculator.call(abi.encodeWithSignature("add(uint256,uint256)", a, b));
+        emit AddedValuesByCall(a, b, success);
+        return abi.decode(result, (uint256));
+    }
+}
+
+
```
bh2smith, line 40:
```suggestion
```

packages/plugin-truffle/test/contracts/InvalidUpgradable/Storage.sol
```diff
@@ -0,0 +1,13 @@
...
+    uint public val;
+    
+    constructor(uint v) public {
+        val = v;
+    }
+    
+    function setValue(uint v) public {
+        val = v;
+    }
+}
```
bh2smith, line 13:
```suggestion
}

```

packages/plugin-truffle/test/test/validation-override.js
```diff
@@ -0,0 +1,81 @@
+const assert = require('assert');
+const { constants, expectEvent, BN } = require('@openzeppelin/test-helpers');
+const upgrades = require('@openzeppelin/truffle-upgrades');
+
+const CalculatorFactory = artifacts.require('Calculator');
+const StorageFactory = artifacts.require('Storage');
+const MachineFactory = artifacts.require('Machine');
+const MachineFactoryV2 = artifacts.require('MachineV2');
+
+contract.only('DelegateCaller', function (accounts) {
```
bh2smith, line 10:
```suggestion
contract('DelegateCaller', function (accounts) {
```

frangio:
Like we discussed on Slack we want to provide the option to skip all validations but in a granular way as opposed to a blanket `skipAll` flag. The proposed interface would be an option `unsafeAllow: ['delegatecall', 'state-variable-immutable']`.

PR #293, opened 2021-02-08
Update wrap-provider.ts

cccssw:
fixed wrapProvider when providers like ledger-provider who used Web3ProviderEngine, which throws an error with 'Web3ProviderEngine.prototype.send'


frangio:
Thank you @cccssw. We need to update the `TruffleProvider` interface to include the optional `sendAsync` method.

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/bf789dc56cee12fb25b20dc44b004b1d80b1c003/packages/plugin-truffle/src/truffle.ts#L66-L71

Is there any documentation about this that you could link?

PR #315, opened 2021-03-17
Add UUPS support

Amxx:
Changelog
===

Core - Internal 
---
- Add a fetchOrDeployProxy that stored new proxies address + type in the manifest
- Update layout to version 3.2: add a list of proxy deployments (address + txHash + kind) to the manifest
- Add a mechanism to extract signatures out of FunctionDefinition (from the AST)
- Add a new error type 'no-public-upgrade-fn'
- Add a list of all public and external function signatures to the validation objects
- Add check for the presence of an upgradeTo(address) function
  - this test is disabled using `unsafeAllow: ['no-public-upgrade-fn']` which is automatically added by plugins when operating on a transparent proxy

Plugin - Public
---
- Add a new field to the DeploymentOptions interface `kind` which allows users to specify the proxy flavor they want to use. Accepted values are `auto` (default), `uups` and `transparent`.
  - When deploying a new contract, `auto` fallback to `transparent` for backward compatibility
  - when upgrading a contract, `auto` will autodetect the proxy type from the manifest and apply the corresponding checks
- Add warning message when:
  - Deploying a UUPS proxy (that doesn't uses the proxyadmin pattern) while a proxyadmin instance is registered in the manifest.
  - Some proxies are not affected by a proxyAdmin being transfered

Plugin - Internal
---
- Detect the presence of a proxyAdmin when doing upgrades. This adds support to transparent proxy that are not owned through a proxyAdmin.


packages/core/contracts/test/Validations.sol
```diff
@@ -160,3 +160,14 @@ contract UsesExplicitUnsafeExternalLibrary {
     UnsafeExternalLibrary.explode(x, y);
   }
 }
+
+contract HasInternalUpgrateToFunction {
```
frangio, line 164: Typo in "Upgrate", in all the contracts in this file.

packages/core/src/validate/run.ts
```diff
@@ -46,6 +52,29 @@ interface ValidationErrorOpcode extends ValidationErrorBase {
...
+function skipCheck<T>(error: string, node: T): boolean {
+  function isAllowed(error: string, documentation: undefined | string | StructuredDocumentation): boolean {
+    switch (typeof documentation) {
+      case 'undefined':
+        return false;
+
+      case 'string': {
+        const entries = documentation.split(' ');
+        return ['@custom:openzeppelin-upgrade-allow', error].every(requierement => entries.includes(requierement));
+      }
```
frangio, line 68: This doesn't look right!! We should only consider the arguments to a `@custom:openzeppelin-upgrade-allow` directive (not all of the words in the docstring), and we should throw an error if we don't recognize one of the arguments.

We need to properly parse the NatSpec. Look at `parseNatSpec` in solidity-docgen. It needs a small change because it's only looking for keywords that match the pattern `@\w+`, which is no longer the right pattern. See https://github.com/ethereum/solidity/issues/8802#issuecomment-788062522 for the allowed characters.

packages/core/src/validate/run.ts
```diff
@@ -46,6 +52,29 @@ interface ValidationErrorOpcode extends ValidationErrorBase {
   kind: 'delegatecall' | 'selfdestruct';
 }
 
+interface ValidationErrorUpgradeability extends ValidationErrorBase {
+  kind: 'no-public-upgrade-fn';
+}
+
+function skipCheck<T>(error: string, node: T): boolean {
```
frangio, line 59: `node` should have type `Node`. (`import { Node } from 'solidity-ast/node'`)

This is the type of an arbitrary node in the AST. There is a problem, though, in that not every node has a `documentation` field, so instead of casting to `any` you should use `if ('documentation' in node)`.

I'm trying to understand why you implemented this function in this way with an internal function and a recursive call. Are we using any contracts compiled with Solidity 0.5? I feel like we only support Solidity 0.6. It seems 0.5 is in the Hardhat config but I'm not sure why... I'm asking this because if I recall correctly from 0.6 onwards the `documentation` field is always an object with a `text` field ([`StructuredDocumetation`](https://solidity-ast.netlify.app/interfaces/structureddocumentation)).

packages/core/src/validate/run.ts
```diff
@@ -93,6 +123,9 @@ export function validate(solcOutput: SolcOutput, decodeSrc: SrcDecoder): Validat
         ];
 
         validation[contractDef.name].layout = extractStorageLayout(contractDef, decodeSrc, deref);
+        validation[contractDef.name].methods = [...findAll('FunctionDefinition', contractDef)]
+          .map(fnDef => getFunctionSignature(fnDef, deref))
+          .filter(Boolean) as string[];
```
frangio, line 128: I think this would be a more explicit way to implement this.

```suggestion
          .filter(fnDef => ['external', 'public'].includes(fnDef.visibility))
          .map(fnDef => getFunctionSignature(fnDef, deref));
```

`getFunctionSignature` should always return a value, not only for external/public functions.

frangio (commented):
Thank you!

Started reviewing the code. Leaving a couple of initial comments, will continue tomorrow.

packages/plugin-truffle/test/test/uups-linked-libraries.js
```diff
@@ -21,29 +21,32 @@ contract('Token with flag', function (accounts) {
...
   it('Redeploy Proxy again with different Library', async function () {
-    await deployProxy(Token, ['TKN', 10000], { unsafeAllowLinkedLibraries: true });
+    await deployProxy(Token, ['TKN', 10000], { unsafeAllow: ['external-library-linking'], kind: 'uups' });
 
     const safeMathLib2 = await SafeMathV2.deployed();
     Token.link('SafeMath', safeMathLib2.address);
-    const tokenNew = await deployProxy(Token, ['TKN', 5000], { unsafeAllowLinkedLibraries: true });
+    const tokenNew = await deployProxy(Token, ['TKN', 5000], {
+      unsafeAllow: ['external-library-linking'],
+      kind: 'uups',
```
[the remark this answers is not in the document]
Amxx: all (relevant) tests have been duplicated. You'll see there is a `uups-linked-libraries.js` and a `transparent-linked-libraries.js `

packages/plugin-truffle/test/test/uups-linked-libraries.js
```diff
@@ -2,16 +2,16 @@ const assert = require('assert');
...
-const TokenV2 = artifacts.require('TokenV2');
-const TokenV3 = artifacts.require('TokenV3');
+const Token = artifacts.require('TokenProxiable');
+const TokenV2 = artifacts.require('TokenV2Proxiable');
+const TokenV3 = artifacts.require('TokenV3Proxiable');
 
 contract('Token without flag', function () {
   it('deployProxy', async function () {
-    await assert.rejects(deployProxy(Token, ['TKN', 10000]));
+    await assert.rejects(deployProxy(Token, ['TKN', 10000], { kind: 'uups' }));
```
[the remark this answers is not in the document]
Amxx: we could. I prefered having this explicit, but I'd be of to making it implicit

packages/core/contracts/import.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+import "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
+import "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
+
+contract AdminUpgradeabilityProxy is TransparentUpgradeableProxy {
+    constructor(address _logic, address admin_, bytes memory _data) payable TransparentUpgradeableProxy(_logic, admin_, _data) {}
```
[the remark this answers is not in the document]
Amxx: admin is already defined, so we cannot name our argument admin. I put trailing _ in all parameters for consistency.

packages/core/src/validate/run.ts
Amxx: The main issue here is that solidity-ast doesn't support documentation everywhere. I opened a PR in the soldity ast repo that include a contract that compiles, but is not parsed correctly. Not sure we can do the validation/run part cleanly unless solidity-ast is fixed, and I haven't been able to fix it (I've tried)

packages/core/src/validate/run.ts
Amxx: right. I use to have a getFunctionSelector, but now that we use signature your version is better

packages/core/src/validate/run.ts
Amxx: I've updated it to avoid the any cast, but I still cannot assume node.documentation is null | StructuredDocumentation ... I still get strings :/

packages/core/src/storage/index.ts
```diff
@@ -52,7 +52,11 @@ export function getStorageUpgradeErrors(
   opts: ValidationOptions = {},
 ): StorageOperation<StorageItem>[] {
   try {
-    assertStorageUpgradeSafe(original, updated, opts.unsafeAllowCustomTypes);
+    assertStorageUpgradeSafe(
+      original,
+      updated,
+      (opts.unsafeAllow || []).includes('struct-definition') || (opts.unsafeAllow || []).includes('enum-definition'),
```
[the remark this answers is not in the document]
Amxx: I reenabled the previous option, and remove this alternative. unsafe upgrades require explicitelly enabling `unsafeAllowCustomTypes`. Is that ok?

Otherwize I'm not sure how it should be computed. would be need to enable both `struct-definition` AND `enum-definition` ? Should `storage/compare.ts` be updated to make the distinction between both ?

packages/core/src/validate/run.ts
Amxx: can you explain `/^(?:@(\w+) )?((?:(?!^@\w+ )[^])*)/m` ?

axic: @frangio please note the pattern in your link is not the final one: https://github.com/ethereum/solidity/issues/8802#issuecomment-806707040

`@custom:([a-z][a-z-]*)` is the correct one for custom tag names.

Amxx: @axic: thanks for the update!

@frangio What do you think of:
```
function* execall(re: RegExp, text: string) {
  re = new RegExp(re, re.flags + (re.sticky ? '' : 'y'));
  while (true) {
    const match = re.exec(text);
    if (match && match[0] !== '') {
      yield match;
    } else {
      break;
    }
  }
}

function skipCheck(error: string, node: Node): boolean {
  function isAllowed(error: string, documentation: StructuredDocumentation | string): boolean {
    switch (typeof documentation) {
      case 'string': {
        for (const { groups } of execall(/^(?:@(?<title>\w+)(?::(?<tag>[a-z][a-z-]*))? )?(?<args>(?:(?!^@\w+ )[^])*)/m, documentation)) {
          if (groups && groups.title === 'custom' && groups.tag === 'openzeppelin-upgrade-allow') {
            for (const [, arg] of execall(/(?<arg>[a-z][a-z-]*)\s?/,groups.args)) {
              if (arg === error) {
                return true;
              }
              if (![
                'state-variable-assignment',
                'state-variable-immutable',
                'external-library-linking',
                'struct-definition',
                'enum-definition',
                'constructor',
                'delegatecall',
                'selfdestruct',
                'no-public-upgrade-fn',
              ].includes(arg)) {
                throw new Error(`NatSpec: openzeppelin-upgrade-allow argument not recognized: ${arg}`);
              }
            }
          }
        }
        return false;
      }
      default:
        return isAllowed(error, documentation.text);
    }
  }
  if ('documentation' in node && node.documentation) {
    return isAllowed(error, node.documentation);
  } else {
    return false;
  }
}```

File: packages/core/src/validate/run.ts [hunk 9]
```diff
@@ -46,6 +53,61 @@ interface ValidationErrorOpcode extends ValidationErrorBase {
...
+    const match = re.exec(text);
+    if (match && match[0] !== '') {
+      yield match;
+    } else {
+      break;
+    }
+  }
+}
+
+function skipCheck(error: string, node: Node): boolean {
```
frangio, line 72: The logic in this function is good, but I think it needs a small refactoring, and can be made a lot simpler.

First I would define a function to get the array of args for our custom tag for a given node.
```typescript
function getAllowed(node: Node): string[]
```
The inner `execall` can be replaced by a simpler `return group.args.split(/\s+/);`

Once that's defined, `skipCheck` becomes a lot simpler.

I would also remove the inner `isAllowed` with the `switch` statement and recursion, I think it obscures things quite a bit. Though I'm not entirely sure why you're doing this, is it to support Solidity 0.5? Something like the following would be much better IMO:

```typescript
if ('documentation' in node) {
  const doc =
    typeof node.documentation === 'string'
      ? node.documentation
      : node.documentation.text;
```

packages/core/src/validate/run.ts
```diff
@@ -111,11 +177,13 @@ export function validate(solcOutput: SolcOutput, decodeSrc: SrcDecoder): Validat
 function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDecoder): Generator<ValidationError> {
   for (const fnDef of findAll('FunctionDefinition', contractDef)) {
     if (fnDef.kind === 'constructor' && ((fnDef.body?.statements.length ?? 0) > 0 || fnDef.modifiers.length > 0)) {
-      yield {
-        kind: 'constructor',
-        contract: contractDef.name,
-        src: decodeSrc(fnDef),
-      };
+      if (!skipCheck('constructor', contractDef) && !skipCheck('constructor', fnDef)) {
```
frangio, line 180: Was there a reason why you used the `prune` argument to `findAll` in some cases and not others?

frangio (commented):
Second round of comments. To be continued tomorrow.

packages/core/src/validate/run.ts
Amxx: Its not to support solidity 0.5 ... I believe anything before 0.8.1 doesn't include the natspec in the AST anyway. 
I believe the issue being solved here comes from solididty-ast not being 100% compatible with the latest AST. See https://github.com/OpenZeppelin/solidity-ast/pull/8

packages/core/src/validate/run.ts
Amxx: MY bad, I should have used prune here.

packages/core/src/validate/overrides.ts
```diff
@@ -5,67 +5,67 @@ import { ValidationError } from './run';
...
+  'struct-definition':
+    `    You are using the \`unsafeAllow.struct-definition\` flag to skip storage checks for structs.\n` +
+    `    Make sure you have manually checked the storage layout for incompatibilities.\n`,
+  'enum-definition':
+    `    You are using the \`unsafeAllow.enum-definition\` flag to skip storage checks for enums.\n` +
+    `    Make sure you have manually checked the storage layout for incompatibilities.\n`,
+  constructor: `    You are using the \`unsafeAllow.constructor\` flag.\n`,
+  delegatecall: `    You are using the \`unsafeAllow.delegatecall\` flag.\n`,
+  selfdestruct: `    You are using the \`unsafeAllow.selfdestruct\` flag.\n`,
+  'no-public-upgrade-fn': undefined,
```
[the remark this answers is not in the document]
Amxx: maybe, what do you think ?

Amxx (commented):
This used to be the case, but caused issues during manual tests. I'll have to check that again.

Details: cache of old implementation doesn't contain this field, which caused issues.

Amxx:
Manual tests look good

frangio:
>cache of old implementation doesn't contain this field

This is solved now because I bumped the validation data version so the cache will be flushed.

frangio (approved):
I've finished reviewing everything. It looks great.

We need to think about how we want to document this. We can do it in a separate PR.

PR #320, opened 2021-04-05
Add generic unsafeAllow option and granular NatSpec overrides

### LINKED ISSUE #280: Add a manual override for every check

### LINKED ISSUE #318: Granular "unsafeAllow" through NatSpec comments
The latest Solidity allows adding NatSpec comments to any statement. We can use this to disable or enable checks on a more granular level than we do now.

Amxx:
Fixes #318 
Fixes #280

Changelog
===

Core - Public
---

- Add a new field to the ValidationOptions interface unsafeAllow with can be used to silent any error type. For example, opts = { unsafeAllow: ['external-library-linking', 'delegatecall'] } will silent the corresponding checks.
- Custom natspec comments can disable check errors directly from the solidity code. See core/contracts/test/ValidationNatspec.sol for usage of these natspec comments.

Core - Internal
---
- Add a new error type 'inline-assembly'
- Add check for inline blocks in the ast (disabled for now)
- Fix a bug with library placeholder when generating smart contract versions.


packages/core/src/validate/run.ts
```diff
@@ -120,25 +175,32 @@ function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDe
...
       };
     }
   }
+  /// TODO: uncomment to enable assembly check
+  // for (const node of findAll('InlineAssembly', contractDef, node => skipCheck('inline-assembly', node))) {
+  //   yield {
+  //     kind: 'inline-assembly',
+  //     src: decodeSrc(node),
+  //   };
+  // }
```
frangio, line 203: Comment should be removed.

packages/core/contracts/test/Validations.sol
```diff
@@ -160,3 +160,11 @@ contract UsesExplicitUnsafeExternalLibrary {
     UnsafeExternalLibrary.explode(x, y);
   }
 }
+
+contract HasInlineAssembly {
```
frangio, line 164: More inline assembly stuff that should be removed I think.

packages/core/contracts/test/ValidationsNatspec.sol
```diff
@@ -0,0 +1,227 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.2;
+
+contract HasEmptyConstructorNatspec {
+  constructor() { }
+}
```
frangio, line 6: This one doesn't need any special comments so shouldn't be needed right?

packages/core/contracts/test/ValidationsNatspec.sol
```diff
@@ -0,0 +1,227 @@
...
+
+/// @custom:openzeppelin-upgrade-allow-unsafe state-variable-assignment
+contract HasStateVariableAssignmentNatspec1 {
+  uint x = 1;
+}
+
+contract HasStateVariableAssignmentNatspec2 {
+  /// @custom:openzeppelin-upgrade-allow-unsafe state-variable-assignment
+  uint x = 1;
+}
```
frangio, line 31: I'd like to see a test that something like the following fails:

```solidity
contract HasStateVariableAssignmentNatspec2 {
  /// @custom:openzeppelin-upgrade-allow-unsafe state-variable-assignment
  uint x = 1;

  uint y = 2;
}
```

Consider adding a few more tests like this that should fail to validate, for other overrides. No need to be exhaustive, just a few strategically chosen ones.

packages/core/package.json
```diff
@@ -26,7 +26,7 @@
   "devDependencies": {
     "@ava/typescript": "^1.1.1",
     "@nomiclabs/hardhat-etherscan": "^2.1.0",
-    "@openzeppelin/contracts": "^3.1.0",
+    "@openzeppelin/contracts": "^4.0.0",
```
frangio, line 29: This looks unrelated to this PR, along with all of the changes to the contracts in `core`.

packages/core/src/validate/overrides.ts
```diff
@@ -5,67 +5,68 @@ import { ValidationError } from './run';
...
-  contractName: string,
-  errorsToProcess: ValidationError[],
-  opts: ValidationOptions,
-): ValidationError[] {
-  const { unsafeAllowCustomTypes, unsafeAllowLinkedLibraries } = withValidationDefaults(opts);
-  let errors: ValidationError[] = errorsToProcess;
+export function withValidationDefaults(opts: ValidationOptions): Required<ValidationOptions> {
+  const unsafeAllowCustomTypes = opts.unsafeAllowCustomTypes ?? false;
+  const unsafeAllowLinkedLibraries = opts.unsafeAllowLinkedLibraries ?? false;
+  const unsafeAllow = opts.unsafeAllow ?? [];
```
frangio, line 32: I think this function should always return options that are consistent: `unsafeAllowCustomTypes` should be `true` if and only if `unsafeAllow.includes('struct-definition')` and `unsafeAllow.includes('enum-definition')`. Right now it's only ensuring a single direction (left to right).

An easy way to ensure the other direction is to set the initial values based on the array:
```suggestion
  const unsafeAllow = opts.unsafeAllow ?? [];
  const unsafeAllowCustomTypes =
    opts.unsafeAllowCustomTypes ||
    (unsafeAllow.includes('struct-definition') &&
      unsafeAllow.includes('enum-definition'));
  const unsafeAllowLinkedLibraries =
    opts.unsafeAllowLinkedLibraries ||
    unsafeAllow.includes('external-library-linking');
```

Just a suggestion, feel free to implement in a different way.

packages/core/src/version.ts
```diff
@@ -20,6 +20,15 @@ export function getVersion(bytecode: string, linkedBytecode?: string): Version {
 }
 
 export function hashBytecode(bytecode: string): string {
```
frangio, line 22: Although I think we should ship this fix as soon as possible, can we remove it from this PR so we keep this one about overrides only?

packages/core/src/validate/overrides.ts
```diff
@@ -5,67 +5,68 @@ import { ValidationError } from './run';
...
-    );
+      errors = errors.filter(error => {
+        const isException = errorType === error.kind;
+        exceptionsFound = exceptionsFound || isException;
+        return !isException;
+      });
+
+      if (exceptionsFound && !silenced && errorDescription) {
+        console.error(
+          '\n' +
```
frangio, line 63: I think I've removed these initial newlines before, I think it looks better without them.

frangio (commented):
Awesome stuff!

Please remove everything related to inline assembly (lots of things!), which I think we agreed to do later.

packages/core/src/version.ts
Amxx: The fix is needed for the tests to work correctly

frangio (approved):
LGTM!

frangio:
I remembered one last thing I wanted us to change. Let's make the NatSpec tag into `@custom:oz-upgrades-unsafe-allow`. (Note that "upgrades" is plural.)

And I think this is missing documentation as well. We need to be clear this only works with Solidity 0.8.

frangio (changes requested):
I remembered one last thing I wanted us to change. Let's make the NatSpec tag into @custom:oz-upgrades-unsafe-allow. (Note that "upgrades" is plural.)

And I think this is missing documentation as well. We need to be clear this only works with Solidity 0.8.

frangio:
There is a `CHANGELOG.md` file in each package. You can add a new `## Unreleased` section at the top with the changes introduced in this PR.

With changelog and documentation this is good to merge. After merging we should write up some more details about the bug that was fixed and assess whether anyone could have been affected by this and what they should do. It's a little complicated to think through this, we can look at it together on Friday.

Amxx:
Changelog and documentation added

frangio (approved):
LGTM!

bh2smith:
Awesome work here! Thanks for the changes, now this library is a whole lot more accessible!

PR #324, opened 2021-04-14
Ignore assign of immutable variable

frangio:
Assignments to immutable variables will not be considered errors. Note that immutable variables are still errors and currently require a manual override to allow them (`unsafeAllow: ['state-variable-immutable']` or `@custom:oz-upgrades-unsafe-allow state-variable-immutable`).

Not adding a changelog entry for this since it doesn't have a visible user effect (`unsafeAllow` hasn't been released yet).

### DIFF
--- a/packages/core/contracts/test/ValidationsNatspec.sol
+++ b/packages/core/contracts/test/ValidationsNatspec.sol
@@ -32,18 +32,18 @@ contract HasStateVariableAssignmentNatspec3 {
   uint y = 2;
 }
 
-/// @custom:oz-upgrades-unsafe-allow state-variable-immutable state-variable-assignment
+/// @custom:oz-upgrades-unsafe-allow state-variable-immutable
 contract HasImmutableStateVariableNatspec1 {
   uint immutable x = 1;
 }
 
 contract HasImmutableStateVariableNatspec2 {
-  /// @custom:oz-upgrades-unsafe-allow state-variable-immutable state-variable-assignment
+  /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
   uint immutable x = 1;
 }
 
 contract HasImmutableStateVariableNatspec3 {
-  /// @custom:oz-upgrades-unsafe-allow state-variable-immutable state-variable-assignment
+  /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
   uint immutable x = 1;
   uint immutable y = 2;
 }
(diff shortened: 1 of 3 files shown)


packages/core/contracts/test/ValidationsNatspec.sol
```diff
@@ -32,18 +32,18 @@ contract HasStateVariableAssignmentNatspec3 {
   uint y = 2;
 }
 
-/// @custom:oz-upgrades-unsafe-allow state-variable-immutable state-variable-assignment
+/// @custom:oz-upgrades-unsafe-allow state-variable-immutable
```
Amxx, line 35: shouldn't we remove the natspec completelly for immutable assignment.

`uint immutable x;` deserves an error, but `uint immutable x = 1;` should't require anything to be accepted

frangio: It's a little more complicated than that IMO. Your example is good, but I don't think we should blindly accept the following:

```
    uint immutable x = block.number;
```

Although this makes me think the error in this case is indeed the assignment, because the user should be aware that it will happen in the implementation and not the proxy. Maybe we should rename `state-variable-assignment` to `implementation-assignment` or something like that.

frangio: To be clear: I do want us to suport `uint immutable x = 1` but we'll need to make sure right hand side is a "pure" value.

Amxx: to me it is clear that `uint immutable x = block.number;` will take the block.number at construction of the implementation. I think this is the entire idea, and it should be authorized.

It's a matter of "who the users are" ... this is a more advanced usecase that require some expertise. You may argue that we need to warn novice user about this, but I think novice user won't go anywhere near immutable anyway.

frangio: I don't agree, I've seen people struggle to understand the interaction between immutables and proxies, and I definitely expect novice users to try to use them.

frangio:
I've changed my mind on this. I think we should keep the warning for now. I don't think it is clear to people what effect an immutable variable assignment has when proxies are involved.

"Unsafe" is not the right term here though, so we should consider using a different term for this in the future, and we should work towards educating people on how these things will behave with proxies.

Amxx:
>  and we should work towards educating people on how these things will behave with proxies.

Sounds like a workshop idea

PR #325, opened 2021-04-15
feat: Utilize generic to better support Typechain types

phated:
This adds a generic to the return result of the DeployFunction. This allows better integration with Typechain types, like so:

```ts
const WhitelistFactory = await hre.ethers.getContractFactory('Whitelist');

// Notice the `: Whitelist` here, which casts the Contract return to the Typechain
const whitelist: Whitelist = await hre.upgrades.deployProxy(WhitelistFactory, [args.address]);

await whitelist.deployed();

// We could then use custom RPC methods on Whitelist and have it fully typechecked here.
```

### DIFF
--- a/packages/plugin-hardhat/src/deploy-proxy.ts
+++ b/packages/plugin-hardhat/src/deploy-proxy.ts
@@ -16,8 +16,8 @@ import { readValidations } from './validations';
 import { deploy } from './utils/deploy';
 
 export interface DeployFunction {
-  (ImplFactory: ContractFactory, args?: unknown[], opts?: DeployOptions): Promise<Contract>;
-  (ImplFactory: ContractFactory, opts?: DeployOptions): Promise<Contract>;
+  <C extends Contract>(ImplFactory: ContractFactory, args?: unknown[], opts?: DeployOptions): Promise<C>;
+  <C extends Contract>(ImplFactory: ContractFactory, opts?: DeployOptions): Promise<C>;
 }
 
 export interface DeployOptions extends ValidationOptions {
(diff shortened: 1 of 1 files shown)


frangio:
Thanks for the suggestion @phated! This is a great idea.

Are you sure that this will achieve the desired outcome though? I would imagine the return value should depend on the type of the factory argument.

phated:
I linked it into the project I'm helping on and it seemed to make the typechecker happy, but I could also see the need to cast the return value using `as C` (I had to do that for ethers.js but TS didn't complain here) 🤔

frangio:
@phated I've pushed in your branch what I believe will give the right results. Can you try it out?

phated:
@frangio Hmm, I just linked that into my project and it doesn't work. The type inferred is still `Contract`. What else can we try?

frimoldi:
hey @frangio @phated , I was looking for this as well. The use of generic types looks good to me, but I think you also need to refactor the `makeDeployProxy` function, which is used to inject `upgrades` into the `hre` object.

frangio:
@frimoldi That would be awesome. Can you somehow test it? If you can also submit a PR that would be great.

frimoldi:
@frangio yep, I will. Wanted to see if there were interest still. 👍🏽

frimoldi:
@frangio I've just created a new PR here https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/535

phated:
Closing in favor of #535 

We no longer use this plugin, so I don't have a need for this anymore.

PR #349, opened 2021-05-04
Accept contract instance as well as address for upgradeProxy

frangio:
`upgradeProxy(contract.address, ContractV2)`

becomes

`upgradeProxy(contract, ContractV2)`

But the previous behavior is kept for backwards compatibility.


Amxx (approved):
LGTM

PR #350, opened 2021-05-04
Remove use of chalk.keyword

logvik:
Keyword has been removed. Need to use direct way.
https://github.com/chalk/chalk/issues/431


frangio (approved):
As far as I can tell it hasn't been removed yet, but I'm happy to make this change as soon as possible regardless.

Thank you!

frangio (changes requested):
Actually, there's a few other instances of `chalk.keyword`. Can you update those as well?

logvik:
Done. All "keywords" were removed.

frangio (approved):
Thank you!

PR #376, opened 2021-06-11
Add UUPS related options in hardhat proposeUpgrade reference

spalladino:


frangio:
There seems to also be a `multisigType` option, was that left out of the docs on purpose?

spalladino:
Yep, it's not actually required. Defender can infer it on its own, and it's possible that we'll remove it for simplicity in the future.

PR #399, opened 2021-07-29
Handle slow eth_getCode providers

spalladino:
If eth_getCode fails to return the code immediately after the tx is mined (which can happen when there is a reorg, or if the provider hits a different node behind a load balancer that has not received that tx yet, or on moonbase), the plugin would return invalid deployment. This change adds a 1-min polling to allow the provider to catch up.

### DIFF
--- a/packages/core/src/deployment.ts
+++ b/packages/core/src/deployment.ts
@@ -85,8 +85,8 @@ export async function waitAndValidateDeployment(provider: EthereumProvider, depl
   }
 
   debug('verifying code in target address', address);
+  const startTime = Date.now();
   while (!(await hasCode(provider, address))) {
-    const startTime = Date.now();
     const elapsedTime = Date.now() - startTime;
     if (elapsedTime >= pollTimeout || txHash === undefined) {
       throw new InvalidDeployment(deployment);
--- a/packages/core/src/deployment.ts
+++ b/packages/core/src/deployment.ts
@@ -2,7 +2,14 @@ import { promisify } from 'util';
 
 import debug from './utils/debug';
 import { makeNonEnumerable } from './utils/make-non-enumerable';
-import { EthereumProvider, getTransactionByHash, hasCode, isDevelopmentNetwork } from './provider';
+import {
+  EthereumProvider,
+  getTransactionByHash,
+  getTransactionReceipt,
+  hasCode,
+  isDevelopmentNetwork,
+  isReceiptSuccessful,
+} from './provider';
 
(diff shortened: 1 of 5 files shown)


packages/core/src/deployment.ts
```diff
@@ -73,11 +74,15 @@ export async function waitAndValidateDeployment(provider: EthereumProvider, depl
   }
 
   debug('succeeded verifying deployment', txHash);
-  if (await hasCode(provider, address)) {
-    return;
+  while (!(await hasCode(provider, address))) {
+    const startTime = Date.now();
+    const elapsedTime = Date.now() - startTime;
+    // We only poll for code if we are in the context of a tx
```
frangio, line 80: I don't understand this comment. What do you mean by being in the context of a tx?

packages/core/src/deployment.ts
```diff
@@ -73,11 +74,15 @@ export async function waitAndValidateDeployment(provider: EthereumProvider, depl
   }
 
   debug('succeeded verifying deployment', txHash);
-  if (await hasCode(provider, address)) {
-    return;
+  while (!(await hasCode(provider, address))) {
```
frangio, line 77: My only concern with this PR is that the original reason we check `hasCode` is that if the transaction was mined but `hasCode` returns empty it was because the transaction failed. This PR seems to add a 1 minute delay to detect a failed transaction. Am I understanding that correctly?

frangio (approved):

packages/core/src/deployment.ts
spalladino: I understand that, if there is no txHash, it means we are just checking an address, and not a transaction that was just run. If so, it doesn't make sense to wait for getCode to resolve, since the tx that deployed that code should have landed long ago.

I may be mistaken though! Let me know if so and I can change this condition.

packages/core/src/deployment.ts
spalladino: Makes sense. How about I add a check for the tx status in the receipt, and only await this 1 minute if the tx was flagged as successful?

packages/core/src/deployment.ts
```diff
@@ -73,11 +74,15 @@ export async function waitAndValidateDeployment(provider: EthereumProvider, depl
...
+    const startTime = Date.now();
```
frangio, line 89: `startTime` should be outside the loop!

```suggestion
  const startTime = Date.now();
  while (!(await hasCode(provider, address))) {
```

frangio (approved):

PR #400, opened 2021-07-30
Bump @truffle/contract version

cds-amal:
This PR updates the @truffle/contract dependency in order to work with newer truffle versions.

See [this issue for details](https://github.com/trufflesuite/truffle/issues/4183#issuecomment-888287945)

### DIFF
--- a/packages/plugin-truffle/package.json
+++ b/packages/plugin-truffle/package.json
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "@openzeppelin/upgrades-core": "^1.8.0",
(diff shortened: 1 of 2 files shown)


frangio (approved):
Sorry for the delay in taking a look at this.

I'm not sure this is the right place to make this fix. It really sounds more like a Truffle bug. But I do think that requiring a more recent @truffle/contract version in our package will result in less people running into this Truffle issue so I agree with this change.

frangio:
Released in @openzeppelin/truffle-upgrades@1.8.1.

PR #407, opened 2021-08-06
Support function objects in arguments

### LINKED ISSUE #405: Error: Unsuported TypeName node type: FunctionTypeName at serializeTypeName (node_modules/@openzeppelin/upgrades-core/src/utils/function.ts:43:13)
Compiling a big project and this error came up, can't trim down the error but i think it relates to me using function type in a library:

```solidity
	function updateVault(
		Vault.Data memory vaultData,
		address TAddr,
		function(
			address,
			uint256,
			uint256,
			uint256,
			uint256,
			uint256
		) external onVaultUpdate
	) internal {
}

```

This happens on 0.8.4 solidity and `"@openzeppelin/hardhat-upgrades": "^1.9.0"`

Amxx:
Fix #405 

Function signature reconstruction did not support functions in the parameters, which caused errors to be thrown.

### DIFF
--- a/packages/core/contracts/test/FunctionSignatures.sol
+++ b/packages/core/contracts/test/FunctionSignatures.sol
@@ -39,6 +39,14 @@ contract ContractFunctionSignatures {
     function f11(bytes32) public {}
 
     function f12(ContractFunctionSignatures) public {}
+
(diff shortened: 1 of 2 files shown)


frangio (approved):
Thank you!

Issue #276, opened 2021-01-08
Support storage layout gaps


frangio:
We often recommend including a storage "gap" in storage layout in order to safely add storage variables when using (multiple) inheritance. We use this pattern ourselves in OpenZeppelin Contracts Upgradeable. However, if a slot in the gap is actually used for a new variable later, the plugins will flag this as an error. We should detect gaps through the conventional `____gap` name and treat them specially for layout compatibility: allow replacing gap slots with new variables in an upgrade.

Amxx:
@aquiladev unfortunately there isn't. I'll see if I can bump it, as I believe it would be valuable ... but I don't want to make any promise I can't keep.

itsermin:
> @aquiladev unfortunately there isn't. I'll see if I can bump it, as I believe it would be valuable ... but I don't want to make any promise I can't keep.

This is quite limiting unfortunately for automating upgrades

aquiladev:
@itsermin so far I see next options:
- [Hack storage layout history](https://forum.openzeppelin.com/t/storage-layout-upgrade-with-hardhat-upgrades/14567/2?u=aquiladev) in order to run upgrade (the option hard to automate)
- Use [storage slots](https://docs.openzeppelin.com/contracts/4.x/api/utils#StorageSlot), in order to avoid storage layout change
- Wait for `_gap` support in the tool (change storage layout only through append)

roshanr95:
It would be great if you could _prefix_ match `____gap` instead of exact. Allows for adding multiple gaps as well as giving them more meaningful names.

sullof:
Solving this issue is quite important.

First, you suggest an approach that does not work as intended.

Second, the entire assumption is that you can write upgradeable contracts, but if the storage cannot be upgraded with new variables, the entire castle crashes, forcing to move towards other solutions for upgradeability.

Long time is passed since the issue has been open. Is there any news?

ericglau:
@sullof Thank you for your feedback.  We have recently started working on this.

sullof:
BTW, I wonder if you are thinking of supporting this:
https://github.com/GeniusVentures/openzeppelin-contracts-diamond
I made some experiments and it works quite well, solving the issue with the changes in the storage.
Still not perfect, but much better.

ericglau:
[...]

Issue #3, opened 2020-07-06
Use storage layout from solc output for more sophisticated storage checks


frangio:
Solidity 0.5.13 introduced the `storageLayout` compiler output, which gives accurate information about the size and alignment of items in a storage layout. We should use this information instead of the AST. Given this accurate information, the use of the Levenshtein algorithm may have to change quite a bit.

frangio:
This is currently not a priority as the only thing it will give us is more sophisticated checks, in the sense that we are currently disallowing some things which would actually be ok. This is because our current algorithm does not consider the fact that multiple variables can occupy the same slot depending on their size and alignment.

If you have a requirement for these more sophisticated checks let us know. If someone wants to take on this issue and send a PR, that would also be fine, though let us know so we can discuss the details.

PR #423, opened 2021-09-03
Compare layouts including storage reservations

aquiladev:
https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/276

Draft storage layout gaps support.

Basic comparison of storage layout including reservations(gaps)

TODO:
- [ ] Handle field sizes for all types. Any ideas how to effectively calculate field size?
- [ ] Handle `allow...` parameters
- [ ] Change way of error handling
- [ ] Merge results from both comparers
- [ ] Add tests

### DIFF
--- a/packages/core/contracts/test/Storage.sol
+++ b/packages/core/contracts/test/Storage.sol
@@ -384,3 +384,24 @@ contract StorageUpgrade_MappingEnumKey_V2_Bad {
     mapping (E1 => uint) m3;
     mapping (uint => uint) m4;
 }
+
+contract StorageUpgrade_Gap_Base {
+    uint public b1;
+    uint[50] private __gap;
+}
+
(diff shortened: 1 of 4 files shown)


frangio:
Hi @aquiladev, Sorry we didn't reply earlier. Are you still interested in working on this?

> Handle field sizes for all types. Any ideas how to effectively calculate field size?

We should use the storage layout as returned by the Solidity compiler rather than extracted by us from the AST, because it contains all of that information: https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/3. This would only be supported for Hardhat because it allows us to request the `storageLayout` output.

frangio:
We now extract and store all information about the storage layout including slot and offset as calculated by solc itself, which was the building block we were missing to safely implement layout comparison ignoring gaps/reservations.

The approach in this PR would likely need to be thoroughly changed.

PR #433, opened 2021-10-06
Support constructorArgs for implementation

Amxx:
Sometime its needed to pass argument to an implementation in other to set immutable state variables (which will be shared by all all proxies). 

- [x] hardhat-plugin
- [ ] truffle-plugin
- [ ] documentation

### DIFF
--- a/packages/plugin-hardhat/src/deploy-proxy.ts
+++ b/packages/plugin-hardhat/src/deploy-proxy.ts
@@ -25,7 +25,8 @@ export interface DeployFunction {
 }
 
 export interface DeployOptions extends ValidationOptions {
-  initializer?: string | false;
+  initializer?:     string | false;
+  constructorArgs?: unknown[];
 }
 
 export function makeDeployProxy(hre: HardhatRuntimeEnvironment): DeployFunction {
@@ -54,7 +55,7 @@ export function makeDeployProxy(hre: HardhatRuntimeEnvironment): DeployFunction 
       }
     }
 
-    const impl = await deployImpl(hre, ImplFactory, requiredOpts);
+    const impl = await deployImpl(hre, ImplFactory, requiredOpts, undefined, opts.constructorArgs);
     const data = getInitializerData(ImplFactory, args, opts.initializer);
 
     let proxyDeployment: Required<ProxyDeployment & DeployTransaction>;
--- a/packages/plugin-hardhat/src/upgrade-proxy.ts
+++ b/packages/plugin-hardhat/src/upgrade-proxy.ts
@@ -3,7 +3,6 @@ import type { ethers, ContractFactory, Contract, Signer } from 'ethers';
 
 import {
   Manifest,
-  ValidationOptions,
   getAdminAddress,
   withValidationDefaults,
   setProxyKind,
@@ -18,14 +17,16 @@ import {
   ContractAddressOrInstance,
 } from './utils';
 
+import { DeployOptions } from './deploy-proxy';
+
 export type UpgradeFunction = (
   proxy: ContractAddressOrInstance,
   ImplFactory: ContractFactory,
-  opts?: ValidationOptions,
+  opts?: DeployOptions,
 ) => Promise<Contract>;
 
 export function makeUpgradeProxy(hre: HardhatRuntimeEnvironment): UpgradeFunction {
-  return async function upgradeProxy(proxy, ImplFactory, opts: ValidationOptions = {}) {
+  return async function upgradeProxy(proxy, ImplFactory, opts: DeployOptions = {}) {
     const { provider } = hre.network;
 
     const proxyAddress = getContractAddress(proxy);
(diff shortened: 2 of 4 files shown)


packages/plugin-hardhat/src/upgrade-proxy.ts
```diff
@@ -18,22 +11,30 @@ import {
   ContractAddressOrInstance,
 } from './utils';
 
+import { DeployOptions } from './deploy-proxy';
+
 export type UpgradeFunction = (
   proxy: ContractAddressOrInstance,
   ImplFactory: ContractFactory,
-  opts?: ValidationOptions,
+  opts?: DeployOptions,
```
frangio, line 19: The reason this was using `ValidationOptions` is that the `initializer` option is not used by `upgradeProxy`.

We should do something like this:

```typescript
// common-options.ts ?
interface CommonOptions extends ValidationOptions {
   constructorArgs?: unknown[];
}

function withCommonDefaults(opts: CommonOptions): Required<CommonOptions> {
  ...
}

// deploy-proxy.ts
interface DeployOptions extends CommonOptions {
  initializer?: string | false;
}
```

etc

packages/plugin-hardhat/src/utils/deploy-impl.ts
```diff
@@ -22,6 +22,7 @@ export async function deployImpl(
   ImplFactory: ContractFactory,
   requiredOpts: Required<ValidationOptions>,
```
frangio, line 23: This can use `CommonOptions` as defined above.

packages/plugin-hardhat/src/deploy-proxy.ts
```diff
@@ -54,7 +45,7 @@ export function makeDeployProxy(hre: HardhatRuntimeEnvironment): DeployFunction
       }
     }
 
-    const impl = await deployImpl(hre, ImplFactory, requiredOpts);
+    const impl = await deployImpl(hre, ImplFactory, requiredOpts, undefined);
```
frangio, line 48:
```suggestion
    const impl = await deployImpl(hre, ImplFactory, requiredOpts);
```

packages/plugin-hardhat/test/000_initial.js
```diff
@@ -0,0 +1,7 @@
+const test = require('ava');
+
+const { upgrades } = require('hardhat');
+
+test('start with empty network manifest', async t => {
+  await t.throwsAsync(upgrades.admin.getInstance(), undefined, 'No ProxyAdmin was found in the network manifest');
```
frangio, line 6: Why was this necessary?

packages/plugin-hardhat/test/constructor.js
```diff
@@ -0,0 +1,55 @@
...
+test('new proxy - do not redeploy with same args', async t => {
+  const { WithConstructor } = t.context;
+
+  const proxy1 = await upgrades.deployProxy(WithConstructor, { constructorArgs: [17] });
+  const implementation1 = await upgrades.erc1967.getImplementationAddress(proxy1.address);
+
+  const proxy2 = await upgrades.deployProxy(WithConstructor, { constructorArgs: [17] });
+  const implementation2 = await upgrades.erc1967.getImplementationAddress(proxy2.address);
+
+  t.is(implementation1, implementation2);
```
frangio, line 18: It would be nice to add an observational assertion about `value()` in each of the proxies.

packages/plugin-truffle/src/utils/deploy-impl.ts
```diff
@@ -40,7 +41,16 @@ export async function deployImpl(
...
-    const deployment = await deploy(requiredOpts.deployer, Contract);
+    const deployment = await deploy(requiredOpts.deployer, Contract, ...requiredOpts.constructorArgs);
     return { ...deployment, layout };
   });
 }
+
+function encodeArgs(Contract: ContractClass, constructorArgs: unknown[]): string {
+  const fragment = (Contract as any).abi.find((entry: any) => entry.type == 'constructor');
+
+  return (Contract as any).web3.eth.abi.encodeParameters(
```
frangio, line 52: There should be a global `web3` instance. Do you know if this works?

```suggestion
  return web3.eth.abi.encodeParameters(
```

packages/plugin-truffle/test/contracts/Constructor.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.5.0;
+
+contract WithConstructor {
```
frangio, line 4: Please use the same contract that we use to test Hardhat. Eventually we should move to a shared set of contracts and using different contracts will make that a little more difficult.

packages/plugin-truffle/src/utils/deploy-impl.ts
Amxx: Cannot find name 'web3'

packages/plugin-truffle/test/contracts/Constructor.sol
Amxx: I'd love to ... but truffle doesn't support multiple version of soldidty used in the same repo

... the other contract are using 0.5 ...
... with 0.5 we cannot put natspec or immutable ...

... truffle is really not something I enjoy working with

packages/core/src/version.ts
```diff
@@ -7,36 +7,46 @@ export interface Version {
...
   if (!/^(0x)?[0-9a-fA-F]*$/.test(bytecode)) {
     throw new Error('Bytecode is not a valid hex string');
   }
 
-  const buf = Buffer.from(bytecode.replace(/^0x/, ''), 'hex');
+  if (!/^(0x)?[0-9a-fA-F]*$/.test(constructorArgs)) {
+    throw new Error('Constructor arguments are not a valid hex string');
+  }
+
+  const buf = Buffer.concat([
```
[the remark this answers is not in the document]
frangio: Ok on second thought I don't think we should be concerned. The concatenation of bytecode + args is exactly what is sent with the deploy transaction. If the result of concatenating (B1, A1) and (B2, A2) is the same C creation code, then it seems we can safely assume it will behave exactly the same.

Issue #342, opened 2021-04-23
Support beacon proxies


frangio:
Given that OpenZeppelin Contracts contains beacon proxies, it would be expected that the Upgrades Plugins should support them as well.

It is not clear if this should take the form of a `kind: 'beacon'` option to `deployProxy`, or if it should be a new family of functions: `deployBeacon` and `upgradeBeacon`. It seems that people interested in beacons will exclusively be those using on chain factories of upgradeable contracts, so `deployProxy` might not really be the right abstraction for them, although it would certainly be useful for testing, so we might want both.

bundle-boy:
Nearing two months since this has been published. Any updates here?

Also wondering why beacon proxies aren't supported in the upgrades plugin; it does seem like a fairly niche use-case, but would be a nice quality of life improvement to be able to use the plugin on these.

frangio:
@bundle-boy We agree it would be awesome to support beacon proxies here and hope we can work on this in the next few weeks.

Can you comment on the question of `deployBeacon` vs `deployProxy`? I'd love to hear how you would expect to use the plugin in the context of beacon proxies so we can have more clarity on the direction we should take.

ylv-io:
I want to share my experience using beacon proxies for development and deployment. We use on-chain factory to deploy beacon proxies so we don't need `deployBeacon` functionality at all. Even tests use factory class to test for upgradability. We do use method `upgradeBeacon` to upgrade our beacon proxies via Hardhat script which is quite handy. Unfortunately, right now script is not capable to verify storage and other checks due to using `hardhat-deploy` for deployment so we are doing line by line peer review in case of upgrade.
I would say having an ability to run checks before calling an upgrade method is the biggest value proposition. Writing a script to deploy or upgrade a beacon proxy is trivial. Checking for upgradability compatibility is not. Especially in case of using different deployment solutions. Just using `artifacts` folder as a place to store deployed artifacts is not viable for most projects.

frangio:
A note for people subscribed: we have one last simplification pending before we release this but it should not be long.

leonwilly:
Any update on when this will be officially supported? I see #491 is pending review, is this the last simplification before beacon upgrades are put in production?

frangio:
@leonwilly Yes, will be done this week.

PR #456, opened 2021-11-08
Support beacon proxies for Hardhat

ericglau:
Part of https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/342

Adds the following APIs to the Hardhat plugin (see [doc](https://github.com/ericglau/openzeppelin-upgrades/blob/beacon/docs/modules/ROOT/pages/api-hardhat-upgrades.adoc)):
`deployBeacon`
`upgradeBeacon`
`deployBeaconProxy`
`loadProxy` (supports UUPS/Transparent/Beacon proxies)

Modifies the following API:
`prepareUpgrade` (accepts a UUPS/Transparent/Beacon proxy address or a Beacon address)


packages/plugin-hardhat/src/utils/deploy-impl.ts
```diff
@@ -14,53 +14,152 @@ import {
...
+}
+
+async function deployImpl(
   hre: HardhatRuntimeEnvironment,
   ImplFactory: ContractFactory,
   opts: Options,
-  proxyAddress?: string,
-): Promise<DeployedImpl> {
+  isBeacon: boolean,
+  proxyOrBeaconAddress?: string,
```
frangio, line 63: I realize now that we can do something simpler and more focused in `deployImpl`, and I think we should.

It seems the only thing that `isBeacon` affects is how we retrieve the implementation address. So my proposal is to move that logic outside of here, and instead of receiving the address of a proxy or a beacon as an argument, receive the implementation address directly. Doing that, `deployImpl` should be a lot simpler. The `kind` return argument can be ignored for beacon proxies.

Let me know what you think!

packages/plugin-hardhat/src/utils/impl-address.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Gets the implementation interface from the manifest.
+ *
+ * @returns a Promise with the interface, or undefined the implementation interface cannot be found in the manifest.
+ */
+export async function getInterfaceFromManifest(
+  hre: HardhatRuntimeEnvironment,
+  implAddress: string,
+): Promise<utils.Interface | undefined> {
+  if (implAddress === undefined) {
```
frangio, line 61: This shouldn't happen according to the type of the argument `implAddress: string`.

packages/plugin-hardhat/src/utils/impl-address.ts
```diff
@@ -0,0 +1,79 @@
...
+  } catch (e: any) {
+    try {
+      const beaconAddress = await getBeaconAddress(provider, proxyAddress);
+      result = await getImplementationAddressFromBeacon(
+        hre,
+        proxy instanceof Contract ? proxy.signer : signer,
+        beaconAddress,
+      );
+    } catch (e: any) {
+      // error expected if the address was not a beacon proxy
```
frangio, line 46: Same as a previous comment, we shouldn't ignore the error if it's a network error or something else.

packages/plugin-hardhat/src/load-proxy.ts
```diff
@@ -0,0 +1,39 @@
...
+    const proxyAddress = getContractAddress(proxy);
+
+    const implAddress = await getImplementationAddressFromProxy(provider, proxyAddress, hre, proxy, signer);
+    if (implAddress === undefined) {
+      throw new Error(`Contract at address ${proxyAddress} doesn't look like an ERC 1967 proxy or beacon proxy.`);
+    }
+
+    const contractInterface = await getInterfaceFromManifest(hre, implAddress);
+    if (contractInterface === undefined) {
+      throw new Error(
```
frangio, line 29: As mentioned previously `UpgradesError` can create a nicer error message here. The error message right now is super long, will wrap, etc., I generally try to make errors look nicer in the console, because I feel otherwise people may ignore them or just be annoyed.

packages/plugin-hardhat/src/deploy-proxy.ts
```diff
@@ -71,25 +78,29 @@ export function makeDeployProxy(hre: HardhatRuntimeEnvironment): DeployFunction
     inst.deployTransaction = proxyDeployment.deployTransaction;
     return inst;
   };
+}
 
-  function getInitializerData(ImplFactory: ContractFactory, args: unknown[], initializer?: string | false): string {
-    if (initializer === false) {
-      return '0x';
-    }
+export function getInitializerData(
```
frangio, line 83: Can we move this to a file in utils, now that it's also used by `deploy-beacon-proxy`?

packages/plugin-hardhat/src/utils/impl-address.ts
```diff
@@ -0,0 +1,79 @@
...
+import { ContractAddressOrInstance, getIBeaconFactory } from '.';
+
+/**
+ * Gets the implementation address from a Beacon.
+ *
+ * @returns the implementation address.
+ */
+export async function getImplementationAddressFromBeacon(
+  hre: HardhatRuntimeEnvironment,
+  signer: ethers.Signer | undefined,
```
frangio, line 13: We shouldn't need a signer for this function. It's an optional argument in `getIBeaconFactory`.

frangio (commented):
Great work!

packages/plugin-hardhat/src/deploy-beacon.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function makeDeployBeacon(hre: HardhatRuntimeEnvironment): DeployBeaconFunction {
+  return async function deployBeacon(ImplFactory: ContractFactory, opts: Options = {}) {
+    const { provider } = hre.network;
+    const manifest = await Manifest.forNetwork(provider);
+
+    const { impl } = await deployBeaconImpl(hre, ImplFactory, opts);
+
+    const UpgradeableBeaconFactory = await getUpgradeableBeaconFactory(hre, ImplFactory.signer);
+    const abi = ImplFactory.interface.format(FormatTypes.json);
```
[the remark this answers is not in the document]
ericglau: Agreed. This can be changed to `FormatTypes.minimal` to get a [minimal array](https://docs.ethers.io/v5/api/utils/abi/interface/#Interface--formatting).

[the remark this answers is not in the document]
ericglau: Addressed in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/456/commits/d4f23ea45e988919def4a546619f9a5d4635d635

packages/core/src/manifest.ts
```diff
@@ -25,14 +26,19 @@ export interface ImplDeployment extends Deployment {
 }
 
 export interface ProxyDeployment extends Deployment {
-  kind: 'uups' | 'transparent';
+  kind: 'uups' | 'transparent' | 'beacon';
+}
+
+export interface BeaconDeployment extends Deployment {
+  abi: string | string[];
```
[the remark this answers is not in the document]
ericglau: Addressed in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/456/commits/d4f23ea45e988919def4a546619f9a5d4635d635

packages/plugin-hardhat/src/utils/impl-address.ts
ericglau: Thanks, missed that while refactoring.

packages/plugin-hardhat/src/deploy-beacon-proxy.ts
```diff
@@ -0,0 +1,91 @@
...
+    const manifest = await Manifest.forNetwork(provider);
+
+    opts.kind = 'beacon';
+
+    const beaconAddress = getContractAddress(beacon);
+    const implAddress = await getImplementationAddressFromBeacon(hre, getSigner(ImplFactoryOrSigner), beaconAddress);
+    let contractInterface = await getInterfaceFromManifest(hre, implAddress);
+
+    if (contractInterface === undefined) {
+      if (ImplFactoryOrSigner instanceof ContractFactory) {
```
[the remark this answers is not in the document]
ericglau: Here we use the user provided beacon address to look up the implementation address, then get the interface that we previously stored in the manifest for that implementation address.  Is there any possibility that the manifest is incorrect, aside from corruption?  Considering the implementation itself is immutable.

Although since the interface is used to create the initializer call data, and if both are provided by the user, I agree it makes sense to use both together.  Then if there is a mismatch during deployment, there would be an error anyways.

packages/plugin-hardhat/src/utils/deploy-impl.ts
```diff
@@ -14,53 +14,152 @@ import {
...
-  if (opts.kind === undefined) {
-    opts.kind = inferProxyKind(validations, version);
-  }
-
-  if (proxyAddress !== undefined) {
-    await setProxyKind(provider, proxyAddress, opts);
+  if (isBeacon) {
+    if (proxyOrBeaconAddress !== undefined) {
+      if (await isTransparentOrUUPSProxy(proxyOrBeaconAddress)) {
+        throw new Error(
```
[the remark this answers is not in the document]
ericglau: Agreed. Having a separate user action section does make errors easier to understand.

packages/plugin-hardhat/src/utils/deploy-impl.ts
ericglau: I agree.  This would allow `deployImpl` to just focus on the core logic of validation and deployment.  The scenario-specific parts could perhaps be moved to the `deployProxyImpl` and `deployBeaconImpl` functions above.

packages/core/src/validate/query.ts
```diff
@@ -144,13 +146,32 @@ export function isUpgradeSafe(data: ValidationData, version: Version): boolean {
   return getErrors(dataV3, version).length == 0;
 }
 
-export function inferProxyKind(data: ValidationData, version: Version): ProxyDeployment['kind'] {
+export async function inferProxyKind(
```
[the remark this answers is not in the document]
ericglau: This seems reasonable.  It isn't really needed in `deployProxy`, and the main purpose was to give a helpful error when attempting to upgrade a beacon proxy through `upgradeProxy` or `prepareUpgrade`.  But those functions could just call `isBeaconProxy()` directly.

Another reason for this original approach though was so the Truffle plugin could also detect previously deployed beacon proxies when the user is trying to upgrade them (and then give an error about mismatching `kind`), rather than infer the wrong `kind`.  Although of course this PR does not cover Truffle and perhaps inferring the wrong `kind` is not much of a problem if it fails gracefully otherwise.

packages/plugin-hardhat/src/deploy-beacon-proxy.ts
```diff
@@ -46,18 +35,27 @@ export function makeDeployBeaconProxy(hre: HardhatRuntimeEnvironment): DeployBea
...
-          `Beacon's current implementation at address ${implAddress} was not found in the network manifest. Call deployBeaconProxy() with a contract factory for the beacon's current implementation.`,
+    let contractInterface;
+    if (opts.implementation !== undefined) {
+      contractInterface = opts.implementation.interface;
+    } else {
+      const implAddress = await getImplementationAddressFromBeacon(hre, beaconAddress);
+      contractInterface = await getInterfaceFromManifest(hre, implAddress);
+      if (contractInterface === undefined) {
+        throw new UpgradesError(
+          `Beacon's current implementation at address ${implAddress} was not found in the network manifest.`,
```
frangio, line 56:
```suggestion
          `Beacon's current implementation at ${implAddress} is unknown`,
```

packages/plugin-hardhat/src/utils/deploy-impl.ts
```diff
@@ -1,66 +1,154 @@
...
+
+  async function getBeaconImplementationAddress(beaconAddress: string): Promise<string> {
+    await assertNotProxy(beaconAddress);
+
+    const IBeaconFactory = await getIBeaconFactory(hre, ImplFactory.signer);
+    const beaconContract = IBeaconFactory.attach(beaconAddress);
+    return await beaconContract.implementation();
+  }
+
+  async function assertNotProxy(address: string) {
```
frangio, line 106: These helper functions will be useful once when we build the Truffle plugin. I would suggest trying to have them in core. But I think this can be done in the separate PR for Truffle.

packages/plugin-hardhat/src/utils/deploy-impl.ts
```diff
@@ -1,66 +1,154 @@
...
+  const deployData = await getDeployData(hre, ImplFactory, opts);
+
+  let currentImplAddress;
+  if (beaconAddress !== undefined) {
+    // upgrade scenario
+    currentImplAddress = await getBeaconImplementationAddress(beaconAddress);
+  }
+  return deployImpl(deployData, ImplFactory, opts, currentImplAddress);
+
+  async function getBeaconImplementationAddress(beaconAddress: string): Promise<string> {
```
frangio, line 98: This one would also be a good fit for core, but not in the way it's written now. We would need to change it so it doesn't use Ethers.js. We can just manually craft the `implementation()` function call and use it with `eth_call` through JSON RPC.

frangio (approved):
Some small suggestions but otherwise good to merge! Thanks.

packages/plugin-hardhat/src/prepare-upgrade.ts
```diff
@@ -1,18 +1,28 @@
...
+  return async function prepareUpgrade(proxyOrBeacon, ImplFactory, opts: Options = {}) {
+    const proxyOrBeaconAddress = getContractAddress(proxyOrBeacon);
+    const { provider } = hre.network;
+    let deployedImpl;
+    if (await isTransparentOrUUPSProxy(provider, proxyOrBeaconAddress)) {
+      deployedImpl = await deployProxyImpl(hre, ImplFactory, opts, proxyOrBeaconAddress);
+    } else if (await isBeaconProxy(provider, proxyOrBeaconAddress)) {
+      const beaconAddress = await getBeaconAddress(provider, proxyOrBeaconAddress);
+      deployedImpl = await deployBeaconImpl(hre, ImplFactory, opts, beaconAddress);
+    } else {
```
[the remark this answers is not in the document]
ericglau: Added a check in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/456/commits/54297730341a409fd926abc8c13be51db9caa12f.

However, if the address has an `implementation()` function that returns something other than an address, the returned result may still look like an address according to `ethers.utils.isAddress()`.  In this case, the plugin would just return that the "address" is not found in the manifest.

This check is not necessary in `defender.proposeUpgrade` at this time, since https://github.com/ericglau/openzeppelin-upgrades/blob/beacon/packages/plugin-defender-hardhat/src/propose-upgrade.ts#L42 would give an error if the address is not a Transparent or UUPS proxy.  In order to add this check, we may need to do some refactoring similar to the suggestion in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/456#discussion_r768174558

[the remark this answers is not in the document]
ericglau: Refactored and added this check in `defender.proposeUpgrade()`

packages/plugin-hardhat/src/utils/deploy-impl.ts
ericglau: This function is mainly implementing custom error messages which do not seem to belong in core.  The helper methods used here are part of core.

ericglau:
@frangio Please re-review since a lot of content has been refactored into core.  For your convenience, [here is a diff](https://github.com/ericglau/openzeppelin-upgrades/compare/beacon-already-reviewed...ericglau:beacon?expand=1) with the changes since the last review.

packages/core/src/beacon.ts
```diff
@@ -0,0 +1,20 @@
...
+import { EthereumProvider } from './provider';
+
+/**
+ * Checks if the address looks like a beacon.
+ *
+ * @returns true if the address has an implementation() function that returns an address, false otherwise.
+ */
+export async function isBeacon(provider: EthereumProvider, beaconAddress: string) {
+  try {
+    return await getImplementationAddressFromBeacon(provider, beaconAddress);
```
frangio, line 12: This is just returning the string, some comparison to get a boolean is missing I believe.

frangio (approved):
LGTM! Feel free to merge after this one fix.

PR #463, opened 2021-12-02
Add chalk to dependencies

kevinji:
These packages were implicitly pulling in chalk (via @openzeppelin/upgrades-core); this makes the dependency more explicit.

### DIFF
--- a/packages/plugin-hardhat/package.json
+++ b/packages/plugin-hardhat/package.json
@@ -31,7 +31,8 @@
     "rimraf": "^3.0.2"
   },
   "dependencies": {
-    "@openzeppelin/upgrades-core": "^1.10.0"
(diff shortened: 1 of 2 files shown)


Amxx:
Hello @kevinji 

As you pointed out, `chalk` is already in the dependencies of `@openzeppelin/upgrades-core`. Is there any good reason to not rely on dependency chaining and making it explicit? If so, why only `chalk` and not the other dependencies in `upgrades-core`?

Unless there is a strong reason to add it, I'd prefer to putting superfluous things in the package.json

frangio:
Chalk is used directly by hardhat-upgrades and truffle-upgrades so I agree with including it as a dependency.

Thanks @kevinji!

PR #475, opened 2021-12-20
Fix incorrect docs for upgrades.admin.changeProxyAdmin

cygnusv:
Change references in the docs to `admin.changeAdminForProxy` into `admin.changeProxyAdmin`.


Amxx (approved):
Thank you @cygnusv

PR #476, opened 2021-12-21
Check address for beacon before deploying beacon proxy

ericglau:
Adds a check to `deployBeaconProxy()` to verify that the beaconAddress points to a beacon before deploying the beacon proxy.


frangio (approved):
LGTM!

PR #478, opened 2021-12-22
Add test for UserDefinedValueType

### LINKED ISSUE #477: Support user defined value types.
When using user defined value types (available since 0.8.8) the ast-dereferencer fails.

```
Compiling 111 files with 0.8.10
An unexpected error occurred:

Error: No node with id 25034 of type StructDefinition,EnumDefinition
    at deref (projectRoot/node_modules/@openzeppelin/upgrades-core/src/ast-dereferencer.ts:38:11)
    at projectRoot/node_modules/@openzeppelin/upgrades-core/src/utils/curry.ts:11:19
    at extractStorageLayout (projectRoot/node_modules/@openzeppelin/upgrades-core/src/storage/extract.ts:59:29)
    at validate (projectRoot/node_modules/@openzeppelin/upgrades-core/src/validate/run.ts:165:67)
    at OverriddenTaskDefinition._action (projectRoot/node_modules/@openzeppelin/hardhat-upgrades/src/index.ts:62:25)
    at Environment._runTaskDefinition (projectRoot/node_modules/hardhat/src/internal/core/runtime-environment.ts:217:14)
    at Environment.run (projectRoot/node_modules/hardhat/src/internal/core/runtime-environment.ts:129:14)
    at SimpleTaskDefinition.action (projectRoot/node_modules/hardhat/src/builtin-tasks/compile.ts:975:37)
    at Environment._runTaskDefinition (projectRoot/node_modules/hardhat/src/internal/core/runtime-environment.ts:217:14)
    at Environment.run (projectRoot/node_modules/hardhat/src/internal/core/runtime-environment.ts:129:14)
error Command failed with exit code 1.
```

Apparently, using [this contract](https://github.com/Amxx/openzeppelin-contracts/blob/breaking/Timers/udvt/contracts/utils/Timers.sol) replicated the issue.

Simple contract to reproduce the current error and test support:

```
pragma solidity ^0.8.9;

contract UserDefinedValueType {
    type Timestamp is uint64;
}
```

Note: As mentioned in https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/438 we need to be careful with Solidity 0.8.8 because it has buggy storage layout rules, they were fixed in 0.8.9. We should probably reject 0.8.8.

Amxx:
This is a minimal test for #477.

Fixe should be built on top of this test.

### DIFF
--- a/packages/core/contracts/test/UserDefinedValueType.sol
+++ b/packages/core/contracts/test/UserDefinedValueType.sol
@@ -4,3 +4,4 @@ pragma solidity ^0.8.8;
 contract UserDefinedValueType {
     type Timestamp is uint64;
 }
+
--- a/packages/core/contracts/test/UserDefinedValueType.sol
+++ b/packages/core/contracts/test/UserDefinedValueType.sol
@@ -1,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.8;
+
+contract UserDefinedValueType {
+    type Timestamp is uint64;
+}
--- a/packages/core/hardhat.config.js
+++ b/packages/core/hardhat.config.js
@@ -24,7 +24,7 @@ module.exports = {
       { version: '0.5.16', settings },
       { version: '0.6.12', settings },
       { version: '0.7.6', settings },
-      { version: '0.8.2', settings },
+      { version: '0.8.8', settings },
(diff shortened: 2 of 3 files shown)


frangio:
Inlined test case in #477 for PR cleanup.

PR #487, opened 2022-01-05
Override contract deployment address

ericglau:
When using certain RPC endpoints with Polygon Mumbai, an incorrect contract address can be returned.  This causes the Hardhat plugin to write the wrong addresses to the manifest.

This occurs in:
https://forum.openzeppelin.com/t/error-no-contract-at-address-0x-removed-from-manifest/20609
https://forum.openzeppelin.com/t/invaliddeployment-error-no-contract-at-address-0x-removed-from-manifest/21383

As a workaround, calculate the correct address using the signer's address and the transaction nonce.

### DIFF
--- a/packages/plugin-hardhat/src/utils/deploy.ts
+++ b/packages/plugin-hardhat/src/utils/deploy.ts
@@ -1,5 +1,7 @@
 import type { Deployment } from '@openzeppelin/upgrades-core';
+import debug from './debug';
 import type { ethers, ContractFactory } from 'ethers';
+import { getContractAddress } from 'ethers/lib/utils';
 
 export interface DeployTransaction {
   deployTransaction: ethers.providers.TransactionResponse;
@@ -9,7 +11,24 @@ export async function deploy(
   factory: ContractFactory,
   ...args: unknown[]
 ): Promise<Required<Deployment & DeployTransaction>> {
-  const { address, deployTransaction } = await factory.deploy(...args);
+  const contractInstance = await factory.deploy(...args);
+  const deployTransaction = contractInstance.deployTransaction;
+
+  const address: string = getContractAddress({
+    from: await factory.signer.getAddress(),
+    nonce: deployTransaction.nonce,
+  });
+  if (address !== contractInstance.address) {
+    debug(
+      'overriding contract address from ' +
+        contractInstance.address +
+        ' to ' +
+        address +
+        ' for nonce ' +
+        deployTransaction.nonce,
+    );
+  }
+
   const txHash = deployTransaction.hash;
   return { address, txHash, deployTransaction };
 }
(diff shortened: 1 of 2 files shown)


packages/plugin-hardhat/src/utils/deploy.ts
```diff
@@ -9,7 +11,24 @@ export async function deploy(
   factory: ContractFactory,
   ...args: unknown[]
 ): Promise<Required<Deployment & DeployTransaction>> {
-  const { address, deployTransaction } = await factory.deploy(...args);
+  const contractInstance = await factory.deploy(...args);
+  const deployTransaction = contractInstance.deployTransaction;
+
+  const address: string = getContractAddress({
+    from: await factory.signer.getAddress(),
```
frangio, line 18: We should already have `deployTransaction.from`. In fact, we should be able to just do `getContractAddress(deployTransaction)`.

packages/plugin-hardhat/src/utils/deploy.ts
```diff
@@ -9,7 +11,24 @@ export async function deploy(
   factory: ContractFactory,
   ...args: unknown[]
 ): Promise<Required<Deployment & DeployTransaction>> {
-  const { address, deployTransaction } = await factory.deploy(...args);
+  const contractInstance = await factory.deploy(...args);
+  const deployTransaction = contractInstance.deployTransaction;
```
frangio, line 15: Shorter syntax that I prefer but let me know if you disagree:

```suggestion
  const { deployTransaction } = contractInstance;
```

packages/plugin-hardhat/src/utils/deploy.ts
```diff
@@ -9,7 +11,24 @@ export async function deploy(
...
+
+  const address: string = getContractAddress({
+    from: await factory.signer.getAddress(),
+    nonce: deployTransaction.nonce,
+  });
+  if (address !== contractInstance.address) {
+    debug(
+      'overriding contract address from ' +
```
frangio, line 23: I prefer to write this as either `[...].join(' ')`, or template strings `` `...${addr}...` ``

packages/plugin-hardhat/src/utils/deploy.ts
```diff
@@ -9,7 +11,24 @@ export async function deploy(
...
-  const { address, deployTransaction } = await factory.deploy(...args);
+  const contractInstance = await factory.deploy(...args);
+  const deployTransaction = contractInstance.deployTransaction;
+
+  const address: string = getContractAddress({
+    from: await factory.signer.getAddress(),
+    nonce: deployTransaction.nonce,
+  });
+  if (address !== contractInstance.address) {
+    debug(
```
frangio, line 22: Great to start including more `debug` statements. We have very few currently. :+1:

packages/plugin-hardhat/src/utils/deploy.ts
ericglau: The problem is `deployTransaction.from` returns the 0x0 address, as described in https://forum.openzeppelin.com/t/error-no-contract-at-address-0x-removed-from-manifest/20609/11

frangio (approved):

Issue #55, opened 2020-07-27
Reuse Truffle and Ethers tx waiting implementations or parameters


frangio:
In https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/47 I introduced `waitAndValidateDeployment` to wait until a tx is mined.

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/b3ae02961a87b990a30f7f4bc9d91668220ff00e/packages/core/src/deployment.ts#L33

There's at least two parameters here that te user might want to configure: the polling interval and the timeout. Libraries like Web3, Truffle, and Ethers already allow the user to configure these parameters and we should try to reuse them. We can also evaluate reusing their implementations for waiting a tx instead of rolling our own.

jacobrosenthal:
Did this get snuck in somewhere else by any chance? With congestion what it is, and several other fixes, Im now not crashing and able to wait for deploys, but now I'm timing out :)

frangio:
Can you clarify if you're running into an internal timeout of the plugin, or is it timing out when you use `.wait()`?

jacobrosenthal:
looks internal
```
An unexpected error occurred:

TransactionMinedTimeout [Error]: Timed out waiting for transaction 0x5581653fa28a6d8a185d20cfa81a81b22e7af685543098070593109dc1010f78
    at Object.waitAndValidateDeployment (/home/proj/node_modules/@openzeppelin/upgrades-core/src/deployment.ts:62:15)
    at fetchOrDeployGeneric (/home/proj/node_modules/@openzeppelin/upgrades-core/src/impl-store.ts:42:5)
    at Object.deployImpl (/home/proj/eth/node_modules/@openzeppelin/hardhat-upgrades/src/utils/deploy-impl.ts:40:10)
    at Proxy.deployProxy (/home/proj/eth/node_modules/@openzeppelin/hardhat-upgrades/src/deploy-proxy.ts:56:18)
    at SimpleTaskDefinition.deployCore [as action] (/home/proj/eth/tasks/deploy-more.ts:171:26)
    at Environment._runTaskDefinition (/home/proj/node_modules/hardhat/src/internal/core/runtime-environment.ts:217:14)
    at Environment.run (/home/proj/node_modules/hardhat/src/internal/core/runtime-environment.ts:129:14)
    at SimpleTaskDefinition.deploy (/home/proj/eth/tasks/deploy.ts:67:54)
[...]

PR #506, opened 2022-01-27
Add configurable timeout and pollingInterval for Hardhat

ericglau:
Addresses https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/55

- Adds `timeout` and `pollingInterval` options for Hardhat. `timeout` value of 0 means poll indefinitely until the transaction confirms.

- Not needed for Truffle since Truffle makes use of [network options for timeout and polling](https://trufflesuite.com/docs/truffle/reference/configuration.html#networks)


packages/core/src/deployment.ts
```diff
@@ -97,8 +109,13 @@ export async function waitAndValidateDeployment(provider: EthereumProvider, depl
...
 
 export class TransactionMinedTimeout extends Error {
-  constructor(readonly deployment: Deployment) {
-    super(`Timed out waiting for transaction ${deployment.txHash}`);
+  constructor(readonly deployment: Deployment, configurableTimeout: boolean) {
+    super(
+      `Timed out waiting for transaction ${deployment.txHash}. Run the function again to continue waiting for the transaction confirmation.` +
+        (configurableTimeout
+          ? ` If the problem persists, adjust the polling parameters with the timeout and pollingInterval options.`
+          : ``),
```
frangio, line 117: Can this class extend `UpgradesError` instead, so we can use the second error line for the suggestions:

```suggestion
      `Timed out waiting for transaction ${deployment.txHash}.`,
      () => `Run the function again to continue waiting for the transaction confirmation.` +
        (configurableTimeout
          ? `\nIf the problem persists, adjust the polling parameters with the timeout and pollingInterval options.`
          : ``),
```

packages/plugin-hardhat/test/beacon-timeout.js
```diff
@@ -0,0 +1,82 @@
+const test = require('ava');
```
frangio, line 1: I don't think it's necessary to have a full test suite for all three kinds of proxies. What do you think? Is there a particular reason we need to test all three?

packages/plugin-hardhat/src/utils/options.ts
```diff
@@ -2,11 +2,15 @@ import { ValidationOptions, withValidationDefaults } from '@openzeppelin/upgrade
 
 export interface Options extends ValidationOptions {
   constructorArgs?: unknown[];
+  timeout?: number;
+  pollingInterval?: number;
```
frangio, line 6: Something I learnt recently is that if we include docstrings for a field, they are shown when it's autocompleted in an IDE. Example:

https://github.com/OpenZeppelin/solidity-docgen/blob/ebae54721adfdc13cbef3141dda63a359133e21a/src/config.ts#L27-L31

image


So I think it'd be a good idea to start adding docstrings at least for the new options.

frangio, line 6: We can't extend two interfaces simultaneously but we can use an intersection type to define `Options`:

```suggestion
export type Options = ValidationOptions & DeployOpts & {
  constructorArgs?: unknown[];
```

Note: The docstrings in this case should be in the definition of DeployOpts.

frangio (commented):
Looks good! Minor comments.

frangio (approved):

frangio:
Looks good. Thanks!

PR #510, opened 2022-02-01
Fix normalization of admin in manifest

frangio:
I realized looking at a manifest that we're storing `admin.deployTransaction` with a bunch of data that we shouldn't have there. The admin deployment object wasn't being normalized to remove extra fields.


ericglau (approved):
Should a test be added?

frangio:
Added a test. Please merge if it looks good!

PR #515, opened 2022-02-09
Dont check storage compatibility is current implementation is address(0)

Amxx:
Might be useful to deploy the first implementation for beacons that are so far unset.


frangio:
Needs issue to discuss. The beacons the plugin provides will never have implementation address 0 so this PR wouldn't affect them.

Issue #474, opened 2021-12-16
A way to reconstruct a lost network file


frangio:
People accidentally delete or forget to commit their network files which actually contain information that is necessary for future upgrades. We need to provide a solution for this.

I'm thinking something like a command that they can run to reconstruct the network file given a proxy address. It can read the admin address (if transparent) from it, and the implementation address and try to match its bytecode to one of the contracts available locally. This is related to #175.

haipro287:
i have just deleted my unknown-xx.json and now i cannot upgrade my contract any more, Is there any way for me to force my contract upgrade?

ericglau:
@haipro287 We will be adding a way to import a previously deployed proxy soon.  For now, please see the workaround at https://forum.openzeppelin.com/t/upgrades-plugins-lost-my-network-files/11376/8 to manually recreate the deleted json file.

haipro287:
> @haipro287 We will be adding a way to import a previously deployed proxy soon. For now, please see the workaround at https://forum.openzeppelin.com/t/upgrades-plugins-lost-my-network-files/11376/8 to manually recreate the deleted json file.

thank you, i will try this way ^^

ericglau:
The latest release of the plugins now have a `forceImport` function that can import a previously deployed proxy.
Refer to the documentation for the `forceImport` function for [Hardhat](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades#force-import) or [Truffle](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-truffle-upgrades#force-import).

This function can be used to reconstruct a lost network file by importing each of the previously deployed proxies using their address and corresponding contract implementation.

Issue #175, opened 2020-09-17
Register impls in network file when source is available


frangio:
Some users may want to register an existing proxy for use with the plugins. It's possible to do this if the source code for the implementation is available. We should try to do this automatically on `deployProxy` and `upgradeProxy`, and maybe provide a way for the user to trigger it manually.

Related to https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/174.

frangio:
I took a shot at this but it's a bit more work than I expected. The reason is that our "version" concept (which is used by the `impl-store` module, where we would add a new `registerImplDeployment` function) hashes the contract creation code, and in order to verify that a local contract is the same as an on chain contract we have to compare the "other" bytecode, that is contained in the artifacts but not in the ContractFactory abstraction.

bonedaddy:
> I took a shot at this but it's a bit more work than I expected. The reason is that our "version" concept (which is used by the `impl-store` module, where we would add a new `registerImplDeployment` function) hashes the contract creation code, and in order to verify that a local contract is the same as an on chain contract we have to compare the "other" bytecode, that is contained in the artifacts but not in the ContractFactory abstraction.

Do you still have the code from when you took a shot at it that you can share?

frangio:
Sorry, I discarded it because I think the implementation will have to be pretty different to the approach I took.

marceljay:
As we encountered the issue in our hardhat setup with the OZ plugin, I posted in the OZ forum:
https://forum.openzeppelin.com/t/hardhat-uups-deployment-at-address-0x-is-not-registered/16529

jimrubenstein:
I definitely need this feature; migration failures happen way too often using truffle, and the last time it happened I manually called the ProxyAdmin to update the implementation address.

Now, I need to run another upgrade, but can't, because the migration didn't complete the last time - so openzep didn't store the deployed contract address. So now, we get an error during upgrade and the only way I can think to get around it is to roll back the implementation to the previous version, and then run the migration again - as if the version in between never existed.

ericglau:
The latest release of the plugins now have a `forceImport` function that can import a previously deployed proxy.
Refer to the documentation for the `forceImport` function for [Hardhat](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades#force-import) or [Truffle](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-truffle-upgrades#force-import).

[...]

PR #517, opened 2022-02-10
Add forceImport function

### LINKED ISSUE #174: Add a suggested solution to 'Deployment at address ... is not registered'
I tried to call `prepareUpgrade` with a proxy address that is not registered in `.openzeppelin` and got this error message:

`Error: Deployment at address 0xfDf2E8257d5E3667Cb744348362AeE1ECFF24125 is not registered`

It'd be nice if, after that message, the function could add a suggestion on how to fix it. Something like:

`If you want to prepare an upgrade for a proxy at that address, register it by ...`

Note: I don't know whether we already provide away for users to trigger that registration by themselves, so this might not be possible at the current version? Either way, I think it would make for a better UX :).

ericglau:
Adds `forceImport` function to import an existing proxy or beacon.

This can be used to recreate a lost network file by importing previously deployed proxies or beacons (#474) or to register proxies or beacons for upgrading even if they were not originally deployed by the plugins (#175).

Related to #474, #175 
Closes #174


packages/core/src/manifest.ts
```diff
@@ -17,11 +17,18 @@ export interface ManifestData {
     [version in string]?: ImplDeployment;
   };
   proxies: ProxyDeployment[];
-  admin?: Deployment;
+  admin?: AdminDeployment;
 }
 
-export interface ImplDeployment extends Deployment {
+export interface GenericDeployment extends Deployment {
+  bytecodeHash?: string;
```
frangio, line 24: Any reason not to add this directly in `Deployment`?

packages/core/src/impl-store.ts
```diff
@@ -41,15 +45,24 @@ async function fetchOrDeployGeneric<T extends Deployment>(
...
-      if (updated !== stored) {
+      let updated;
+      const stored = await getAndValidate<T>(deployment, lens, provider);
+      if (merge) {
+        updated = await deploy();
         await checkForAddressClash(provider, data, updated);
-        deployment.set(updated);
+        if (deployment.merge) {
+          await deployment.merge(updated);
+        } else {
```
frangio, line 55: It seems to me that we should only allow `merge = true` if `deployment.merge` exists, would you say this is right? If so, the else branch shouldn't be here and we should error merge is set incorrectly.

packages/core/src/impl-store.ts
```diff
@@ -41,15 +45,24 @@ async function fetchOrDeployGeneric<T extends Deployment>(
...
       const deployment = lens(data);
-      const stored = deployment.get();
-      if (stored === undefined) {
-        debug('deployment of', lens.description, 'not found');
-      }
-      const updated = await resumeOrDeploy(provider, stored, deploy);
-      if (updated !== stored) {
+      let updated;
+      const stored = await getAndValidate<T>(deployment, lens, provider);
+      if (merge) {
```
frangio, line 50: Something like this feels a little easier to read to me, what do you think?

```typescript
      const updated = merge ? await deploy() : await resumeOrDeploy(provider, stored, deploy);
      if (updated !== stored) {
        await checkForAddressClash(provider, data, updated);
        if (merge) {
          deployment.merge(updated);
        } else {
          deployment.set(updated);
        }
        await manifest.write(data);
      }
```

packages/core/src/impl-store.ts
```diff
@@ -70,39 +83,119 @@ async function fetchOrDeployGeneric<T extends Deployment>(
...
       });
-      e.removed = true;
+      e.reason = Reason.Removed;
     }
 
     throw e;
   }
 }
 
+async function getAndValidate<T extends GenericDeployment>(
```
frangio, line 93: One thing I don't understand is why we have validatoin both in `waitAndValidate` and `getAndValidate`, and there also seem to be two layers of validation here with the separate `validate` function.

Is there a possibility to simplify all this by simply throwing an error and letting the `catch` statement in `fetchOrDeployGeneric` delete it if it's a development network?

packages/core/src/impl-store.ts
```diff
@@ -70,39 +83,119 @@ async function fetchOrDeployGeneric<T extends Deployment>(
...
+  }
+
+  return { address: existing.address, allAddresses: Array.from(merged) };
+}
+
 export async function fetchOrDeployAdmin(
   provider: EthereumProvider,
   deploy: () => Promise<Deployment>,
-  opts?: DeployOpts,
+  merge?: DeployOpts,
```
frangio, line 190: Is it ever possible to merge an admin? Should this remain `opts`?

packages/plugin-hardhat/src/import-proxy.ts
```diff
@@ -0,0 +1,108 @@
...
+    opts: ImportProxyOptions = {},
+  ) {
+    const { provider } = hre.network;
+    const manifest = await Manifest.forNetwork(provider);
+
+    const proxyOrBeaconAddress = getContractAddress(proxyOrBeacon);
+
+    const implAddress = await getImplementationAddressFromProxy(provider, proxyOrBeaconAddress);
+    if (implAddress !== undefined) {
+      const importKind = await detectProxyKind(provider, proxyOrBeaconAddress);
```
frangio, line 44: This should probably be moved inside `importProxyToManifest`.

packages/plugin-hardhat/src/utils/simulate-deploy.ts
```diff
@@ -0,0 +1,72 @@
...
+  hre: HardhatRuntimeEnvironment,
+  ImplFactory: ContractFactory,
+  opts: Options,
+  implAddress: string,
+  runtimeBytecode?: string,
+) {
+  const deployData = await getDeployData(hre, ImplFactory, opts);
+  const simulateDeploy = async () => {
+    const abi = ImplFactory.interface.format(FormatTypes.minimal) as string[];
+    const deployment = Object.assign({ abi });
```
frangio, line 63: This looks a little weird, what's the goal? Why not just inline `abi` below in the object below?

packages/plugin-truffle/src/utils/simulate-deploy.ts
```diff
@@ -0,0 +1,58 @@
...
+  await fetchOrDeploy(deployData.version, deployData.provider, simulateDeploy, opts, true);
+}
+
+async function getSimulatedData(Contract: ContractClass, opts: Options, implAddress: string, runtimeBytecode?: string) {
+  const deployData = await getDeployData(opts, Contract);
+  const simulateDeploy = async () => {
+    const abi = (Contract as any).abi;
+    const deployment = Object.assign(
+      { abi },
+      await deploy(deployData.fullOpts.deployer, Contract, ...deployData.fullOpts.constructorArgs),
```
frangio, line 48: Isn't this running the actual deployment?

packages/core/src/impl-store.ts
ericglau: Must have been a refactoring mistake.  It should be `opts`.

packages/plugin-truffle/src/utils/simulate-deploy.ts
ericglau: Good catch!

packages/plugin-hardhat/src/utils/simulate-deploy.ts
```diff
@@ -43,40 +43,30 @@ export async function simulateDeployImpl(
...
   const simulateDeploy = async () => {
-    const abi = contractFactory.interface.format(FormatTypes.minimal) as string[];
+    const abi = ImplFactory.interface.format(FormatTypes.minimal) as string[];
     const deployment = Object.assign({ abi });
-    return { ...deployment, layout: deployData.layout, address: addr, bytecodeHash };
+    return {
+      ...deployment,
+      layout: deployData.layout,
+      address: implAddress,
+      bytecodeHash: runtimeBytecode && hashBytecode(runtimeBytecode),
```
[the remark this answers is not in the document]
ericglau: Changed `runtimeBytecode` to be required.

packages/core/src/impl-store.ts
ericglau: Changed it to throw an error [here](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/517/commits/57e110956fdcda903939d85ee9ee7d60b3ddcdc3#diff-74a100ab37e69fa1bd58695b1f5b6870c10f5a7c542ef38eb19e3ebe7a3d195eR47), but the TS compiler gave an error that `deployment.merge` may be undefined [here](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/517/commits/57e110956fdcda903939d85ee9ee7d60b3ddcdc3#diff-74a100ab37e69fa1bd58695b1f5b6870c10f5a7c542ef38eb19e3ebe7a3d195eR58) so I had to force it with `deployment.merge!(updated);`.  This leads to a linting warning `Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion`

I could put the check directly before calling `deployment.merge(updated);` but then this could be after an actual deployment occurred.  It is better to check it earlier.  Any thoughts?

frangio: The simplest fix is to write `if (merge && deployment.merge)` even if we know the first condition implies the second.

packages/core/src/impl-store.ts
ericglau: Changed `resumeOrDeploy` to do additional validation on the existing deployment.  `waitAndValidate` is still needed as an additional validatation _after_ the new deployment occurs.

packages/core/src/impl-store.ts
```diff
@@ -50,13 +50,16 @@ async function fetchOrDeployGeneric<T extends GenericDeployment>(
         );
       }
 
-      const stored = await getAndValidate(deployment, lens, provider);
-      const updated = merge ? await deploy() : await resumeOrDeploy(provider, stored, deploy);
+      const stored = deployment.get();
+      const updated = await resumeOrDeploy(provider, stored, deploy, lens.type, opts, deployment, merge);
```
frangio, line 54: `stored` as an argument seems redundant given `deployment` is there.

ericglau: agreed, but `resumeOrDeploy` is exported and if we want to maintain backward compatibility in this version of core, we need to keep the `stored` parameter.

frangio (approved):
Looks good! :rocket:

PR #519, opened 2022-02-14
Use Solc storage layout

JulissaDantes:
This change includes:

- Add the settings to obtain the `storageLayout` during compilation.
- Include sending it as a paramter to the `extractStorageLayout` function only on the hardhat side `validate` function.
- Include the changes inside `extractStorageLayout` to use the storageLayout provided by the compiler if any.


packages/core/src/storage/extract.ts
```diff
@@ -17,62 +16,77 @@ export function extractStorageLayout(
...
+      const varDecl = contractDef.nodes.filter(n => n.id == storage.astId && isNodeType('VariableDeclaration', n))[0];
+      if (varDecl) {
         layout.storage.push({
           contract: contractDef.name,
-          label: varDecl.name,
-          type,
+          label: storage.label,
+          type: storage.type,
           src: decodeSrc(varDecl),
         });
```
frangio, line 33:
```suggestion
        const { label, offset, slot, type } = storage;
        const src = decodeSrc(varDecl);
        const contract = /* ??? */;
        layout.storage.push({ label, offset, slot, type, contract, src });
```

Note: `contract` must be the contract where the variable is defined, which is not `contractDef.name`, because `storageLayout.storage` is already linearized.

frangio: Idea: iterate over `contractDef.linearizedBaseParents` and see which one contains this variable declaration.

packages/plugin-hardhat/hardhat.config.js
```diff
@@ -1,20 +1,21 @@
 require('@openzeppelin/hardhat-upgrades');
 
+const settings = {
```
frangio, line 3: We need to use `extendConfig` to add this automatically.

packages/plugin-hardhat/src/index.ts
```diff
@@ -116,3 +116,15 @@ extendEnvironment(hre => {
...
   });
 });
+
+extendConfig((config: HardhatConfig) => {
+  for (const compiler of config.solidity.compilers) {
+    if (compiler.settings.outputSelection['*']['*'] === undefined) {
+      compiler.settings.outputSelection['*']['*'] = [];
+    }
+
+    if (compiler.settings.outputSelection['*']['*'].indexOf('storageLayout') === -1) {
```
frangio, line 126:
```suggestion
    if (!compiler.settings.outputSelection['*']['*'].includes('storageLayout')) {
```

packages/plugin-hardhat/src/index.ts
```diff
@@ -116,3 +116,15 @@ extendEnvironment(hre => {
     };
   });
 });
+
+extendConfig((config: HardhatConfig) => {
+  for (const compiler of config.solidity.compilers) {
+    if (compiler.settings.outputSelection['*']['*'] === undefined) {
+      compiler.settings.outputSelection['*']['*'] = [];
+    }
```
frangio, line 124: `??=` is a very useful operator, it does exactly the same.

```suggestion
    compiler.settings.outputSelection['*']['*'] ??= [];
```

packages/plugin-hardhat/hardhat.config.js
```diff
@@ -1,20 +1,20 @@
...
```
frangio, line 3: Can you remove these settings, they shouldn't be necesary with the new `extendConfig` code.

packages/core/src/storage/layout.ts
```diff
@@ -9,13 +9,17 @@ export interface StorageLayout {
   layoutVersion?: string;
   storage: StorageItem[];
   types: Record<string, TypeItem>;
+  flat?: boolean;
 }
 
 export interface StorageItem<Type = string> {
+  astId?: number;
```
frangio, line 16: We don't want to store ast ids. I don't think we need them.

packages/core/src/storage/layout.ts
```diff
@@ -28,6 +32,10 @@ export type TypeItemMembers<Type = string> = StructMember<Type>[] | EnumMember[]
 export interface StructMember<Type = string> {
   label: string;
   type: Type;
+  contract?: string;
+  astId?: number;
+  numberOfBytes?: number;
+  encoding?: string;
```
frangio, line 38: Do we need these new fields anywhere? I think we just want `label` and `type`.

packages/core/src/storage/extract.ts
```diff
@@ -108,3 +149,23 @@ function getTypeMembers(typeDef: StructDefinition | EnumDefinition): TypeItem['m
     return typeDef.members.map(m => m.name);
   }
 }
+
+function getOriginContract(
+  basecontract: ContractDefinition,
```
frangio, line 154: Please rename to `contract`. "Base contract" can be interpreted as "parent contract".

packages/core/src/storage/extract.ts
```diff
@@ -108,3 +149,23 @@ function getTypeMembers(typeDef: StructDefinition | EnumDefinition): TypeItem['m
...
+  astId: number | undefined,
+  deref: ASTDereferencer,
+): [VariableDeclaration | undefined, string] {
+  for (const id of basecontract.linearizedBaseContracts.reverse()) {
+    if (id === basecontract.id) {
+      continue;
+    }
+    const parentContract = deref(['ContractDefinition'], id);
+
+    const varDecl = parentContract.nodes.filter(n => n.id == astId)[0];
```
frangio, line 164:
```suggestion
    const varDecl = parentContract.nodes.find(n => n.id == astId);
```

packages/core/src/storage/extract.ts
```diff
@@ -108,3 +149,23 @@ function getTypeMembers(typeDef: StructDefinition | EnumDefinition): TypeItem['m
     return typeDef.members.map(m => m.name);
   }
 }
+
+function getOriginContract(
+  basecontract: ContractDefinition,
+  astId: number | undefined,
+  deref: ASTDereferencer,
+): [VariableDeclaration | undefined, string] {
```
frangio, line 157: This should be `undefined | [VariableDeclaration, string]`.

packages/core/src/storage/extract.ts
```diff
@@ -108,3 +149,23 @@ function getTypeMembers(typeDef: StructDefinition | EnumDefinition): TypeItem['m
...
+    }
+    const parentContract = deref(['ContractDefinition'], id);
+
+    const varDecl = parentContract.nodes.filter(n => n.id == astId)[0];
+    if (varDecl && isNodeType('VariableDeclaration', varDecl)) {
+      return [varDecl, parentContract.name];
+    }
+  }
+
+  return [undefined, ''];
```
frangio, line 170: If we change the type as I suggested above, the explicit return statement is not necessary.

packages/core/src/storage/extract.ts
```diff
@@ -17,62 +22,98 @@ export function extractStorageLayout(
...
+          // this is for those cases
+          type[1].members = undefined;
+        }
+      }
+    }
 
-  // Note: A UserDefinedTypeName can also refer to a ContractDefinition but we won't care about those.
-  const derefUserDefinedType = deref(['StructDefinition', 'EnumDefinition']);
+    for (const storage of storageLayout.storage) {
+      let varDecl: any = contractDef.nodes.filter(n => n.id == storage.astId)[0];
```
frangio, line 49: No need for this optimization IMO. It should be handled in `getOriginContract`, instead of skipping the "base contract" in that function.

packages/core/src/storage/extract.ts
```diff
@@ -17,62 +22,98 @@ export function extractStorageLayout(
...
+            if (typeof member === 'object' && 'contract' in member && member.contract) {
+              const [, contract] = getOriginContract(contractDef, member.astId, deref);
+              member.contract = contract;
+            }
+          }
+        } else {
+          // In some cases the past version didnt have members and the updated version has members = undefined,
+          // this is for those cases
+          type[1].members = undefined;
+        }
```
frangio, line 44: This branch looks unnecessary.

packages/core/src/storage/extract.ts
```diff
@@ -17,62 +22,98 @@ export function extractStorageLayout(
...
+      }
+    }
```
frangio, line 46: I want this to use `mapValues` so we make a copy of the object instead of modifying it.

I think it can be written like this:

```typescript
    layout.types = mapValues(storageLayout.types, t => {
      let { label, members } = t;
      if (members) {
        members = members.map(m => pick(m, ['label', 'type']));
      }
      return { label, members };
    });
```

packages/core/src/storage/extract.ts
```diff
@@ -17,62 +22,98 @@ export function extractStorageLayout(
...
   decodeSrc: SrcDecoder,
   deref: ASTDereferencer,
+  storageLayout?: StorageLayout | undefined,
 ): StorageLayout {
-  const layout: StorageLayout = { storage: [], types: {}, layoutVersion: currentLayoutVersion };
+  const layout: StorageLayout = { storage: [], types: {}, layoutVersion: currentLayoutVersion, flat: false };
+  if (storageLayout !== undefined) {
+    layout.types = storageLayout.types;
+    if (layout.types) {
+      for (const type of Object.entries(layout.types)) {
```
Amxx, line 31: maybe it would be clearer to for

```
for (const [ key, type ] of Object.entries(layout.type)) { 
```

You could even remove line 30 by doing
```
for (const [ key, type ] of Object.entries(layout.type || {})) { 
```

packages/core/src/storage/extract.ts
Amxx, line 49: `.filter(n => n.id == storage.astId)[0]` 
can be replaced by
`.find(n => n.id == storage.astId)`

packages/plugin-hardhat/src/index.ts
```diff
@@ -116,3 +116,15 @@ extendEnvironment(hre => {
     };
   });
 });
+
+extendConfig((config: HardhatConfig) => {
+  for (const compiler of config.solidity.compilers) {
+    if (compiler.settings.outputSelection['*']['*'] === undefined) {
```
Amxx, line 122: we could possibly replace this first if with

```
compiler.settings.outputSelection['*']['*'] ||= [];
```

Amxx, line 122: this will fail if:
- `compiler.settings` is undefined,
- `compiler.settings.outputSelection` is undefined
- `compiler.settings.outputSelection['*']` is undefined is undefined 

we need another way of setting that

Amxx: maybe

```
Object.assign({ settings: {}}, compiler);
Object.assign({ outputSelection: {}}, compiler.settings);
Object.assign({ '*': {}}, compiler.settings.outputSelection);
Object.assign({ '*': {}}, compiler.settings.outputSelection['*']);
```

Amxx:
```
const set = (obj, key, value = {}) => (key in obj) || (obj[key] = value);
set(compiler, 'settings');
set(compiler.settings, 'outputSelection');
set(compiler.settings.outputSelection, '*');
set(compiler.settings.outputSelection['*'], '*', []);
```

packages/plugin-hardhat/hardhat.config.js
JulissaDantes: Those were there before, should I add them to the extendConfig?

packages/core/src/storage/extract.ts
JulissaDantes: Could not use mapValues on typescript so I did iterated over `Object.keys(storageLayout.types)` instead

packages/core/src/storage/layout.ts
frangio: Oh from when we read these objects from solc's `storageLayout`. I see.

We need this optional field here but we don't want

packages/plugin-hardhat/src/index.ts
Amxx: this is marked outdated, but is not resolved

packages/core/src/storage/extract.ts
```diff
@@ -17,62 +22,96 @@ export function extractStorageLayout(
...
   decodeSrc: SrcDecoder,
   deref: ASTDereferencer,
+  storageLayout?: StorageLayout | undefined,
 ): StorageLayout {
-  const layout: StorageLayout = { storage: [], types: {}, layoutVersion: currentLayoutVersion };
+  const layout: StorageLayout = { storage: [], types: {}, layoutVersion: currentLayoutVersion, flat: false };
+  if (storageLayout !== undefined) {
+    if (storageLayout.types) {
+      for (const key of Object.keys(storageLayout.types)) {
+        const label = storageLayout.types[key].label;
```
Amxx, line 31: any reason for not doing 

```
      for (const [ key, label ] of Object.entries(storageLayout.types)) {
```
?

packages/core/src/storage/extract.ts
```diff
@@ -17,62 +22,96 @@ export function extractStorageLayout(
   contractDef: ContractDefinition,
   decodeSrc: SrcDecoder,
   deref: ASTDereferencer,
+  storageLayout?: StorageLayout | undefined,
 ): StorageLayout {
-  const layout: StorageLayout = { storage: [], types: {}, layoutVersion: currentLayoutVersion };
+  const layout: StorageLayout = { storage: [], types: {}, layoutVersion: currentLayoutVersion, flat: false };
+  if (storageLayout !== undefined) {
+    if (storageLayout.types) {
+      for (const key of Object.keys(storageLayout.types)) {
```
Amxx, line 30: I'm not sure I understand this. It feels like:

- `layout.types` is a key => { label, member } object.
- `layout.types` is almost a copy of `storageLayout.types`
- the difference it that, going from `storageLayout.types` to `layout.types` you apply a map
- the map really doesn't do anything ...

frangio: The map does a `pick` to remove unwanted fields. In a previous comment that is now lost I suggested using the `pick` function directly. This should be:

```
solcMembers.map(m => typeof m === 'object' ? pick(m, ['label', 'type']) : m);

frangio (approved):

PR #529, opened 2022-03-02
Support function types in storage layout

### LINKED ISSUE #526: Function Type Support
First thank you for this wonderful open source library. It seems upgrades doesn't support function types inside structs. Even when simply upgrading the same contract unchanged it throws an incompatibility error. I'll give an example below to allow reproducing the error.

`Test.sol`
```solidity
pragma solidity ^0.8.0;
import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";


contract TestV1 is Initializable, OwnableUpgradeable, UUPSUpgradeable {
    struct TestStruct {
        function () internal view returns (uint256) callback;
    }


    TestStruct test;

    constructor() initializer {}

    function initialize() initializer public {
        __Ownable_init();
        __UUPSUpgradeable_init();
        test.callback = _callback;
    }

    function _callback() internal view returns (uint256) {
        return 1;
    }

    function _authorizeUpgrade(address newImplementation)
    internal
    onlyOwner
    override
    {}

}

contract TestV2 is TestV1 {

    function helloWorld() external view returns (uint256) {
        return 2;
    }
}

```

`deploy-test.js`
```js
// We require the Hardhat Runtime Environment explicitly here. This is optional
// but useful for running the script in a standalone fashion through `node <script>`.
//
// When running the script with `npx hardhat run <script>` you'll find the Hardhat
// Runtime Environment's members available in the global scope.
const { ethers, upgrades } = require("hardhat");

async function main() {
  // Hardhat always runs the compile task when running scripts with its command
  // line interface.
  //
  // If this script is run directly using `node` you may want to call compile
  // manually to make sure everything is compiled
  // await hre.run('compile');
  const TestV1 = await ethers.getContractFactory("TestV1");
  const testV1 = await upgrades.deployProxy(TestV1, [], { kind: "uups" });
  const TestV2 = await ethers.getContractFactory("TestV2");
  await upgrades.upgradeProxy(testV1.address, TestV2);
}

// We recommend this pattern to be able to use async/await everywhere
// and properly handle errors.
main()
  .then(() => process.exit(0))
  .catch((error) => {
    console.error(error);
    process.exit(1);
  });
```

# Procedure
> $ hardhat node
$ hardhat run --network localhost ./scripts/deploy-test.js

# Results
> Solidity compilation finished successfully
contracts/Test.sol:13: Upgraded `test` to an incompatible type
> - Bad upgrade to struct TestV1.TestStruct
> - In struct TestV1.TestStruct
>   - Upgraded `callback` to an incompatible type
>     - Unknown type function () view returns (uint256)

Unless I am mistaken I see no reason why the above would fail considering the storage layout is unchanged in `TestV2`. 
Thanks in advance for any help you provide.

leonwilly:
Added basic function type checking to ensure visibility and mutability is maintained when upgrading. Fixes #526.
Updated the `report` & `type-id` tests.


frangio:
Thank you @leonwilly! This is a valuable change, but I would like the implementation to be a little different. Can you please do the following changes?

- Revert the changes to `parseTypeId`, it should remain agnostic to what type it's parsing. Mutability and visibility are part of the "head".
  - If necessary, add a separate function to parse a function type head, or to at least extract its visibility.
- We only care about storage layout compatibility when comparing types. Mutability doesn't seem to affect storage layout in any way. Visibility is important though because in storage internal functions are 8 bytes, and external functions are 24 bytes. So we have to check that visibility is the same, but we can ignore mutability changes.

leonwilly:
> Thank you @leonwilly! This is a valuable change, but I would like the implementation to be a little different. Can you please do the following changes?
> 
> * Revert the changes to `parseTypeId`, it should remain agnostic to what type it's parsing. Mutability and visibility are part of the "head".
>   
>   * If necessary, add a separate function to parse a function type head, or to at least extract its visibility.
> * We only care about storage layout compatibility when comparing types. Mutability doesn't seem to affect storage layout in any way. Visibility is important though because in storage internal functions are 8 bytes, and external functions are 24 bytes. So we have to check that visibility is the same, but we can ignore mutability changes.

Frangio,

Sure thing I'll make those changes now.

Gabe

leonwilly:
@frangio 

I made the changes you requested. Let me know if you see anything else that needs to be changed.

packages/core/src/storage/report.ts
```diff
@@ -90,11 +90,12 @@ function explainStorageOperation(op: StorageOperation<StorageField>, ctx: Storag
 
 function explainTypeChange(ch: TypeChange): string {
   switch (ch.kind) {
+    case 'visibility change':
+      return `Bad upgrade visibility changed ${describeTransition(ch.original, ch.updated)}`;
```
frangio, line 94: This is more in line with the rest of the function:

```suggestion
      return `Bad upgrade ${describeTransition(ch.original, ch.updated)}\nDifferent visibility`;

```

packages/core/src/storage/report.ts
```diff
@@ -90,11 +90,12 @@ function explainStorageOperation(op: StorageOperation<StorageField>, ctx: Storag
...
     case 'obvious mismatch':
     case 'struct members':
     case 'enum members':
       return `Bad upgrade ${describeTransition(ch.original, ch.updated)}`;
-
```
frangio, line 97: Please restore this newline here.

packages/core/src/storage/compare.ts
```diff
@@ -78,6 +79,16 @@ export class StorageLayoutComparator {
     }
   }
 
+  getVisibilityChange(original: ParsedTypeDetailed, updated: ParsedTypeDetailed): TypeChange | undefined {
+    const re = /(internal|external|public|private)/
```
frangio, line 83: Better to match more strictly what we assume we will see:

```suggestion
    const re = /^t_function_(internal|external)/
```

packages/core/src/storage/compare.ts
```diff
@@ -121,10 +132,11 @@ export class StorageLayoutComparator {
     updated: ParsedTypeDetailed,
     { allowAppend }: { allowAppend: boolean },
   ): TypeChange | undefined {
-    if (original.head !== updated.head) {
+    if (updated.head.startsWith('t_function')) {
+      return this.getVisibilityChange(original, updated);
+    } else if (original.head !== updated.head) {
       return { kind: 'obvious mismatch', original, updated };
     }
```
frangio, line 139: I prefer this order:

```suggestion
    if (original.head !== updated.head) {
      return { kind: 'obvious mismatch', original, updated };
    }

    if (updated.head.startsWith('t_function')) {
      return this.getVisibilityChange(original, updated);
    }

```

leonwilly: I don't think the above will work. That's why I had them ordered the other way. Because the head will never match if visibility is changed. So the second if statement will never trigger. That being said the "obvious mismatch" does work in that situation. The real error was caused by the fact that the `t_function` types weren't being caught in the switch statement.

packages/core/src/storage/report.ts
leonwilly: I agree 100%. I should have used the `function (<parameter types>) {internal|external} [pure|view|payable] [returns (<return types>)]` EBNF.

packages/core/src/storage/compare.ts
frangio: Ahh yes you're right, my bad.

Just removing else then:

```suggestion
    if (updated.head.startsWith('t_function')) {
      return this.getVisibilityChange(original, updated);
    }

    if (original.head !== updated.head) {
      return { kind: 'obvious mismatch', original, updated };
    }

```

leonwilly (commented):
Thanks!

frangio (approved):
LGTM. Thank you!

frangio:
This makes total sense to be honest. Should've thought of that before. Internal functions in storage should not be allowed.

PR #530, opened 2022-03-02
Use lockfile when reading validations.json

### LINKED ISSUE #528: Compilation failures due parallel compilation introduced in Hardhat 2.9.0
### Description
Hardhat 2.9.0 released a few days ago enables parallel compilation of Solidity code (https://github.com/NomicFoundation/hardhat/pull/2372). This breaks openzeppelin-upgrades because the library adds a Hardhat task that passes information through a single file (`cache/validations.json`) and now that file is being read and written by multile instances of the task in parallel. This sometimes results in the file being corrupted, and one of the following errors shows up upon reading it:
```
SyntaxError: Unexpected string in JSON at position 503
```
```
SyntaxError: Unexpected end of JSON input
```

This potential problem seems to have been considered in #59 but it was decided that it's not actionable and the parallelism just has to be disabled. I'm not sure if Hardhat even offers an option to do that now.

### Details
Yesterday we started getting random compilation failures in one of external projects we use to test the Solidity compiler ([Perpetual Pools](https://github.com/solidity-external-tests/perpetual-pools-contracts)). For example [989084](https://app.circleci.com/pipelines/github/ethereum/solidity/22500/workflows/94e16ffa-fbc9-4314-af4c-45b02060c5c3/jobs/989084):
```
An unexpected error occurred:
SyntaxError: Unexpected string in JSON at position 503
    at JSON.parse (<anonymous>)
    at readValidations (/tmp/ext-test-Tracer-Perpetual-Pools-mz6vJa/ext/node_modules/@openzeppelin/hardhat-upgrades/src/utils/validations.ts:34:23)
    at async writeValidations (/tmp/ext-test-Tracer-Perpetual-Pools-mz6vJa/ext/node_modules/@openzeppelin/hardhat-upgrades/src/utils/validations.ts:16:18)
    at async OverriddenTaskDefinition._action (/tmp/ext-test-Tracer-Perpetual-Pools-mz6vJa/ext/node_modules/@openzeppelin/hardhat-upgrades/src/index.ts:72:5)
    at async Environment._runTaskDefinition (/tmp/ext-test-Tracer-Perpetual-Pools-mz6vJa/ext/node_modules/hardhat/src/internal/core/runtime-environment.ts:217:14)
    at async Environment.run (/tmp/ext-test-Tracer-Perpetual-Pools-mz6vJa/ext/node_modules/hardhat/src/internal/core/runtime-environment.ts:129:14)
    at async SimpleTaskDefinition.action (/tmp/ext-test-Tracer-Perpetual-Pools-mz6vJa/ext/node_modules/hardhat/src/builtin-tasks/compile.ts:985:37)
    at async Environment._runTaskDefinition (/tmp/ext-test-Tracer-Perpetual-Pools-mz6vJa/ext/node_modules/hardhat/src/internal/core/runtime-environment.ts:217:14)
    at async Environment.run (/tmp/ext-test-Tracer-Perpetual-Pools-mz6vJa/ext/node_modules/hardhat/src/internal/core/runtime-environment.ts:129:14)
    at async /tmp/ext-test-Tracer-Perpetual-Pools-mz6vJa/ext/node_modules/p-map/index.js:57:22
```

The problem is not fully deterministic and does not happen every time but it's still easily reproducible.

The failure happens in [`validations.ts`](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/%40openzeppelin/hardhat-upgrades%401.15.0/packages/plugin-hardhat/src/utils/validations.ts) when `cache/validations.json` is being read - the file is corrupted and looks like it's stitched from halves of two different JSON files. The cause seems to be `writeValidations()`, which creates the file. It looks like a race condition between multiple processes/threads using that function in parallel to write to the file.

### How to reproduce
```bash
git clone https://github.com/solidity-external-tests/perpetual-pools-contracts.git
cd perpetual-pools-contracts/
yarn install
yarn add hardhat@2.9.0
while npx hardhat compile; do
    rm artifacts/ cache/ -rf
done
```

frangio:
Fixes #528 

Lockfile use was copied from Manifest.lock.

### DIFF
--- a/packages/plugin-hardhat/src/utils/validations.ts
+++ b/packages/plugin-hardhat/src/utils/validations.ts
@@ -20,17 +20,17 @@ export async function writeValidations(hre: HardhatRuntimeEnvironment, newRunDat
   let releaseLock;
   try {
     releaseLock = await lock(cachePath);
-    const storedData = await readValidations(hre, false);
-
+    const storedData = await readValidations(hre, false).catch(e => {
+      // If there is no previous data to append to, we ignore the error and write
+      // the file from scratch.
+      if (e instanceof ValidationsCacheNotFound) {
+        return undefined;
+      } else {
+        throw e;
+      }
+    });
     const validations = concatRunData(newRunData, storedData);
-
     await fs.writeFile(cachePath, JSON.stringify(validations, null, 2));
-  } catch (e) {
-    // If there is no previous data to append to, we ignore the error and write
-    // the file from scratch.
-    if (!(e instanceof ValidationsCacheNotFound)) {
-      throw e;
-    }
   } finally {
     await releaseLock?.();
   }
--- a/packages/plugin-hardhat/src/utils/validations.ts
+++ b/packages/plugin-hardhat/src/utils/validations.ts
@@ -10,7 +10,8 @@ import {
   isCurrentValidationData,
 } from '@openzeppelin/upgrades-core';
 
-function lock(file: string) {
+async function lock(file: string) {
+  await fs.mkdir(path.dirname(file), { recursive: true });
   return lockfile.lock(file, { retries: 3, realpath: false });
 }
 
@@ -23,7 +24,6 @@ export async function writeValidations(hre: HardhatRuntimeEnvironment, newRunDat
 
     const validations = concatRunData(newRunData, storedData);
 
-    await fs.mkdir(hre.config.paths.cache, { recursive: true });
     await fs.writeFile(cachePath, JSON.stringify(validations, null, 2));
   } catch (e) {
     // If there is no previous data to append to, we ignore the error and write
--- a/packages/plugin-hardhat/src/utils/validations.ts
+++ b/packages/plugin-hardhat/src/utils/validations.ts
@@ -36,7 +36,10 @@ export async function writeValidations(hre: HardhatRuntimeEnvironment, newRunDat
   }
 }
 
-export async function readValidations(hre: HardhatRuntimeEnvironment, acquireLock = true): Promise<ValidationDataCurrent> {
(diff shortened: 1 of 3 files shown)


packages/plugin-hardhat/src/utils/validations.ts
```diff
@@ -9,28 +10,42 @@ import {
   isCurrentValidationData,
 } from '@openzeppelin/upgrades-core';
 
-export async function writeValidations(hre: HardhatRuntimeEnvironment, newRunData: ValidationRunData): Promise<void> {
-  let storedData: ValidationDataCurrent | undefined;
+function lock(file: string) {
+  return lockfile.lock(file, { retries: 3, realpath: false });
```
ericglau, line 14: There are some test errors:

- If the `plugin-hardhat/cache` folder does not exist, this causes an error (which is strange since `realpath` is false):
```
[Error: ENOENT: no such file or directory, mkdir '/home/runner/work/openzeppelin-upgrades/openzeppelin-upgrades/packages/plugin-hardhat/cache/validations.json.lock'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'mkdir',
  path: '/home/runner/work/openzeppelin-upgrades/openzeppelin-upgrades/packages/plugin-hardhat/cache/validations.json.lock'
}
```

- If `plugin-hardhat/cache` exists but is empty, the tests fail at line 57 instead with `ValidationsCacheNotFound`

frangio: With or without realpath, the containing directory needs to exist. I added mkdir before lock.

ericglau (approved):

PR #534, opened 2022-03-08
Fix bug that caused missing members when using solc storageLayout

frangio:
In `loadLayoutType` when we recursively look for types for which to load members, we were entirely skipping types that themselves already had their members loaded, even though we have to recurse into them to look for other types that don't. This can manifest when a struct member has an enum type.


JulissaDantes:
Looks good!

PR #535, opened 2022-03-09
Support TypeChain in deployProxy function

frimoldi:
As discussed in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/325, this change allows `deployProxy` to returned a typed `Contract` instance, so the user can use things like Typechain without loosing contract's types definition.

Usage example in a hardhat script 👇🏽
```
import { ethers, upgrades } from "hardhat";
import { Greeter__factory } from "../typechain";

async function main() {
  const signer = (await ethers.getSigners())[0];
  const factory = new Greeter__factory(signer);
  const greeter = await upgrades.deployProxy<Greeter__factory>(factory, {
    constructorArgs: ["Fran"],
  });

  await greeter.deployed();

  // this is now type safe
  await greeter.greet();
}
```

### DIFF
--- a/packages/plugin-hardhat/src/deploy-proxy.ts
+++ b/packages/plugin-hardhat/src/deploy-proxy.ts
@@ -20,17 +20,19 @@ import {
   getInitializerData,
 } from './utils';
 
+type InstanceOf<F extends ContractFactory> = ReturnType<F['attach']>;
+
 export interface DeployFunction {
-  (ImplFactory: ContractFactory, args?: unknown[], opts?: DeployProxyOptions): Promise<Contract>;
-  (ImplFactory: ContractFactory, opts?: DeployProxyOptions): Promise<Contract>;
+  <F extends ContractFactory>(ImplFactory: F, args?: unknown[], opts?: DeployProxyOptions): Promise<InstanceOf<F>>;
+  <F extends ContractFactory>(ImplFactory: F, opts?: DeployProxyOptions): Promise<InstanceOf<F>>;
 }
 
 export function makeDeployProxy(hre: HardhatRuntimeEnvironment): DeployFunction {
-  return async function deployProxy(
-    ImplFactory: ContractFactory,
+  return async function deployProxy<F extends ContractFactory>(
+    ImplFactory: F,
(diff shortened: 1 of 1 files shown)


frimoldi:
@frangio please let me know your thoughts on this.

packages/plugin-hardhat/src/deploy-proxy.ts
```diff
@@ -20,17 +20,19 @@ import {
   getInitializerData,
 } from './utils';
 
+type InstanceOf<F extends ContractFactory> = ReturnType<F['attach']>;
```
frangio, line 23: Can you call this `InstanceOfFactory` or something with Factory?

frangio (changes requested):
Thank you @frimoldi! This looks solid. I'd love to have a test for this but our tests in this package are written in JavaScript so I don't think we can set one up.

packages/plugin-hardhat/src/deploy-proxy.ts
frimoldi: yeah, good call.

frangio (commented):
Sorry for the delay. I realized we also have to cover the other functions like `upgradeProxy`, `deployBeacon`, etc. Do you want to take a look at those?

frimoldi:
@frangio sure, I'll take a look as soon as I can.

frangio:
@frimoldi Let us know if you can carry on with this. :slightly_smiling_face:

frimoldi:
@frangio Yes. Sorry for the delay, I'll find some time this week.

0xakihiko:
Would love this

0xWhiteleaf:
Any update on this?

bulbazavr1024:
@frimoldi Hey! How it's going ?)

frankbonnet:
What's the status of this?

ericglau:
Completed in #1099 and available in [@openzeppelin/hardhat-upgrades@3.8.0](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fhardhat-upgrades%403.8.0)

Issue #546, opened 2022-03-24
Polygon and Polygon Mumbai treated as unknown networks by the upgrades plugin 


ShivamArora:
The plugin recommends to gitignore files with the `unknown-*.json` pattern to avoid committing config meant for local networks.

Since Polygon and Polygon Mumbai networks are being treated as unknown networks by the upgrades plugin, this unknowingly gitignores the config meant to be tracked into the repository since these are live networks.

e.g. Contracts deployed to Polygon Mumbai testnet are being stored in `unknown-80001.json`.

As a result of this, developers and any organization may lose the important data stored in these config files.

PR #547, opened 2022-03-24
Add support for polygon and polygon mumbai to the manifest

ShivamArora:
Adds support for Polygon and Polygon Mumbai to the manifest.

Fixes: #546


fassko:
Thanks, I hope this can be merged at some point.

frangio:
Thanks @ShivamArora. This is more complicated because we need to check if an existing "unknown-id" file exist and if so migrate it to the new name. This PR currently does not accomplish that.

ShivamArora:
Thanks for your feedback @frangio . I agree this doesn't accomplish migrating existing state to the new file name.

I'll introduce changes which help migrate existing projects maintaining state with the `unknown-chainId` file names.

Can you please confirm if we don't need anything else for new projects since they would be tracking their state from scratch using the new filenames?

ShivamArora:
Hi @frangio , just added some code to rename the file if required in case of existing projects. Please review the same and let me know if any changes are required.

frangio:
Thanks! However, this implementation is not what we're looking for. Some extra requirements to keep in mind:

- File system interactions should be async
- Any changes to the manifest (including moving it!) need to first lock the file (see `lockedRun`)

I believe we should do the renaming in the `write` function.

I also would like us to change the lockfile name. We currently use a lockfile based on the file name itself, but this can change as we add new known network names.

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/d1f4c7acb3e1e72d18b0c23aaccfd58fac7efdda/packages/core/src/manifest.ts#L129-L130

This should look something like:

```typescript
await fs.mkdir(manifestDir, { recursive: true }); 
const release = await lockfile.lock(
  path.join(manifestDir, 'chain-' + this.chainId), // <- same lockfile regardless of known network or not
  { retries: 3, realpath: false },
); 
```

ShivamArora:
Thanks for the feedback @frangio . I'll incorporate the changes required.

Can you please also guide me to how can I test these changes within a hardhat package? 

It's a bit complicated and I can't directly install the upgrades package since we're maintaining a monorepo here.

I tried using gitpkg using the commit id but it didn't quite work out so had to push these changes without actually testing them.

frangio:
If you mean for manual testing, you should be able to use `yarn link`. You need to run `yarn link` in the `packages/plugin-hardhat` directory, and then `yarn link @openzeppelin/hardhat-upgrades` in your test project.

For automated testing, we don't need to test this in a Hardhat project. We should test the `Manifest` class directly. This is a bit complicated to test and we don't have a setup ready to test it, so if it sounds too difficult feel free to not include tests in this PR. If you want to set up automated tests, I think you should follow an approach similar to how we test this script, where we create a temp directory and chdir into it:

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/d1f4c7acb3e1e72d18b0c23aaccfd58fac7efdda/packages/core/src/scripts/migrate-oz-cli-project.test.ts#L32

ShivamArora:
Hi @frangio , got busy with some work earlier. I just took a note of your feedback and made some changes. Please review them and also go through some of the points I mentioned below stating the reasons some of the changes haven't been made as you suggested.

Let me know, if you think otherwise and if you've any other suggestions.

**Summary:**
- File system interactions are now async
- File rename happens within the `forNetwork` static method instead of `write` function.
- Lock function implementation stays as is with support for a parameterized file name.

**Rename within Write analysis:**
While trying to rename while writing the file, I realized that the renamed file would end up being empty. The reason for that is if a file already exists with unknown network name and we introduce the support for its network name, the read operations would be happening from the network name file and not the unknown network name file.

**Example:** 
If there's an existing project using Polygon Mumbai and is tracking the file `unknown-80001.json`. We now add support for `polygonMumbai.json` for the same.

When performing the read operation, it will always look out for the existence of `polygonMumbai.json` as our package now knows the chainId `80001` is mapped to `polygonMumbai` which it won't be able to find since it hasn't been migrated or renamed yet.

So we need to rename the file even before it's read to ensure it's reading the existing content but using the newer file name.

I did some analysis for the usage of `Manifest` class in the code and found out we're not directly using it's constructor anywhere and using the static `forNetwork` method to initialize it's instance which seems a good place to perform a rename.

**New naming convention for locked files**
I think we can't change the naming convention within the lock function since the lock should be placed on the file being read/written to.
If we start maintaining a separate convention for read/write operations vs lock operations, we would lose the benefits of file locking so that other processes can't modify the file at the same time when we're reading/writing.

ShivamArora:
I also just went ahead now to test this integration and it worked perfectly and renamed my existing file successfully keeping the original data intact.

Only thing I had to change was to link the `@openzeppelin/upgrades-core` package instead of `@openzeppelin/hardhat-upgrades`.

Thanks for your suggestion on testing using `yarn link` @frangio

ShivamArora:
Hi @frangio , did you get a chance to review this again?

frangio:
@ShivamArora Really sorry for the delay this is having. I will be able to get back to this next week. You've done great work so far, there's just a few edge cases I want to consider before merging.

ShivamArora:
Hi @frangio , do you have any update on this? I was thinking to add some more chains to the list but waiting on this to get merged first. If there are any more edge cases to be covered, please let me know.

ericglau:
Hi @ShivamArora, thanks for following up.  I am continuing to look into this and there are a few changes that I'd like to make.  If you don't mind, I will make the changes and add them into this PR.

Here is what we have in mind:
1. For the lock file name, we should make it always based on the chain ID rather than the network name.  This would avoid the possibility of using different lock files based on the old file name and new file name at the same time.
2. Ideally the renaming should occur in the `write` function instead of when instantiating.  Reason is that instantiating happens often and this would cause a lot of checks for the old file name, and (if there are many concurrent processes) would cause contention on the lock file until it has been renamed.  By using a consistent lock file name in (1), it should avoid the issues that you mentioned in your previous [comment about 'Rename within Write analysis'](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/547#issuecomment-1100612253).

For the above, [here is a basic test case](https://github.com/OpenZeppelin/openzeppelin-upgrades/commit/7fdd8727a78655aae524cfc3cb283f61d0005694) of how the implementation currently works, and [here is the updated version](https://github.com/OpenZeppelin/openzeppelin-upgrades/commit/9d90938155627c9926d6597cce7e4ba1a447f5d7) on how I think it should work.  I will make some changes so that the implementation satisfies the updated test case.

--- 

If you would like to add more chains to the list, you can go ahead since I think that part should not conflict with the changes that I am proposing above.  Thank you.

frangio (approved):
Thank you both!

frangio:
@ShivamArora Please make a separate PR to add more network names.

Issue #548, opened 2022-03-24
Defender `proposeUpgrade` method fails in this scenario.


DsAtHuH:
When running `yarn deploy rinkeby` on the [branch defender of this project](https://github.com/HUH-Token/huh-governance/tree/6728f1067a2641fc7df6ef09ae457b2758643565), defender fails to propose the upgrade at [line 17 of the upgrade script](https://github.com/HUH-Token/huh-governance/blob/6728f1067a2641fc7df6ef09ae457b2758643565/src/upgrade.js#L17) resulting in the following error:

```bash
❯ yarn deploy rinkeby
yarn run v1.22.17
$ yarn hardhat deploy --network rinkeby
$ node --require esm node_modules/.bin/hardhat deploy --network rinkeby
Nothing to compile
deploying "ERC20Mock" (tx: 0xac95267815d45f8842a5309965d37ec6db3f6925361673c31411918647159f82)...: deployed at 0xc8EF7026CfF4765f797543330681aB1eB5D97805 with 648507 gas
deploying "Timestamp" (tx: 0x913b47ba4421abec12eedc615905943cdb91c42b7ccb2ac3574a015511001efc)...: deployed at 0x347897a7C3963CF6d05ddb9e8Fe037ea4D081c28 with 127633 gas
deploying "HUHGovernance_Implementation" (tx: 0xa859fc45365048408696340faa77c100f01c584329900f21c8117ca44bebbac4)...: deployed at 0xc9428ea129EE105a137ADD6C19EE4611a00C9eaE with 2795949 gas
deploying "HUHGovernance_Proxy" (tx: 0xb98eba87ef5e10185e66353ea7de482d9151c0c97fa8ab4c3c34c5289f7c5517)...: deployed at 0x8e626F68B07b9CC68654e92d0C8e5BcC4210b1d9 with 393857 gas
Transferring ownership of ProxyAdmin...
Transferred ownership of ProxyAdmin to: 0x9f01873CE9eD346b2ae76eb36dD98ff0198DF9aA
Preparing proposal...
An unexpected error occurred:

Error: ERROR processing /home/dsathuh/Workspaces/HUH/huh-governance/deploy/03-upgrade-huh-governance.js:
[...]

PR #550, opened 2022-03-25
Support importing UUPS proxy with admin

ericglau:
For https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/548

- Instead of detecting transparent/UUPS proxies kind based on admin storage slot, infer the kind through existence of `upgradeTo(address)` in the provided implementation contract instead.  This will allow UUPS proxies that have an admin to be properly imported as UUPS rather than transparent.
- Support optional `kind` to allow user to force the proxy kind to be uups, transparent, or beacon


frangio (approved):
LGTM!

Issue #549, opened 2022-03-24
Migration Error: The requested contract was not found.


boyuanx:
I've never had a problem with Truffle upgrades until my current project. Migration gives the following error:
`Error: The requested contract was not found. Make sure the source code is available for compilation`

Things I've tried:
- Deleting build folder
- Deleting the local repo and cloning again
- Re-adding truffle-upgrades plugin
- Making sure the plugins are of the latest version

I've tried looking up this issue online and none of the posts helped. Need urgent help on this as this is blocking our entire project. Thanks!

boyuanx:
Yes - do I share it with you?

ericglau:
@boyuanx You can share it with me and I'll take a look.

ericglau:
I was able to reproduce the issue. Looking into it.

ericglau:
@boyuanx I've identified the problem in the upgrades plugin where contracts are ignored if multiple parent contracts have the same name.  

In this case, this happens because a contract imports `@openzeppelin/contracts-upgradeable/interfaces/IERC20Upgradeable.sol` which imports `@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol`.  We should fix the plugin to detect this situation.

As a workaround for now, you can change the following line in your contract:
```
import "@openzeppelin/contracts-upgradeable/interfaces/IERC20Upgradeable.sol";
```
to
```
import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
```

ericglau:
Minimal example to reproduce this issue:

Contract:
```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;
import "@openzeppelin/contracts-upgradeable/interfaces/IERC20Upgradeable.sol";
contract ImportIERC20  {
}
```
Migration:
```
const { deployProxy } = require("@openzeppelin/truffle-upgrades");
const ImportIERC20 = artifacts.require("ImportIERC20");
module.exports = async function (deployer) {
  await deployProxy(ImportIERC20, { deployer });
};
```

ericglau:
@boyuanx As this issue can be reproduced with my comment above, please go ahead and remove me from your repo.  Thank you for bringing this to our attention.

boyuanx:
Appreciate the quick response!

adeebabdulsalam:
I had this same issue with hardhat plugin.

steps to reproduce

in contracts directory
[...]

PR #551, opened 2022-03-29
Ignore AST for imports that have the same name as dependency

ericglau:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/549
Related to https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3218

If an import looks like an OpenZeppelin interface that transitively imports an interface at a different path with the same file name, ignore the AST and allow processing to continue.

### DIFF
--- a/packages/plugin-truffle/src/utils/validations.ts
+++ b/packages/plugin-truffle/src/utils/validations.ts
@@ -11,6 +11,7 @@ import {
 import type { SolcInput, SolcOutput, SolcLinkReferences } from '@openzeppelin/upgrades-core/dist/solc-api';
 
 import { TruffleArtifact, ContractClass, NetworkObject } from './truffle';
+import debug from './debug';
 
 export async function validateArtifacts(artifactsPath: string, sourcesPath: string): Promise<ValidationRunData> {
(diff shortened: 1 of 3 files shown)


packages/plugin-truffle/src/utils/validations.ts
```diff
@@ -86,18 +87,41 @@ function checkForImportIdConsistency(
...
         //
-        // 2. A contract was imported from a dependency using artifacts.require.
+        // 2. An imported file has only other import statements without an actual contract/library/interface.
+        //    Truffle does not output the AST for this file.
+        //    If this looks like an OpenZeppelin interface that transitively imports an interface at a different path,
+        //    ignore this import.
+        //
+        // The code below corresponds to scenario 2.
+        let ignoreImport = false;
+        for (const sourceUnitPath of Object.keys(sourceUnitId) as Array<string>) {
```
frangio, line 103: This doesn't look necessary:
```suggestion
        for (const sourceUnitPath of Object.keys(sourceUnitId)) {
```

frangio (approved):

PR #558, opened 2022-04-14
Consider rename annotation in original variable

### LINKED ISSUE #557: Storage upgrade reports false positives when old contract uses rename annotations
It seems that in `@openzeppelin/upgrades-core` version 1.14.1, the `UpgradeableContract.getStorageUpgradeReport()` reports false errors when analyzing a storage upgrade from a contract that annotates the storage variables with `@custom:oz-renamed-from`.

I [created a gist](https://gist.github.com/vvp/ff9dff52864bb761df31e7a7bcea940e) containing example contracts along with a Hardhat+Mocha test case and the `ops` array output. I have not tested if the `@custom:oz-retyped-from` tags behave the same.

frangio:
Fixes #557 
Needs a test, we can use https://gist.github.com/vvp/ff9dff52864bb761df31e7a7bcea940e


frangio:
@ericglau Can you merge and release this fix?

ericglau (approved):

Issue #10, opened 2020-07-07
Add comprehensive Buidler plugin tests


frangio:
There is currently only a simple test for the happy path.

PR #564, opened 2022-05-03
Add gap support

Amxx:
Providing that we have the full (solc) storage layout, this PR allows an upgrade to "eat into" the gaps when adding new variables.

Fixes: #276


packages/core/src/storage/compare.ts
```diff
@@ -72,7 +72,14 @@ function storageItemBegin(entry: StorageItemFull): number {
...
 function storageItemEnd(entry: StorageItemFull): number {
-  return storageItemBegin(entry) + Number(entry.type.item.numberOfBytes);
+  return storageItemBegin(entry) + padBytes(Number(entry.type.item.numberOfBytes));
+}
+
+/**
+ * Pad to nearest 32 bytes beyond the given number
+ */
+function padBytes(numBytes: number): number {
+  return numBytes % 32 > 0 ? Math.floor(numBytes / 32) * 32 + 32 : numBytes;
```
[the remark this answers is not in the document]
ericglau: This should be rounding up to the nearest multiple of 32.  So I simplified it to use [this](https://stackoverflow.com/a/3407124/13775848).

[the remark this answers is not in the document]
Amxx: Yes my bad, that is what I meant ((num +31) /32*32)

packages/core/src/storage/compare.ts
```diff
@@ -60,6 +60,42 @@ export interface LayoutChange {
...
+
+function storageItemEnd(entry: StorageItemFull): number {
+  return storageItemBegin(entry) + roundUpBytes(Number(entry.type.item.numberOfBytes));
+}
+
+/**
+ * Round up to nearest 32 bytes beyond the given number
+ */
+function roundUpBytes(numBytes: number): number {
+  return Math.floor((numBytes + 31) / 32) * 32;
```
Amxx, line 82:
```suggestion
  return Math.ceil(x / 32) * 32;
```

packages/core/src/storage/compare.ts
```diff
@@ -60,6 +60,42 @@ export interface LayoutChange {
...
+function isFullInfo(entry: StorageField): entry is StorageItemFull {
+  return entry.slot != undefined && entry.offset != undefined;
+}
+
+function storageItemBegin(entry: StorageItemFull): number {
+  return Number(entry.slot) * 32 + entry.offset;
+}
+
+function storageItemEnd(entry: StorageItemFull): number {
+  return storageItemBegin(entry) + roundUpBytes(Number(entry.type.item.numberOfBytes));
```
Amxx, line 75: I'm not sure about that. `storageItemBegin` is not necessary "aligned".

We could have an `entry` with slot: 10, offset: 4, number of bytes: 4.
The beginning is at `10*32+4=324`  and the end is at `10*32+4+4=328`. The entry is in slot #10, which goes from 320 to 351

Yet the roundUp will say that is starts at 324, and that is ends at 356, which is wrong.

I'm not sure where the `roundUpBytes` is actually needed, but I think we should have 
```
function storageItemEnd(entry: StorageItemFull): number {
  return storageItemBegin(entry) + Number(entry.type.item.numberOfBytes);
}
```
and possibly doing `roundUpBytes(storageItemEnd(entry))` in some places.

packages/core/src/storage/compare.ts
```diff
@@ -60,6 +60,42 @@ export interface LayoutChange {
...
+// Get a subset of `layout` that exactly covers the space from `begin` to `end`
+function subLayout(begin: number, end: number, layout: StorageItemFull[]): StorageItemFull[] | undefined {
+  if (begin === end) {
+    return [];
+  }
+
+  const sublayout = layout.filter(entry => begin <= storageItemBegin(entry) && storageItemEnd(entry) <= end);
+  return sublayout.length > 0 &&
+    begin === storageItemBegin(sublayout[0]) &&
+    (end === storageItemEnd(sublayout[sublayout.length - 1]) || end === Infinity)
```
Amxx, line 94: I think this is where we want to do
```
end === roundUpBytes(storageItemEnd(sublayout[sublayout.length - 1])) || end === Infinity)
```

packages/core/src/storage/compare.ts
```diff
@@ -70,7 +106,64 @@ export class StorageLayoutComparator {
...
+
+        // Merge consecutive gaps
+        for (; i < gaps.length - 1; ++i) {
+          if (storageItemEnd(gaps[i]) !== storageItemBegin(gaps[i + 1])) {
+            break;
+          }
+        }
+
+        // End of the gap(s) is the begining of the next section
+        const gapEnd = storageItemEnd(gaps[i]);
```
Amxx, line 129: I think we are making assumptions that the gap has the right format. If someone was to put `bool private __gap` in their system, it would possibly break things.

Maybe we need to check in the filter (line 116) that the gap is a struct (that enforces begin and end are aligned with slots)

packages/core/src/storage/compare.ts
```diff
@@ -70,7 +106,64 @@ export class StorageLayoutComparator {
...
+      if (original.length > 0 && ptr < storageItemEnd(original[original.length - 1])) {
+        sections.push({ begin: ptr, end: Infinity });
+      }
+
+      // Now that sections are isolated, we can compare them one by one
+      const ops = sections.flatMap(({ begin, end }) => {
+        const originalSection = subLayout(begin, end, original) || [];
+        const updatedSection = subLayout(begin, end, updated) || [];
+        return this.layoutLevenshtein(originalSection, updatedSection, { allowAppend: end === Infinity });
+      });
```
Amxx, line 161:
```suggestion
      const ops = sections.flatMap(({ begin, end }) => this.layoutLevenshtein(
            subLayout(begin, end, original) || [],
            subLayout(begin, end, updated) || [],
            { allowAppend: end === Infinity },
      ));
```

packages/core/src/storage/compare.ts
ericglau: Should non-array gaps be considered a gap?
e.g. `bool private __gap` or `uint256 private __gap`

ericglau: Added a check to ensure the gap is an array.  Is struct relevant?

Amxx: I think this is good !

packages/core/src/storage/compare.ts
```diff
@@ -60,6 +60,42 @@ export interface LayoutChange {
...
+}
+
+function storageItemEnd(entry: StorageItemFull): number {
+  return storageItemBegin(entry) + Number(entry.type.item.numberOfBytes);
+}
+
+/**
+ * Aligns to the next storage slot by rounding up to nearest 32 bytes beyond the given number.
+ */
+function alignToStorageSlot(numBytes: number): number {
```
Amxx, line 81: Maybe this should be named `alignToNextStorageSlot` or just `nextStorageSlot`.

ericglau:
Reworked this PR to use the design described [here](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/276#issuecomment-1217291102).

packages/core/src/storage/report-gap.test.ts.md
```diff
@@ -0,0 +1,45 @@
...
+      - Bad storage gap resize from 48 to 47␊
+        Size decrease must match with corresponding variable inserts`
+
+## insert var and shrink gap too much
+
+> Snapshot 1
+
+    `file.sol:1: Upgraded \`__gap\` to an incompatible type␊
+      - Bad storage gap resize from 48 to 46␊
+        Size decrease must match with corresponding variable inserts`
```
frangio, line 39: Same here I think we should be able to calculate and report the actual size it should have.

packages/core/src/storage/report-gap.test.ts.md
```diff
@@ -0,0 +1,45 @@
...
+
+## insert vars and shrink gap not enough
+
+> Snapshot 1
+
+    'file.sol:1: Replaced `__gap` with `d` of incompatible type'
```
frangio, line 45: I find it weird that we're seeing this error here. Makes me think we should give lower cost to `typechange`.

packages/core/contracts/test/Storage.sol
```diff
@@ -431,4 +431,256 @@ contract StorageUpgrade_StructEnum_V2 {
...
+    address a;
+    address b;
+    address c;
+    address[47] __gap;
+    address z;
+}
+
+contract StorageUpgrade_Uint128Gap_V1 {
+    uint128 a;
+    uint128[49] __gap;
```
Amxx, line 582: Its nice if we support that, but gaps made of array that are not `bytes32[]` or `uint256[]` are not essential to support

Amxx:
Something we don't test is

```
contract V1 {
    struct S {
        uint256 a;
        uint256 b;
        bytes32[50] __gap;
        uint256 d;
    }
    mapping(uint256 => S) store_mapping;
    S[10] store_fixed_array;
    S[] store_dynamic_array;
}
contract V2_Ok {
    struct S {
        uint256 a;
        uint256 b;
        uint256 c;
        bytes32[49] __gap;
        uint256 d;
    }
    mapping(uint256 => S) store;
    S[10] store_fixed_array;
    S[] store_dynamic_array;
}
```

Is that supported ?

packages/core/src/storage/compare.ts
```diff
@@ -80,10 +115,44 @@ export class StorageLayoutComparator {
...
+            return true;
+          }
+
+          // An insertion that overlaps with a non-gap in the original layout is not allowed
+          if (!isGap(orig) && overlaps(origStart, origEnd, updatedStart, updatedEnd)) {
+            return true;
+          }
+        }
+        // Otherwise the insertion is safe
+        return false;
```
Amxx, line 143:
```suggestion
        // If any entry (that is not a gap) in the original layout overlaps, then the insertion is not safe
        return original
            .filter(entry => !isGap(entry))
            .some(entry => {
                const origStart = storageFieldBegin(orig);
                const origEnd = storageFieldEnd(orig);
                return origStart === undefined || origEnd === undefined || overlaps(origStart, origEnd, updatedStart, updatedEnd);
            });

```

Amxx, line 143:
```suggestion
        const updatedStart = storageFieldBegin(o.updated);
        const updatedEnd = storageFieldEnd(o.updated);
        // If we cannot get the updated position, or if any entry in the original layout (that is not a gap) overlaps, then the insertion is unsafe.
        return updatedStart === undefined || updatedEnd === undefined || original
            .filter(entry => !isGap(entry))
            .some(entry => {
                const origStart = storageFieldBegin(orig);
                const origEnd = storageFieldEnd(orig);
                return origStart === undefined || origEnd === undefined || overlaps(origStart, origEnd, updatedStart, updatedEnd);
            });
```

packages/core/src/storage/compare.ts
```diff
@@ -108,7 +168,11 @@ export class StorageLayoutComparator {
     const layoutChange = this.getLayoutChange(original, updated);
 
     if (updated.retypedFrom && layoutChange) {
-      return { kind: 'layoutchange', original, updated, change: layoutChange };
+      return { kind: 'layoutchange', original, updated, change: layoutChange, cost: LAYOUTCHANGE_COST };
+    } else if (nameChange && endMatchesGap(original, updated)) {
+      return { kind: 'finishgap', original, updated, cost: FINISHGAP_COST };
+    } else if (typeChange && typeChange.kind === 'array shrink' && endMatchesGap(original, updated)) {
```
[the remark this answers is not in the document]
Amxx: I'm not really sure I understand this case, and the previous one. The fact that `isGap(original)` is hidden in the `endMatchesGap` doesn't help.

How do we know that `updated` is a gap?

packages/core/src/storage/compare.ts
```diff
@@ -108,25 +177,31 @@ export class StorageLayoutComparator {
...
       // add this check as a safety fallback.
-      return { kind: 'layoutchange', original, updated, change: layoutChange };
+      return { kind: 'layoutchange', original, updated, change: layoutChange, cost: LAYOUTCHANGE_COST };
     }
   }
 
   getLayoutChange(original: StorageField, updated: StorageField): LayoutChange | undefined {
     const validPair = ['uint8', 'bool'];
     const knownCompatibleTypes =
       validPair.includes(original.type.item.label) && validPair.includes(updated.type.item.label);
```
Amxx, line 203: I know that is not part of this PR, but do we really think there is only one valid pair, and that they are always pairs, we'll ever consider?

I would have done 
```suggestion
    const validChanges = [['uint8', 'bool']];
    const knownCompatibleTypes = validChanges.some(group => goup.includes(original.type.item.label) && group.includes(updated.type.item.label));
```

Also, that means a bool can become a uint8, but also that a uint8 can become a bool. Is that ok ?

packages/core/contracts/test/Storage.sol
ericglau: Agree, although the implementation does not treat different types in any special way so they should all be supported.

Added tests for `bytes32[]` gaps.

ericglau:
> Something we don't test is
> 
> ```
> contract V1 {
>     struct S {
>         uint256 a;
>         uint256 b;
>         bytes32[50] __gap;
>         uint256 d;
>     }
>     mapping(uint256 => S) store_mapping;
>     S[10] store_fixed_array;
>     S[] store_dynamic_array;
> }
> contract V2_Ok {
>     struct S {
>         uint256 a;
>         uint256 b;
>         uint256 c;
>         bytes32[49] __gap;
>         uint256 d;
>     }
>     mapping(uint256 => S) store_mapping;
>     S[10] store_fixed_array;
>     S[] store_dynamic_array;
> }
> ```
> 
> Is that supported ?

This is supported.  Added tests for this.

packages/core/src/storage/compare.ts
```diff
@@ -80,10 +106,44 @@ export class StorageLayoutComparator {
...
+    return ops.filter(o => {
+      if (o.kind === 'insert') {
+        const updatedStart = storageFieldBegin(o.updated);
+        const updatedEnd = storageFieldEnd(o.updated);
+        if (isNaN(updatedStart) || isNaN(updatedEnd)) {
+          // unable to get storage position, so treat this as unsafe
+          return true;
+        }
+
+        for (let i = 0; i < original.length; i++) {
```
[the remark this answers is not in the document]
ericglau: Refactored based on [this suggestion](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/564#discussion_r957334743) which makes the code quite readable at the moment.

If efficiency is important, we could change it to stop when it goes past the range.  

But I'm not sure if a binary search is possible: currently we treat the whole thing as unsafe (when we are searching sequentially) if a variable does not have position info; could a binary search miss these?

packages/core/src/storage/compare.ts
ericglau: Reworked this part of the code to include this suggestion and also ensure that it does not omit layout changes for these valid pairs.

packages/core/src/storage/compare.ts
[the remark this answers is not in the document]
frangio: Yes but that never happens. Either we have position info for all variables or we don't have it at all. This is a problem in our types, they aren't properly expressing that invariant. This is something we should tackle in a future PR.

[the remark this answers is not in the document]
frangio: On second thought I don't know if it's possible to improve the types in that way, at least without a significant refactor of the code.

docs/modules/ROOT/pages/writing-upgradeable.adoc
```diff
@@ -452,4 +452,6 @@ contract Base {
     uint128 base2b; // 16 bytes - continues from the same slot as above
     uint256[48] __gap;
 }
-----
\ No newline at end of file
+----
+
+To help determine the proper storage gap size in the new version of your contract, you can write a test that uses the `validateUpgrade()` function from xref:api-hardhat-upgrades.adoc#validate-upgrade[Hardhat Upgrades] or xref:api-truffle-upgrades.adoc#validate-upgrade[Truffle Upgrades]. If a storage gap is not being reduced properly, you will see an error message indicating the expected size of the storage gap.
```
frangio, line 457:
```suggestion
To help determine the proper storage gap size in the new version of your contract, you can simply attempt an upgrade using `upgradeProxy` or just run the validations with `validateUpgrade` (see docs for xref:api-hardhat-upgrades.adoc[Hardhat] or xref:api-truffle-upgrades.adoc[Truffle]). If a storage gap is not being reduced properly, you will see an error message indicating the expected size of the storage gap.
```

packages/core/src/storage/report.ts
```diff
@@ -42,10 +52,47 @@ interface StorageOperationContext {
...
+    origEnd === undefined ||
+    updatedStart === undefined ||
+    origNumBytes === undefined ||
+    origArraySize === undefined
+  ) {
+    return undefined;
+  }
+
+  // use the floor since number of bytes may be larger than actual array (for non 32-byte types due to unused space after the array before the next slot)
+  const bytesPerItem = Math.floor(parseInt(origNumBytes) / parseInt(origArraySize, 10));
```
frangio, line 76: BigInt division is integer division so it automatically applies floor. I removed the comment because it confused me more than it helped.

```suggestion
  const bytesPerItem = BigInt(origNumBytes) / BigInt(origArraySize);
```

frangio (approved):
:rocket:

packages/core/src/storage/report.ts
ericglau: Changed it to `BigInt(parseInt(origTail, 10))` instead of `BigInt(origArraySize)` since the tail is a string like `49_storage`

PR #566, opened 2022-05-06
Support unsafeSkipStorageCheck option in ValidationOptions

miloops:
Add `unsafeSkipStorageCheck` into `ValidationOptions` which was added in [@openzeppelin/hardhat-upgrades@1.16.0](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fhardhat-upgrades%401.16.0)

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Support unsafeSkipStorageCheck option in ValidationOptions. ([#566](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/566))
+
 ## 1.14.2 (2022-04-18)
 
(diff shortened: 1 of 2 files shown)


ericglau (changes requested):
Thanks for this PR!  Just a couple of comments:

- Can you please revert all files other than `packages/core/src/validate/overrides.ts`, since those versions will be bumped as part of the release process.

- Please add a changelog entry at the top of `packages/core/CHANGELOG.md`.  The version heading can be `## Unreleased`

miloops:
> Thanks for this PR! Just a couple of comments:
> 
> * Can you please revert all files other than `packages/core/src/validate/overrides.ts`, since those versions will be bumped as part of the release process.
> * Please add a changelog entry at the top of `packages/core/CHANGELOG.md`.  The version heading can be `## Unreleased`

Thanks for your feedback and prompt response, I have updated the PR accordingly.

ericglau (approved):
Thank you!

PR #571, opened 2022-05-13
include openethereum dev chain in recognised dev chains

jamie-arpeggi:


Amxx:
Note: it looks like there is some network using that chain id
https://chainlist.org/chain/17

ericglau:
> Note: it looks like there is some network using that chain id https://chainlist.org/chain/17

I don't see any [blocks on that chain](https://exp.thaifi.com/blocks) in the past 6 months so I think this is fine.

ericglau (approved):
Thanks for the PR

ericglau:
@jamie-arpeggi OpenEthereum is now deprecated (references: [1](https://github.com/openethereum/openethereum), [2](https://twitter.com/OpenEthereumOrg/status/1529147048758595585)) so we've reverted this change.  Thanks for your understanding.

Issue #573, opened 2022-05-16
Dry-run mode in prepareUpgrade


kargakis:
Currently, if we want to dry-run an upgrade just to validate it we have to stitch together logic from `prepareUpgrade` which is cumbersome as there is a lot of conditional code that is already in place. It would be nice to enable dry-running `prepareUpgrade` to allow for more robust processes downstream. The other workaround for now is to run in forking mode and have custom logic that copies the network file at the right place which is also not ideal.

ericglau:
https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/580 will cover this use case with a different approach.

Closing this issue as described in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/578#issuecomment-1133323217

PR #578, opened 2022-05-18
Add dry-run mode in prepareUpgrade

kargakis:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/573

### DIFF
--- a/packages/plugin-hardhat/src/utils/deploy-impl.ts
+++ b/packages/plugin-hardhat/src/utils/deploy-impl.ts
@@ -116,6 +116,9 @@ async function deployImpl(
     deployData.provider,
     async () => {
       const abi = ImplFactory.interface.format(FormatTypes.minimal) as string[];
+      if (opts.dryRun === true) {
+        return { abi, layout, address: currentImplAddress ?? '' };
+      }
       const deployment = Object.assign({ abi }, await deploy(ImplFactory, ...deployData.fullOpts.constructorArgs));
       return { ...deployment, layout };
     },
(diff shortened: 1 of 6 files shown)


kargakis:
Marking as WIP, I still need to fix the truffle test.

kargakis:
@ericglau @Amxx I've completed the unit test changes I wanted, please take a look, thanks!

ericglau:
Hi @kargakis, thank you for taking a look into this.  I don't think we should take this implemented approach since there are a few potential issues:
- `deploy-impl` is used by a number of functions aside from just `prepareUpgrade`, for example it is also used by `deployProxy` or `upgradeProxy`.  So this flag would cause the implementation to not be deployed, but the other functions would still proceed to do other things like deploy or upgrade the proxy which would fail (since the new implementation is not deployed and has address as '')
- The network file is also getting updated during the dryRun, which should not happen.  More specifically, since the current address is being passed in to `{ abi, layout, address: currentImplAddress ?? '' };`, that causes the plugin to assign the new storage layout to the existing address, which is wrong.

Overall, I propose a different approach.  I've opened a high level feature here https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/580 to make the plugins have more granular commands, specifically a way to validate an implementation (whether for a first implementation or for an upgrade).  That is the same goal as what you are trying to achieve, and I think that approach would work better with the different scenarios.

I will go ahead and close this issue but feel free to continue the discussion here or in https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/580.  If you are interested, feel free to try taking on the `validateImplementation` and/or `validateUpgrade` functions proposed there.  Those would be new functions but would be like cut-down versions of what `prepareUpgrade` currently does.

PR #581, opened 2022-05-24
Set node version for builds

ericglau:


frangio:
IMO it's better to test with the oldest supported version. In our case it's 14.x, currently ["maintenance LTS"](https://nodejs.org/en/about/releases/).

ericglau:
GitHub actions will use 14.x but the netlify build would need to use 16 due to the dependency on the docs site and https://github.com/OpenZeppelin/docs.openzeppelin.com/pull/319

Issue #554, opened 2022-04-06
Return ethers TransactionResponse for defender.proposeUpgrade


DsAtHuH:
After running `defender.proposeUpgrade` method from the [defender branch of HUH Governance](https://github.com/HUH-Token/huh-governance/tree/851670531c2d0e0c0d8dfd1aa48883fa0d5dc159) by running `yarn deploy rinkeby` the following error was triggered on the [verification task](https://github.com/HUH-Token/huh-governance/blob/851670531c2d0e0c0d8dfd1aa48883fa0d5dc159/src/upgrade.js#L36) just after the upgrade proposal was created:

```bash
Error in plugin @nomiclabs/hardhat-etherscan: Failed to send contract verification request.
Endpoint URL: https://api-rinkeby.etherscan.io/api
Reason: The Etherscan API responded that the address 0x76E12006dCDBdb62b591B0E87315ceC2a5Aae799 does not have bytecode.
This can happen if the contract was recently deployed and this fact hasn't propagated to the backend yet.
Try waiting for a minute before verifying your contract. If you are invoking this from a script,
try to wait for five confirmations of your contract deployment transaction before running the verification subtask.
```

When looking for the proposal object returned [here](https://github.com/HUH-Token/huh-governance/blob/851670531c2d0e0c0d8dfd1aa48883fa0d5dc159/src/upgrade.js#L24), I was expecting something similar to a `TransactionResponse` mentioned in [this answer in StackOverflow](https://stackoverflow.com/a/69013840). With a `TransactionResponse` coming from the deployment of the new implementation that takes place inside the `proposeUpgrade` method I would be able to write:

```javascript
const receipt = await tx.wait(5)
```

in order to programatically solve the requirement of Etherscan API.

frangio:
This sounds reasonable. We currently return Defender's response object though, but we could add one more property to this object with the tx response.

DsAtHuH:
> This sounds reasonable. We currently return Defender's response object though, but we could add one more property to this object with the tx response.

Is there already an estimation of when this feature will be available? In the meanwhile I can insert one minute delay as a workaround...

ericglau:
Hi @DsAtHuH, we are looking into this but don't have an estimate at the moment.

ericglau:
This is available in the latest release.

PR #584, opened 2022-06-01
Return ethers transaction response with defender.proposeUpgrade

ericglau:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/554


frangio (approved):
I made some changes to the types so we could better reflect the deployment types and remove `deployment as any`. Aside from my OCD about `any` I was genuinely not sure why `deployTransaction` would be in the deployment until I looked deeper into the deployment code and I think the new types better reflect that. Let me know what you think about these changes.

PR #603, opened 2022-06-24
Simplify GitHub actions

frangio:
Uses a [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) for shared setup steps. Merges tests and coverage under a single workflow. Removes node-gyp-cache which I haven't been maintaining and it shouldn't be necessary anymore (native dependencies [no longer used](https://medium.com/nomic-foundation-blog/turning-a-page-in-ethereum-javascript-history-4ec89136fccc) in the ecosystem). Limits concurrency so that redundant in-progress workflow runs are cancelled.

We will need to change the branch protection rules with respect to required checks because the names are slightly different.


ericglau (approved):

frangio:
Ok I've changed the branch protection requirements. Open PRs will have to merge master so they can get the appropriate checks for merging.

Issue #580, opened 2022-05-20
Provide granular functions to allow more customizable deployments


ericglau:
Background
--

The current functions for [Hardhat](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades) and [Truffle](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-truffle-upgrades) such as `deployProxy`, `upgradeProxy`, and `prepareUpgrade` can be considered as all-in-one commands that do a number of things together.

In the case of `deployProxy` for example, it does the following:
- Checks that the implementation is upgrade safe
- Deploys the implementation
- Deploys the admin if needed for transparent proxy
- Deploys the proxy to use the implementation and admin
- Saves the above deployments and the implementation's storage layout to the network file

For `upgradeProxy` or `prepareUpgrade`, these also verify that the new implementation is upgrade safe when compared to the old implementation, in addition to deploying the implementation contract (and then upgrading the proxy in the case of `upgradeProxy`).

Problem statement
--

The all-in-one nature of the existing commands make it hard for users to customize their deployments.  

For example, users may want to customize [gas parameters](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/85) for specific contract deployments, verify [their implementation without deploying it](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/573), use [different signers](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/560), or deploy [only the proxy](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/464).

Proposal
--

Provide granular functions representing the different stages of implementation or proxy deployments or upgrades.  This would allow advanced users to do each step independently and can customize when or how each contract is deployed.

The following enhancements are proposed for the different categories related to this:

### Implementation contract

Since one of the major benefits of using the Upgrades Plugins is validating the implementation for upgrade safety (for the first version and later versions of the implementation), this should at minimum include the following functions:
1. `validateImplementation(Contract)` - Only validates the implementation for upgrade safety.  Similar to `prepareUpgrade` but for a first version of a contract, without deploying it, and without updating the network file.
2. `validateUpgrade(proxyOrBeaconAddress, Contract) or validateUpgrade(Contract, Contract)` - Validates the implementation for upgrade safety and compares its storage layout with the old version of the implementation for storage conflicts. Similar to `prepareUpgrade` but without deploying it, and without updating the network file.
[...]

PR #608, opened 2022-07-12
Use granular functions

ericglau:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/580
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/609

Status:
- [x] Hardhat
- [x] Truffle
- [x] Docs


packages/plugin-hardhat/src/utils/deploy-impl.ts
```diff
@@ -59,67 +53,63 @@ export async function getDeployData(
...
 
-async function deployImpl(
-  hre: HardhatRuntimeEnvironment,
+async function fetchOrDeployImpl(
   deployData: DeployData,
   ImplFactory: ContractFactory,
-  opts: PrepareUpgradeOptions,
-  currentImplAddress?: string,
+  opts: DeployImplementationOptions,
+  hre: HardhatRuntimeEnvironment,
```
frangio, line 92: Make this the first argument for consistency.

packages/plugin-hardhat/src/validate-upgrade.ts
```diff
@@ -0,0 +1,57 @@
...
+      const deployData = await getDeployData(hre, newImplFactory, opts);
+      if (await isTransparentOrUUPSProxy(provider, proxyOrBeaconAddress)) {
+        await validateProxyImpl(deployData, opts, proxyOrBeaconAddress);
+      } else if (await isBeaconProxy(provider, proxyOrBeaconAddress)) {
+        const beaconAddress = await getBeaconAddress(provider, proxyOrBeaconAddress);
+        await validateBeaconImpl(deployData, opts, beaconAddress);
+      } else if (await isBeacon(provider, proxyOrBeaconAddress)) {
+        await validateBeaconImpl(deployData, opts, proxyOrBeaconAddress);
+      } else {
+        throw new ValidateUpgradeUnsupportedError(proxyOrBeaconAddress);
```
frangio, line 53: It really sounds like we should be using the same Error instance here and in `prepareUpgrade` (and potentially in other places?).

I get that it's currently named `PrepareUpgradeUnsupportedError`, but we can rename it, add `export const PrepareUpgradeUnsupportedError = __newName__` for backwards compat.

packages/plugin-hardhat/src/force-import.ts
```diff
@@ -30,28 +29,29 @@ export interface ForceImportFunction {
...
-      const beaconImplAddress = await getImplementationAddressFromBeacon(provider, proxyOrBeaconAddress);
+      return ImplFactory.attach(address);
+    } else if (await isBeacon(provider, address)) {
+      const beaconImplAddress = await getImplementationAddressFromBeacon(provider, address);
       await addImplToManifest(hre, beaconImplAddress, ImplFactory, opts);
 
       const UpgradeableBeaconFactory = await getUpgradeableBeaconFactory(hre, ImplFactory.signer);
-      return UpgradeableBeaconFactory.attach(proxyOrBeaconAddress);
+      return UpgradeableBeaconFactory.attach(address);
     } else {
```
frangio, line 52: Do we at least know whether there is a contract at this address? Not sure if that's something we should require but it seems like a reasonable sanity check.

packages/plugin-hardhat/src/deploy-implementation.ts
```diff
@@ -0,0 +1,24 @@
...
+) => Promise<DeployImplementationResponse>;
+
+export type DeployImplementationResponse = string | ethers.providers.TransactionResponse;
+
+export function makeDeployImplementation(hre: HardhatRuntimeEnvironment): DeployImplementationFunction {
+  return async function deployImplementation(ImplFactory, opts: DeployImplementationOptions = {}) {
+    const deployedImpl = await deployStandaloneImpl(hre, ImplFactory, opts);
+
+    if (opts.getTxResponse && deployedImpl.txResponse !== undefined) {
+      return deployedImpl.txResponse;
```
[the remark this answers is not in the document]
ericglau: If an implementation was imported by `forceImport`, its tx hash is not stored in the manifest, so the tx response is not available and we will need to return just the address.  So we should keep `getTxReponse` as an option to be consistent with `prepareUpgrade`.

packages/plugin-hardhat/src/validate-upgrade.ts
ericglau: Instead of throwing an error, we could just support this condition where we validate an upgrade where the original implementation address was passed in as a parameter (as long as the impl is in the manifest).  Changed it to do this.

frangio (approved):
Overall looks good to me, just minor comments.

frangio (approved):
LGTM!

PR #614, opened 2022-07-23
Clarify the usage of isTopLevelCall and Initializing

fongchinghinunsw:
The meaningfulness of these two variables might not be obvious at first, added comments for newcomers to understand the contract better.

### DIFF
--- a/packages/core/contracts/Initializable.sol
+++ b/packages/core/contracts/Initializable.sol
@@ -33,6 +33,9 @@ contract Initializable {
   modifier initializer() {
     require(initializing || isConstructor() || !initialized, "Contract instance has already been initialized");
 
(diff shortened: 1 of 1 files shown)


frangio:
Thanks @fongchinghinunsw. This contract is here for backwards compatibility reasons. Users should instead opt for [`@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol`](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/release-v4.7/contracts/proxy/utils/Initializable.sol). Note that the contract has evolved a little bit as it now has "reinitializers".

The file you're modifying should be considered deprecated so we shouldn't make any changes to it, other than adding a deprecation notice.

PR #622, opened 2022-08-03
Add support for avalanche and fuji to manifest file names

ShivamArora:


ericglau (approved):
Thanks for the PR!  Note: I've changed the fuji testnet to use the convention `<networkname>-<testnetname>` (to be consistent with polygon-mumbai for example).

PR #660, opened 2022-09-23
Override verify:verify subtask from hardhat-etherscan

### LINKED ISSUE #619: Override verify:verify subtask for Etherscan verification
When verifying a regular contract programmatically using hardhat-etherscan, the [verify:verify subtask must be used](https://hardhat.org/hardhat-runner/plugins/nomiclabs-hardhat-etherscan#using-programmatically).

When verifying a proxy using the Hardhat Upgrades plugin, the plugin only overrides `verify` but not `verify:verify` which is inconsistent with the usage for regular contracts.  See https://forum.openzeppelin.com/t/cannot-auto-verify-proxy-contract-with-implementation-and-proxyadmin/30305

The Hardhat Upgrades plugin should override `verify:verify` for proxies to make this consistent.

ericglau:
Fixes #619

### DIFF
--- a/docs/modules/ROOT/pages/api-hardhat-upgrades.adoc
+++ b/docs/modules/ROOT/pages/api-hardhat-upgrades.adoc
@@ -402,9 +402,9 @@ function silenceWarnings()
 [[verify]]
 == verify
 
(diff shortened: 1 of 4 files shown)


frangio (approved):
LGTM

PR #658, opened 2022-09-20
Add support for user defined value types

frangio:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/438
Fixes #477 

User defined value types are not supported without storage layout info, this is fine because they only exist in >=0.8.8, and I think we always have storage layout info for those versions. Unsure about Truffle to be honest, do we have storage layout in Truffle?

Issue #438, opened 2021-10-12
Detect storage incompatibilites due to bug in Solidity 0.8.8


frangio:
Packing of variables of user defined value types is different in 0.8.8 than it is in subsequent versions. See [User Defined Value Types Bug](https://blog.soliditylang.org/2021/09/29/user-defined-value-types-bug/).

The easiest thing is to reject Solidity 0.8.8.

PR #662, opened 2022-09-27
Log solc version in manifest

frangio:
Related to https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/658#issuecomment-1252951724.

If we want a chance to support UDVTs in Truffle in the future, we will need the solc version used for a deployed implementation because of #438. This simply adds that information in the StorageLayout objects in the manifest.


packages/plugin-truffle/src/utils/validations.ts
```diff
@@ -41,6 +47,21 @@ function reconstructSolcInputOutput(artifacts: TruffleArtifact[]): { input: Solc
...
     }
+    if (artifact.compiler.name !== 'solc') {
+      continue;
+    }
+
+    const versionMatch = artifact.compiler.version.match(/^([0-9.]+)\+commit/);
+    if (!versionMatch) {
+      throw new UpgradesError('Unknown or nightly compiler version');
+    }
+    const [artifactSolcVersion] = versionMatch;
```
ericglau, line 58: Currently this captures the whole match so it appears as `0.5.16+commit` in the manifest.  Was this intended to only capture the first group? 
```suggestion
    const artifactSolcVersion = versionMatch[1];
```

ericglau:
Should this include changelog entries (in core, hardhat, and truffle)?

ericglau (approved):

Issue #665, opened 2022-09-28
How to silence 2 storage layout warnings for the same variable


Ramarti:
Hi, I'm deprecating a variable, formerly typed as an interface, and I'm having problems passing the storage layout validation

```
/// @custom:oz-renamed-from _router
/// @custom:oz-retyped-from IRouter
address private _deprecated_router;
```

I'm getting 
```contracts/components/utils/Routed.sol:18: Replaced `_router` with `_deprecated_router` of incompatible type.```

It seems upgrades plugin only parses one of the comment, not both. If I remove the `renamed` decorator, I get the corresponding error to having renamed the variable.

Is there a way to signal the plugin that these 2 things are happening and it's fine? I want to retype to address because `IRouter` is no longer in the codebase.

I'm using:
```"@openzeppelin/hardhat-upgrades": "1.19.1",```

PR #666, opened 2022-09-28
Ignore whitespace in renamed/retyped annotations

frangio:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/665 and adds test for extraction of these annotations.


packages/core/src/storage-0.8.test.ts
```diff
@@ -110,3 +110,8 @@ test('user defined value types - no layout info - bad upgrade', async t => {
     },
   });
 });
+
+test('renamed retyped - extraction', async t => {
```
ericglau, line 114: Is this test case related in any way to solidity 0.8 (since this is in `storage-0.8.test.ts`)?

ericglau (approved):

packages/core/src/storage-0.8.test.ts
frangio: `@custom` annotations were added in Solidity 0.8.2.

PR #671, opened 2022-10-03
Consider UDVTs when stabilizing type ids

frangio:
Found this issue via https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/669.

The relevant function is used mainly for testing but also in `manifest-storage-layout.ts` as part of `getUpdatedStorageLayout`. I believe this should not be affecting users because `getUpdatedStorageLayout` is only triggered when the layout in the manifest is a previous version of the layout, and if there are UDVTs it will necessarily be the most recent version.

This implies there is no need to release a patch.


ericglau (approved):

ericglau:
If this does not affect users, is a changelog entry not needed?

PR #675, opened 2022-10-18
Update writing-upgradeable.adoc

mickayz:
The current example code demonstrating unchained functions will fail. The init function uses the modifier initializer and then calls the unchained method that also has the initializer modifier.  I've updated this code example to use onlyInitializing on the unchained method instead.

### DIFF
--- a/docs/modules/ROOT/pages/writing-upgradeable.adoc
+++ b/docs/modules/ROOT/pages/writing-upgradeable.adoc
@@ -141,7 +141,7 @@ contract ERC20Upgradeable is Initializable, ContextUpgradeable, IERC20Upgradeabl
         __ERC20_init_unchained(name, symbol);
     }
 
-    function __ERC20_init_unchained(string memory name, string memory symbol) internal initializer {
+    function __ERC20_init_unchained(string memory name, string memory symbol) internal onlyInitializing {
         _name = name;
         _symbol = symbol;
         _decimals = 18;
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/writing-upgradeable.adoc
```diff
@@ -454,4 +454,4 @@ contract Base {
 }
 ----
 
-To help determine the proper storage gap size in the new version of your contract, you can simply attempt an upgrade using `upgradeProxy` or just run the validations with `validateUpgrade` (see docs for xref:api-hardhat-upgrades.adoc[Hardhat] or xref:api-truffle-upgrades.adoc[Truffle]). If a storage gap is not being reduced properly, you will see an error message indicating the expected size of the storage gap.
\ No newline at end of file
+To help determine the proper storage gap size in the new version of your contract, you can simply attempt an upgrade using `upgradeProxy` or just run the validations with `validateUpgrade` (see docs for xref:api-hardhat-upgrades.adoc[Hardhat] or xref:api-truffle-upgrades.adoc[Truffle]). If a storage gap is not being reduced properly, you will see an error message indicating the expected size of the storage gap.
```
mickayz, line 457: Not sure why this shows a change in line 457 since the context was unchanged

frangio: Probably a newline was added or removed.

frangio (approved):
Thanks @mickayz! Sorry we missed this earlier. I've udpated the links to the latest version and also changed the snippets to correspond to the current code even more closely.

PR #682, opened 2022-10-31
Use underlying value when storage layout is not available

JulissaDantes:
Support contracts that have user defined value types, when there is not a storage layout. Truffle does not provide a storage layout so in order to support user defined value types we need to use the underlying type to compare between the original and the updated contract.

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
- - Use underlying type of user defined value types in the storage layout for layout comparison.
+ - Use underlying type of user defined value types in the storage layout for layout comparison. ([#682](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/682))
 
 ## 1.20.2 (2022-10-26)
 
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+ - Use underlying type of user defined value types in the storage layout for layout comparison.
+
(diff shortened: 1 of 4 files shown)


packages/core/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+ - Use underlying type of user defined value types in the storage layout for layout comparison.
```
ericglau, line 5:
```suggestion
 - Use underlying type of user defined value types in the storage layout for layout comparison. ([#682](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/682))
```

packages/core/src/storage/compare.ts
```diff
@@ -362,7 +362,15 @@ export class StorageLayoutComparator {
       }
 
       case 't_userDefinedValueType': {
-        if (original.item.numberOfBytes === undefined || updated.item.numberOfBytes === undefined) {
+        const underlyingMatch =
+          original.item.underlying &&
+          updated.item.underlying &&
```
ericglau, line 367: It would be clearer to specifically check that these are not `undefined`:
```suggestion
          original.item.underlying !== undefined &&
          updated.item.underlying !== undefined &&
```

packages/core/src/storage/compare.ts
```diff
@@ -362,7 +362,15 @@ export class StorageLayoutComparator {
...
+          original.item.underlying.id == updated.item.underlying.id;
```
ericglau, line 368: Let's prefer strict equals since the types should match in this case:
```suggestion
          original.item.underlying.id === updated.item.underlying.id;
```

Issue #263, opened 2020-12-21
Support multiple contracts with same name


frangio:
Our handling of validations does not support multiple contracts with the same name, which is a definite possibility. This is caused by our indexing of the validations object using the contract name only. We should switch to source path + contract name.

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/a5fa60a691846080fd7ba84e9a5d6601d304902c/packages/core/src/validate.ts#L82

PR #684, opened 2022-11-02
Support multiple contracts with same name

ericglau:
Validations were stored as a map with the contract names as the keys.  However, there could be multiple contracts with the same name in different Solidity files, which can lead to the wrong contract being validated.  Instead, use fully qualified contract names (e.g. `path/file.sol:ContractName`) as the keys instead.

Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/263

### DIFF
--- a/docs/modules/ROOT/pages/api-core.adoc
+++ b/docs/modules/ROOT/pages/api-core.adoc
@@ -18,6 +18,8 @@ This class represents the implementation for an upgradeable contract and gives a
 new UpgradeableContract(name, solcInput, solcOutput, opts?);
 ```
 
+`name` is the name of the implementation contract as either a fully qualified name or contract name. If multiple contracts have the same name, you must use the fully qualified name e.g., `contracts/Bar.sol:Bar`
+
 `opts` is an object with options as defined in xref:api-hardhat-upgrades.adoc#common-options[Common Options].
 
 TIP: In Hardhat, `solcInput` and `solcOutput` can be obtained from the Build Info file, which itself can be retrieved with `hre.artifacts.getBuildInfo`.
--- a/docs/modules/ROOT/pages/api-core.adoc
+++ b/docs/modules/ROOT/pages/api-core.adoc
@@ -18,7 +18,7 @@ This class represents the implementation for an upgradeable contract and gives a
 new UpgradeableContract(name, solcInput, solcOutput, opts?);
 ```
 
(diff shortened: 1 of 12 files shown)


docs/modules/ROOT/pages/api-core.adoc
```diff
@@ -18,6 +18,8 @@ This class represents the implementation for an upgradeable contract and gives a
 new UpgradeableContract(name, solcInput, solcOutput, opts?);
 ```
 
+`name` is the name of the implementation contract as either a fully qualified name or contract name. If multiple contracts have the same name, you must use the fully qualified name e.g., `contracts/Bar.sol:Bar`
```
frangio, line 21:
```suggestion
`name` is the name of the implementation contract as either a fully qualified name or contract name. If multiple contracts have the same name, you must use the fully qualified name e.g., `contracts/Bar.sol:Bar`.
```

packages/core/src/validate.test.ts
```diff
@@ -133,3 +133,17 @@ testValid('UsesExplicitSafeExternalLibraryNatspec', 'transparent', true);
 testValid('UsesExplicitUnsafeInternalLibraryNatspec', 'transparent', true);
 testValid('UsesExplicitUnsafeExternalLibraryNatspec', 'transparent', true);
 testValid('TransitiveLibraryIsUnsafe', 'transparent', false);
+
+testValid('contracts/test/ValidationsSameNameSafe.sol:SameName', 'transparent', true);
+testValid('contracts/test/ValidationsSameNameUnsafe.sol:SameName', 'transparent', false);
+
+test('ambiguous name', t => {
+  const getAmbiguousName = () => getContractVersion(t.context.validation, 'SameName');
+  const error = t.throws(getAmbiguousName);
```
frangio, line 142: I'd usually write:

```suggestion
  const error = t.throws(() => getContractVersion(t.context.validation, 'SameName'));
```

Do you have a preference?

packages/core/src/validate/query.ts
```diff
@@ -11,54 +11,85 @@ const upgradeToSignature = 'upgradeTo(address)';
...
-  const { version } = runData[contractName];
+  let version = undefined;
+  if (contractName.includes(':')) {
+    version = runData[contractName].version;
+  } else {
+    const foundNames = Object.keys(runData).filter(element => {
+      if (element.endsWith(`:${contractName}`)) {
+        return true;
+      }
+    });
```
frangio, line 40:
```suggestion
    const foundNames = Object.keys(runData).filter(element => element.endsWith(`:${contractName}`));
```

frangio (approved):
Very straightforward. Looks good, only cosmetic comments:

PR #688, opened 2022-11-22
Support retype from contract or interface to address

### LINKED ISSUE #687: Cannot retype from interface to address
Hi, I'm having some issues with the security checks of the plugin when upgrading some implementations. They have to be with refactor (renaming and retyping).

I have used `unsafeSkipStorageLayoutCheck: true` to test they don't actually break storage layout, and they seem to actually work.

One is the transition from this:
`mapping(uint8 => IStakeSubject) private _stakeSubjectHandlers;`

to this:
```
/// @custom:oz-renamed-from _stakeSubjectHandlers
/// @custom:oz-retyped-from mapping(uint8 => IStakeSubject)
mapping(uint8 => address) private _stakeSubjects;

```

The other one is the same described here:
https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/665

The errors:
```
contracts/components/utils/Routed.sol:16: Upgraded `_deprecated_router` to an incompatible type
  - Bad upgrade from contract IRouter to address

contracts/components/staking/stake_subjects/StakeSubjectGateway.sol:21: Upgraded `_stakeSubjects` to an incompatible type
  - In mapping(uint8 => address)
    - Bad upgrade from contract IStakeSubject to address
      at assertStorageUpgradeSafe (node_modules/@openzeppelin/upgrades-core/src/storage/index.ts:35:11)
      at validateImpl (node_modules/@openzeppelin/hardhat-upgrades/src/utils/validate-impl.ts:42:31)
      at deployProxyImpl (node_modules/@openzeppelin/hardhat-upgrades/src/utils/deploy-impl.ts:73:3)
      at Proxy.upgradeProxy (node_modules/@openzeppelin/hardhat-upgrades/src/upgrade-proxy.ts:25:32)
      at Context.<anonymous> (test/components/upgrades.test.js:299:35)


```

Versions:
```
"@openzeppelin/hardhat-upgrades@1.21.0", "@openzeppelin/hardhat-upgrades@^1.20.0":
  version "1.21.0"
  resolved "https://registry.yarnpkg.com/@openzeppelin/hardhat-upgrades/-/hardhat-upgrades-1.21.0.tgz#e90fb7d858093f35a300b3a5a2fd32bca6179dfc"
  integrity sha512-Kwl7IN0Hlhj4HluMTTl0DrtU90OI/Q6rG3sAyd2pv3fababe9EuZqs9DydOlkWM45JwTzC+eBzX3TgHsqI13eA==
  dependencies:
    "@openzeppelin/upgrades-core" "^1.20.0"
    chalk "^4.1.0"
    debug "^4.1.1"
    proper-lockfile "^4.1.1"

"@openzeppelin/upgrades-core@^1.20.0":
  version "1.20.1"
  resolved "https://registry.yarnpkg.com/@openzeppelin/upgrades-core/-/upgrades-core-1.20.1.tgz#6734770e50796b941736f5a77394f5b73d3978e6"
  integrity sha512-GXvqLkMNY1dGj9BAIXiqYjdj/qgpoeTed+pH2OL4UOqMlxIh8yIYdkLE9wOWiUVVr0rhUGqaoaJ9NeFLlVzBQQ==
  dependencies:
    cbor "^8.0.0"
    chalk "^4.1.0"
    compare-versions "^5.0.0"
    debug "^4.1.1"
    ethereumjs-util "^7.0.3"
    proper-lockfile "^4.1.1"
    solidity-ast "^0.4.15"
   ```

Thanks a lot.

### PR DESCRIPTION
**Pull request by ericglau**
Fixes #687 

This fixes two related issues:
- Consider contract or interface types to be identical to addresses, so that the user does not need `oz-retyped-from` when switching between these.
- The type label of a contract or interface is represented internally as `contract MyContract`, so `oz-retyped-from` does not work if the user just says to retype from `MyContract`.
  - This also occurs when used in mappings, where the internal representation is like `mapping(uint8 => contract MyContract)`.
  - To fix this, we will remove all "contract " substrings from type labels before checking whether a field is retyped.
    - This allows contract or interface types to be retyped to other types (in addition to addresses) when used with `oz-retyped-from`

### CODE REVIEW

**Comment by frangio** [comment]
This method can potentially remove substrings where it shouldn't, like if a contract or custom type was named `Xcontract` and it shows up followed by a space. We should match against `/\bcontract /`.

**Comment by frangio** [comment]
Can you try this with structs and enums as well? Pretty sure there are more types that are printed similar to contracts.

File: packages/core/src/storage/compare.ts [hunk 1]
```diff
@@ -410,6 +410,9 @@ function enumSize(memberCount: number): number {
 
 export function stripContractSubstrings(label?: string) {
   if (label !== undefined) {
-    return label.replace(/\bcontract /g, '');
+    return label
+      .replace(/\bcontract /g, '')
+      .replace(/\bstruct /g, '')
+      .replace(/\benum /g, '');
```
frangio, line 416: Clearer intention IMO:

```suggestion
    return label.replace(/\b(contract|struct|enum) /g, '');
```

frangio (approved):

PR #690, opened 2022-11-23
Normalize memory pointer during type comparison

### LINKED ISSUE #689: Incompatible type error when upgrading from mapping with strings
Reported in https://forum.openzeppelin.com/t/upgrade-plugin-reports-unsafe-upgrade-between-t-string-memory-and-t-string-memory-ptr/33281

In Solidity 0.5.16, have an implementation contract with:
```
mapping(string => address) a;
```
Attempt to upgrade to a contract with the same storage variables but compiled with Solidity 0.8.9:
```
mapping(string => address) a;
```
This causes an error:
```
Upgraded `a` to an incompatible type
  - In key of mapping(string => address)
    - Bad upgrade from undefined to string
```

Storage type id (as seen in manifest file) for 0.5.16 appears as: `t_string_memory`
Storage type id (as seen in manifest file) for 0.8.9 appears as: `t_string_memory_ptr`
These should be considered equivalent.

ericglau:
Fixes #689 

Normalize memory pointer during type comparison rather than when extracting the storage layout, since manifests may already have storage layouts with either representation `_memory_ptr` or `_memory`


frangio (approved):
LGTM

PR #695, opened 2022-12-05
Enable verify to verify beacons

0xmichalis:
Currently the verify task can only verify
a beacon if a proxy for the beacon is already
deployed. This is not always the case, a proxy
for a beacon may get deployed long after the
beacon and implementation contracts. This change
makes `verify` compatible with the results of
`deployBeacon`.


0xmichalis:
@ericglau would appreciate a review here, currently we need to use this code in a custom fork of the OZ plugin to verify new beacon deployments.

Issue #240, opened 2020-11-24
Only consider errors from the library functions in use


frangio:
Our current checks will look at an entire library and consider it either upgrade safe or not. Due to this lack of granularity, a new function we wanted to introduce in the `Address` utility library in OpenZeppelin Contracts 3.3 caused the entire library to be unusable with the upgrades plugins. (https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/230)

We should only consider errors from those functions that are in use.

This opens an entire can of worms because we obviously need to consider errors from functions used indirectly (by the functions that are used directly, and so on transitively). The logic to implement this probably shares a lot with #159 so I would consider fixing both of these issues at the same time.

kevupton:
Agree, it is a bit annoying to have a mix of upgrade safe and non-upgrade safe contracts in a project. is there any way for it to just search the contract being deployed? Instead of the entire library?

frangio:
@kevupton Sorry this is not trivial to fix which is why we haven't done it yet.

Na3aga:
Hi, guys. Is this still an issue? 

I have my upgradeable contract operating with **ERC20** contracts using the **SafeMath** library which includes the **Address** utility library. And it also causes my Contract to be considered as not safely deployed by your **hardhat-upgrades** tool. This is very weird because I don't try to make third-party **ERC20** contracts _upgradeable_.
Should I use **SafeMathUpgradeable** for interactions with non-upgradeable contracts ?????

frangio:
I'm not sure I understood the full picture but I think the answer to your question is yes. Can you share an example code snippet?

Mouzayan:
@frangio I am encountering the same error.
I have UUPS upgradeable contracts imported as dev dependencies, as well as one Transparent proxy upgradeable contract also imported as a dependency.
When I deploy them in my tests for my contract to interact with them, I get an error saying that the UUPS upgradeable contracts are 
```
not upgrade safe.
@openzeppelin/contracts/utils/Address.sol:191: Use of delegatecall is not allowed.
```
The upgradeable contracts are using AddressUpgradeable... so not sure why this issue is occurring.

Na3aga:
> I'm not sure I understood the full picture but I think the answer to your question is yes. Can you share an example code snippet?

I have solved it by ignoring the error and silencing the warning. 
The issue is, that my Upgradeable contract calls non-upgradeable ERC20 Contracts, and I have the line:
[...]

Issue #159, opened 2020-09-07
Validate free functions in Solidity 0.7.1


frangio:
Solidity 0.7.1 [introduces](https://solidity.ethereum.org/2020/09/02/solidity-0.7.1-release-announcement/) file-level free functions, which the plugins have to validate for upgrade safety. They behave just like internal library functions, and the implementation in the plugins should be the same too.

We have to get all referenced free functions, and then also run the appropriate validations in them.

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/608f133bca3500e8721c0867ed9a534e7353aa20/packages/core/src/validate.ts#L74

I think the only relevant validations are "delegate call errors".

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/608f133bca3500e8721c0867ed9a534e7353aa20/packages/core/src/validate.ts#L250-L253

We have to run these also in the free functions and then make a contract that references the function inherit its errors too.

~Note that `solidity-ast` doesn't support this new feature yet, so `findAll` will not find free functions in a file.~
`solidity-ast` supports free functions [since v0.4.4](https://github.com/OpenZeppelin/solidity-ast/releases/tag/v0.4.4).

julianmrodri:
Hi, did some research on this.  It is **not that straightforward** but can certainly be accommodated in the current process with a few changes:

The main issues are:
1. These functions show up in `solc-output` as isolated `FunctionDefinition`, so with the current process they are not even picked up for validation (as we only analyze `ContractDefinition` which included Libraries). 
2. The function call look exactly like normal function calls, so they are not easily identified as Library calls (which use `type (library)` and can be uniquely identified.* 

To overcome this issues one possible **approach** is the following:

* In `validate.ts` create an additional step to get all `freeFunctions` and populate them in the `Validation` object.  The only field that would be completed though are the `errors` as it will not have any of the other attributes:

```javascript
  for (const fnDef of findAll('FunctionDefinition', solcOutput.sources[source].ast)) {
      fromId[fnDef.id] = fnDef.name;

      if (fnDef.kind === 'freeFunction') {
        validation[fnDef.name] = {
          inherit: [],
          libraries: [],
          freeFunctions: [],
          linkReferences: [],
          errors: [...getDelegateCallErrors(fnDef, decodeSrc)],
          layout: {
            storage: [],
            types: {},
          },
        };
      }
    }
```
[...]

PR #702, opened 2022-12-15
Only consider errors from functions in use. Validate free functions.

ericglau:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/240
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/159

### DIFF
--- a/packages/core/src/validate/run.ts
+++ b/packages/core/src/validate/run.ts
@@ -1,6 +1,6 @@
 import { Node } from 'solidity-ast/node';
-import { isNodeType, findAll } from 'solidity-ast/utils';
-import type { ContractDefinition } from 'solidity-ast';
+import { isNodeType, findAll, ASTDereferencer } from 'solidity-ast/utils';
+import type { ContractDefinition, FunctionDefinition } from 'solidity-ast';
 
 import { SolcOutput, SolcBytecode } from '../solc-api';
 import { SrcDecoder } from '../src-decoder';
@@ -68,6 +68,15 @@ interface ValidationErrorOpcode extends ValidationErrorBase {
   kind: 'delegatecall' | 'selfdestruct';
 }
 
+interface OpcodePattern {
+  kind: 'delegatecall' | 'selfdestruct';
+  pattern: RegExp;
+}
+
+export function isOpcodeError(error: ValidationErrorBase) {
+  return error.kind === 'delegatecall' || error.kind === 'selfdestruct';
+}
+
 interface ValidationErrorUpgradeability extends ValidationErrorBase {
   kind: 'missing-public-upgradeto';
 }
@@ -84,8 +93,10 @@ function* execall(re: RegExp, text: string) {
   }
 }
 
-function getAllowed(node: Node): string[] {
+function getAllowed(node: Node, reachable: boolean): string[] {
   if ('documentation' in node) {
+    const tag = `oz-upgrades-unsafe-allow${reachable ? '-reachable' : ''}`;
+
     const doc = typeof node.documentation === 'string' ? node.documentation : node.documentation?.text ?? '';
 
     const result: string[] = [];
@@ -93,14 +104,14 @@ function getAllowed(node: Node): string[] {
       /^\s*(?:@(?<title>\w+)(?::(?<tag>[a-z][a-z-]*))? )?(?<args>(?:(?!^\s@\w+)[^])*)/m,
       doc,
     )) {
-      if (groups && groups.title === 'custom' && groups.tag === 'oz-upgrades-unsafe-allow') {
+      if (groups && groups.title === 'custom' && groups.tag === tag) {
         result.push(...groups.args.split(/\s+/));
       }
     }
 
     result.forEach(arg => {
       if (!(errorKinds as readonly string[]).includes(arg)) {
-        throw new Error(`NatSpec: oz-upgrades-unsafe-allow argument not recognized: ${arg}`);
+        throw new Error(`NatSpec: ${tag} argument not recognized: ${arg}`);
       }
     });
 
@@ -110,8 +121,13 @@ function getAllowed(node: Node): string[] {
   }
 }
 
+function skipCheckReachable(error: string, node: Node): boolean {
+  return getAllowed(node, true).includes(error);
+}
+
 function skipCheck(error: string, node: Node): boolean {
-  return getAllowed(node).includes(error);
+  // skip both allow and allow-reachable errors in the lexical scope
+  return [...getAllowed(node, false), ...getAllowed(node, true)].includes(error);
 }
 
 function getFullyQualifiedName(source: string, contractName: string) {
@@ -148,7 +164,8 @@ export function validate(solcOutput: SolcOutput, decodeSrc: SrcDecoder, solcVers
       };
     }
 
-    for (const contractDef of findAll('ContractDefinition', solcOutput.sources[source].ast)) {
+    const allContractDefs = findAll('ContractDefinition', solcOutput.sources[source].ast);
+    for (const contractDef of allContractDefs) {
       const key = getFullyQualifiedName(source, contractDef.name);
 
       fromId[contractDef.id] = key;
@@ -160,10 +177,12 @@ export function validate(solcOutput: SolcOutput, decodeSrc: SrcDecoder, solcVers
         inheritIds[key] = contractDef.linearizedBaseContracts.slice(1);
         libraryIds[key] = getReferencedLibraryIds(contractDef);
 
+        const opcodeErrors = [...getOpcodeErrors(contractDef, deref, decodeSrc)];
+
         validation[key].src = decodeSrc(contractDef);
         validation[key].errors = [
           ...getConstructorErrors(contractDef, decodeSrc),
-          ...getOpcodeErrors(contractDef, decodeSrc),
+          ...opcodeErrors,
           ...getStateVariableErrors(contractDef, decodeSrc),
           // TODO: add linked libraries support
           // https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/52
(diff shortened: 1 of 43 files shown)


packages/core/src/validate/run.ts
```diff
@@ -206,27 +225,159 @@ function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDe
...
+  decodeSrc: SrcDecoder,
+): Generator<ValidationErrorOpcode> {
+  yield* getContractOpcodeErrors(
+    contractOrFunctionDef,
+    deref,
+    decodeSrc,
+    {
+      kind: 'delegatecall',
+      pattern: /^t_function_baredelegatecall_/,
+    },
```
frangio, line 240: I wonder if this function would look better if we define these description objects in variables at the top level. Along with the default value for the visited set, the result could be:

```typescript
  yield* getContractOpcodeErrors(contractOrFunctionDef, deref, decodeSrc, opcodes.delegatecall, false);
  yield* getContractOpcodeErrors(contractOrFunctionDef, deref, decodeSrc, opcodes.selfdestruct, false);
```

packages/core/src/validate/run.ts
```diff
@@ -206,27 +225,159 @@ function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDe
...
+    contractOrFunctionDef,
+    node => skipCheckReachable(opcode.kind, node) || skipInternalFunctions(skipInternal, node),
+  )) {
     const fn = fnCall.expression;
-    if (fn.typeDescriptions.typeIdentifier?.match(/^t_function_selfdestruct_/)) {
-      yield {
-        kind: 'selfdestruct',
-        src: decodeSrc(fnCall),
-      };
+    const fnReference = (fn as any).referencedDeclaration;
```
frangio, line 334: The way to type this in TypeScript is to write `if ('referencedDeclaration' in fn) { ... }`.

In the body, `fn` will be narrowed to those types in the union that have the field.

This would be preferable so the type of `fnReference` is not `any`.

packages/core/src/validate/run.ts
```diff
@@ -206,27 +225,159 @@ function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDe
...
+    if (fnReference !== undefined && fnReference > 0) {
```
frangio, line 335: Leave a little comment saying "non-positive numbers refer to built-in functions" or something like this please.

packages/core/src/validate/run.ts
```diff
@@ -206,27 +225,159 @@ function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDe
...
+function* getContractOpcodeErrors(
+  contractDef: ContractDefinition,
+  deref: ASTDereferencer,
+  decodeSrc: SrcDecoder,
+  opcode: OpcodePattern,
+  skipInternal: boolean,
+  visitedNodeIds: Set<number>,
+): Generator<ValidationErrorOpcode> {
+  if (visitedNodeIds.has(contractDef.id)) {
+    return;
```
frangio, line 266: Let's add a comment explaining why the early return is correct. Something like: "We return early here, but the errors from this node were emitted when the node was marked as visited.

packages/core/src/validate/run.ts
```diff
@@ -206,27 +225,159 @@ function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDe
...
+      try {
+        const referencedNode = deref(['FunctionDefinition'], fnReference);
+        if (!visitedNodeIds.has(referencedNode.id)) {
+          visitedNodeIds.add(referencedNode.id);
+          yield* getFunctionOpcodeErrors(referencedNode, deref, decodeSrc, opcode, false, visitedNodeIds);
+        }
+      } catch (e: any) {
+        if (!e.message.includes('No node with id')) {
+          throw e;
+        }
```
frangio, line 345: Can we define a helper `tryDeref` instead?

packages/core/src/validate/run.ts
```diff
@@ -206,27 +225,159 @@ function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDe
...
+  contractDef: ContractDefinition,
+  deref: ASTDereferencer,
+  decodeSrc: SrcDecoder,
+  opcode: OpcodePattern,
+  visitedNodeIds: Set<number>,
+) {
+  if (!skipCheckReachable(opcode.kind, contractDef)) {
+    for (const base of contractDef.baseContracts) {
+      const referencedContract = deref('ContractDefinition', base.baseName.referencedDeclaration);
+      yield* getContractOpcodeErrors(referencedContract, deref, decodeSrc, opcode, true, visitedNodeIds);
```
frangio, line 361: I'm not sure that we're catching errors in parent constructors if they're internal.

Do we have a test for that? We should test the two different syntaxes for invoking a parent constructor:

```solidity
contract Parent {
    constructor(uint x) internal {
      selfdestruct(msg.sender);
    }
}
contract Child1 is Parent {
    constructor() Parent(0) {}
}
contract Child2 is Parent(0) {}
```

What if theif the parent constructor has no arguments, i.e. it's implicitly invoked?

packages/core/src/validate/run.ts
```diff
@@ -206,27 +225,159 @@ function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDe
...
-        kind: 'selfdestruct',
-        src: decodeSrc(fnCall),
-      };
+    const fnReference = (fn as any).referencedDeclaration;
+    if (fnReference !== undefined && fnReference > 0) {
+      try {
+        const referencedNode = deref(['FunctionDefinition'], fnReference);
+        if (!visitedNodeIds.has(referencedNode.id)) {
+          visitedNodeIds.add(referencedNode.id);
+          yield* getFunctionOpcodeErrors(referencedNode, deref, decodeSrc, opcode, false, visitedNodeIds);
```
frangio, line 340: Hm... I'm slightly concerned that we would visit a node with `skipInternal=true` _before_ we visit it with `skipInternal=false`, thus skipping over it the second time and potentially missing errors.

My intuition is that this is not happening but only becasue we run `getFunctionOpcodeErrors` before we run `getInheritedContractOpcodeErrors`. This seems fragile. I would appreciate if the correctness relied on a more explicit mechanism. For example, instead of a `Set` for visited nodes, we could have a `Map<number, 'internal' | 'all'>` where we explicitly mark visited and visibility.

packages/core/src/validate/run.ts
```diff
@@ -160,10 +177,12 @@ export function validate(solcOutput: SolcOutput, decodeSrc: SrcDecoder, solcVers
         inheritIds[key] = contractDef.linearizedBaseContracts.slice(1);
         libraryIds[key] = getReferencedLibraryIds(contractDef);
 
+        const opcodeErrors = [...getOpcodeErrors(contractDef, deref, decodeSrc)];
```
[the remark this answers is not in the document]
ericglau: Moved it back.

packages/core/src/validate/run.ts
ericglau: Defined `tryDerefFunction` in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/702/commits/d6c006ff44d4413d69e728bac93fcf7b4bdf47d2

packages/core/src/validate/run.ts
ericglau: Addressed in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/702/commits/6f47413d1db5a208abe2c75e926814338ed6a40e

packages/core/src/validate/run.ts
ericglau: Addressed this in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/702/commits/4aa9b4c457db7df6ddeb4ba70b1dbed5ad7c5b1e by always checking parent constructors even if they are 'internal' functions.  This is because child contracts always call parent constructors, either explicitly or implicitly.

ericglau:
There are a few additional issues that I've found:

1. Modifiers with unsafe opcodes in parent contracts are reported, but even if those modifiers are not used.  This is because `getDirectFunctionOpcodeErrors` finds the modifier's `FunctionCall` containing the unsafe opcode, although the `NodeType` is a `ModifierDefinition` rather than a `FunctionDefinition`.  Therefore `isInternalFunction` evaluates to `false` since this is not a `FunctionDefinition`, and the check is not skipped.  

Instead, the above should skip modifiers (which are internal), but `getReferencedFunctionOpcodeErrors` should also look for `ModifierInvocation` nodes and dereference those.

2. If a constructor calls a parent constructor with a modifier-like syntax like the below, `@custom:oz-upgrades-unsafe-allow-reachable` has no effect if that is defined on the constructor:
```
abstract contract Parent {
    constructor() {
        msg.sender.delegatecall('');
    }
}

contract Child is Parent {
    /// @custom:oz-upgrades-unsafe-allow-reachable delegatecall
    constructor() Parent() {}
}
```
This is because we are not dereferencing this constructor call, but are always reporting errors in inherited contract constructors.  Note that the constructor call is actually a `ModifierInvocation` that references the parent contract itself, not the parent constructor.

The workaround for users is to define `@custom:oz-upgrades-unsafe-allow-reachable` at the `contract` level instead.

3. Specifying `allow` and `allow-reachable` together like this:
```
abstract contract Parent {
    constructor() {
        msg.sender.delegatecall('');
    }
}

contract Child is Parent {
    /// @custom:oz-upgrades-unsafe-allow-reachable delegatecall
    constructor() Parent() {}
}
```
causes the following error:
```
  Error {
    message: 'NatSpec: oz-upgrades-unsafe-allow argument not recognized: ',
  }

  › Error: NatSpec: oz-upgrades-unsafe-allow argument not recognized:
  › <anonymous> (src/validate/run.ts:125:15)
  › Array.forEach (<anonymous>)
  › getAllowed (src/validate/run.ts:123:12)
  › skipCheck (src/validate/run.ts:141:10)
  › <anonymous> (src/validate/run.ts:225:74)
  › findAll (/Users/eric/git/openzeppelin-upgrades/node_modules/solidity-ast/utils/find-all.js:8:16)
  › findAll.next (<anonymous>)
  › findAll (/Users/eric/git/openzeppelin-upgrades/node_modules/solidity-ast/utils/find-all.js:21:18)
  › findAll.next (<anonymous>)
  › getConstructorErrors (src/validate/run.ts:225:14)
```

frangio (approved):
LGTM!

PR #706, opened 2022-12-31
Handle getLogs error for blockscout explorer

islishude:
https://github.com/blockscout/blockscout/blob/620178117fc008d0a926fb0c989864515f897487/apps/block_scout_web/lib/block_scout_web/controllers/api/rpc/logs_controller.ex#L6-L22

blockscout will return `No logs found` error message if no logs for the request.


frangio (approved):
Thanks @islishude! This will get released next week.

PR #722, opened 2023-01-23
Support rename / retype inside structs

Amxx:
This PR includes an example of rename that we need to support (possibly with a different natspec comment)

### DIFF
--- a/packages/core/contracts/test/RenamedRetyped.sol
+++ b/packages/core/contracts/test/RenamedRetyped.sol
@@ -62,3 +62,44 @@ contract LayoutChangeV2 {
     /// @custom:oz-retyped-from bool
     uint8 b;
 }
+
+
+contract RenameStructV1 {
+    struct History {
+        Checkpoint[] _checkpoints;
+    }
+
+    struct Checkpoint {
+        uint32 _blockNumber;
+        uint224 _value;
+    }
+
+    History history;
+}
+
+contract RenameStructV2a {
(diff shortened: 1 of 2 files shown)


frangio:
Don't we need retyping in addition to renaming? To convert the Timer struct into an integer type.

frangio:
@Amxx I fixed one of the tests you had written. There is already a built-in way to make that work, by using "retyped" on the state variable.

I also added another test case which we need for the Governor changes if we want to replace the timer structs with integers. This change also can be accepted by the plugin with the same annotation as the other case. However, that got me thinking that I'm not sure it's an actually acceptable change because I'm not sure about the alignment of the `uint32` value inside of the struct. Further testing makes me think that the plugin is not properly validating the retype annotation when it comes to structs: `InnerRetypeV2Bad` should clearly be considered a bad upgrade, but with the retype annotation it's accepted. (Retyped is not meant to override those deeper layout compatibility protections.)

frangio:
Ok I see what's happening. If there is a retype annotatoin we explicitly skip the type comparison where the inside of the struct would get compared. What we do make sure is that there are no "outer layout changes". @ericglau Should we also consider checking for "inner layout changes" when the retype refers to a struct?

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/1bca9c3ad6f6760642f0ff9b68da1085d27dc83f/packages/core/src/storage/compare.ts#L171-L173

PR #724, opened 2023-01-25
Use global caches for opcode validations

ericglau:
Improves compilation performance by using global caches for delegatecall and selfdestruct opcodes during validations. 

Fixes LIB-648

### DIFF
--- a/packages/core/src/validate/run.ts
+++ b/packages/core/src/validate/run.ts
@@ -302,9 +302,7 @@ function* getContractOpcodeErrors(
 function* getCached(key: number, scope: string, cache: Cache) {
   const cached = scope === 'main' ? cache.mainContractErrors.get(key) : cache.inheritedContractErrors.get(key);
   if (cached !== undefined) {
-    for (const r of cached) {
-      yield r;
-    }
+    yield* cached;
   } // else the node is currently being visited at a shallower level of recursion, so no need to report its errors at this level
 }
 
@@ -386,9 +384,7 @@ function* cacheAndYieldResult(key: number, scope: string, cache: Cache, result: 
   } else {
     cache.inheritedContractErrors.set(key, result);
   }
-  for (const r of result) {
-    yield r;
-  }
+  yield* result;
 }
 
 function tryDerefFunction(deref: ASTDereferencer, referencedDeclaration: number): FunctionDefinition | undefined {
--- a/packages/core/src/validate/run.ts
+++ b/packages/core/src/validate/run.ts
@@ -161,7 +161,8 @@ export function validate(solcOutput: SolcOutput, decodeSrc: SrcDecoder, solcVers
 
   const deref = astDereferencer(solcOutput);
 
-  const { delegateCallCache, selfDestructCache } = initCaches();
+  const delegateCallCache = initOpcodeCache();
+  const selfDestructCache = initOpcodeCache();
 
   for (const source in solcOutput.contracts) {
     for (const contractName in solcOutput.contracts[source]) {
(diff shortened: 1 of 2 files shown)


packages/core/src/validate/run.ts
```diff
@@ -241,13 +243,35 @@ function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDe
   }
 }
 
+function initCaches() {
```
frangio, line 246: I would simplify this to just create a single cache, perhaps renaming to `initOpcodeCache`, and `Cache` to `OpcodeCache` to be a little more specific.

frangio, line 246: I think I would just change this to a function that just creates and returns a single cache. Then we just call it once for each cache we want.

packages/core/src/validate/run.ts
```diff
@@ -241,13 +243,35 @@ function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDe
...
+  };
+  const selfDestructCache: Cache = {
+    mainContractErrors: new Map<number, ValidationErrorOpcode[]>(),
+    inheritedContractErrors: new Map<number, ValidationErrorOpcode[]>(),
+    visitedNodeIds: new Map<number, Scope>(),
+  };
+  return { delegateCallCache, selfDestructCache };
+}
+
+interface Cache {
```
frangio, line 260: I would consider renaming `Cache` to `OpcodeCache` to be a little more specific. Also `initCaches` to `initOpcodeCaches`, etc.

packages/core/src/validate/run.ts
```diff
@@ -241,13 +243,35 @@ function* getConstructorErrors(contractDef: ContractDefinition, decodeSrc: SrcDe
   }
 }
 
+function initCaches() {
+  const delegateCallCache: Cache = {
+    mainContractErrors: new Map<number, ValidationErrorOpcode[]>(),
+    inheritedContractErrors: new Map<number, ValidationErrorOpcode[]>(),
+    visitedNodeIds: new Map<number, Scope>(),
+  };
```
frangio, line 251: We should avoid type annotations if they can be inferred. Here, because we're assigning to a variable that is explicitly annotatated as `Cache`, the type of the maps will be inferred from that.

If you change this function to `function initOpcodeCache(): OpcodeCache` then the return statement will also be able to infer the type automatically.

```suggestion
  const delegateCallCache: Cache = {
    mainContractErrors: new Map(),
    inheritedContractErrors: new Map(),
    visitedNodeIds: new Map(),
  };
```

packages/core/src/validate/run.ts
```diff
@@ -261,24 +285,31 @@ function* getContractOpcodeErrors(
...
+}
 
-  yield* getFunctionOpcodeErrors(contractDef, deref, decodeSrc, opcode, scope, visitedNodeIds);
-  yield* getInheritedContractOpcodeErrors(contractDef, deref, decodeSrc, opcode, visitedNodeIds);
+function* getCached(key: number, scope: string, cache: Cache) {
+  const cached = scope === 'main' ? cache.mainContractErrors.get(key) : cache.inheritedContractErrors.get(key);
+  if (cached !== undefined) {
+    for (const r of cached) {
+      yield r;
+    }
```
frangio, line 307:
```suggestion
    yield* cached;
```

packages/core/src/validate/run.ts
```diff
@@ -337,15 +368,29 @@ function* getReferencedFunctionOpcodeErrors(
...
 
+function* cacheAndYieldResult(key: number, scope: string, cache: Cache, result: ValidationErrorOpcode[]) {
+  if (scope === 'main') {
+    cache.mainContractErrors.set(key, result);
+  } else {
+    cache.inheritedContractErrors.set(key, result);
+  }
+  for (const r of result) {
+    yield r;
+  }
```
frangio, line 391:
```suggestion
  yield* result;
```

packages/core/src/validate/run.ts
```diff
@@ -261,24 +285,31 @@ function* getContractOpcodeErrors(
...
+  if (wasVisited(contractDef.id, scope, cache.visitedNodeIds)) {
+    yield* getCached(contractDef.id, scope, cache);
   } else {
-    visitedNodeIds.set(contractDef.id, scope);
+    cache.visitedNodeIds.set(contractDef.id, scope);
+    const result = [
+      ...getFunctionOpcodeErrors(contractDef, deref, decodeSrc, opcode, scope, cache),
+      ...getInheritedContractOpcodeErrors(contractDef, deref, decodeSrc, opcode, cache),
+    ];
+    yield* cacheAndYieldResult(contractDef.id, scope, cache, result);
```
frangio, line 298: Now that we have separate caches for inherited and main, we may not need the `visitedNodeIds` set any more. I'll propose an alternative but let me know what you think is best.

We can just use the caches directly.

```typescript
    const cached = cache.[main or inherited].get(contractDef.id);
    if (cached !== undefined) {
      yield* cached;
    } else {
      const errors = [];
      cache.[main or inherited].set(contractDef.id, errors);
      errors.push(
        ...getFunctionOpcodeErrors(contractDef, deref, decodeSrc, opcode, scope, cache),
        ...getInheritedContractOpcodeErrors(contractDef, deref, decodeSrc, opcode, cache),
      );
      yield* errors;
    }
```

If there is a cache hit, we yield that. If there is a cache miss, we create an empty array and store it in the cache, then we get the errors and push them into the cached array. Any recursive cache hits will yield that array but before any errors are pushed into it.

I've implement this pattern a few times so I'm familiar with it but what do you think?

frangio (commented):
This looks good though I have a few suggestions.

Did we see perf improvement with this?

packages/core/src/validate/run.ts
ericglau: This makes sense since the empty array is a simpler equivalent to the code before, where it had a visitedNodeId that was set but did not have anything in the cached errors.

ericglau:
Time to run `validate` when compiling contracts for [this project](https://github.com/forta-network/forta-contracts):
Before this PR: 10.1s
After this PR: 6.3s (60% faster)

frangio (approved):
LGTM!

PR #726, opened 2023-01-26
Support Hardhat tests in --parallel mode, support forked networks

### LINKED ISSUE #465: Improve network file naming
Development networks (Harhdat, Ganache), are named `.openzeppelin/unknown-3117.json` (or similar). Other "unknown" networks for which we don't have a name get the similarly named files, with the corresponding chain id, even if they're production networks (see https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/229). This causes a situation where it's not clear which of these files should be gitignored or not.

We can do better for development networks because by inspecting the `web3_clientVersion` we know it's a dev network if it contains TestRPC, I think.

In that case we should place things in a dedicated directory, possibly even in an ephemeral directory (e.g. `/tmp`) and we can also go further and try to somehow identify different instances of the test network and use a different file for each, since sharing the same file is a situation that can cause issues (although there are measures in place so that it doesn't).

### LINKED ISSUE #645: Better support Hardhat tests in --parallel mode
In Parallel mode, each thread gets its own Hardhat Network instance, yet they all have the same chain id so they use the same manifest file. I don't think this will cause errors with the manifest file because due to being a development network we basically ignore the contents if they don't match expectations, but it will for sure cause unnecessary contention and lead to errors such as https://github.com/NomicFoundation/hardhat/discussions/3093. 

This should probably be fixed along with https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/465, by detecting different instances and using a different manifest file for each.

ericglau:
Fixes #645
Fixes #465
Fixes LIB-622

**Hardhat development networks**:

When a Hardhat network is detected which has `hardhat_metadata` with `instanceId` in its RPC:
- use an `openzeppelin-upgrades` folder under the OS temp directory as the development networks' manifest directory. This is NOT an ephemeral directory.
- use manifest file named with `hardhat-<chain_id>-<instance_id>.json` 
- use lock file named with `chain-<chain_id>-<instance_id>.lock`

This allows Hardhat tests running in parallel to use different manifest and lock files, since Hardhat assigns different instance ids to each Hardhat network instance that is run concurrently.

It also avoids writing manifests for Hardhat networks to the project's `.openzeppelin` directory, since these files should not need to be included into source control.  

**Forked networks in Hardhat**:

When using a [forked network](https://hardhat.org/hardhat-network/docs/guides/forking-other-networks), the manifest file will be first read from the real manifest file of the network that was forked, then any changes will be written to the development manifest location described above.  Once the development manifest exists, any further reads and writes will happen on that development manifest.  The real manifest file will not be modified.

For example, if you deployed a proxy on mainnet, then run a Hardhat node that forks mainnet, you can run tests against that Hardhat node to test upgrading your mainnet proxy address.


ericglau:
Note that this does not address #465.  Specifically, it does not use an ephemeral or `/tmp` directory, and does not delete the manifest file afterwards for the following reasons:
- We don't currently have a way to determine if the Hardhat network is a short-lived network that was spawned for a Hardhat test, or if the user manually ran `hardhat node` and is running a series of scripts against that node.  It would be wrong to delete the manifest automatically in the latter case. 
- Because of the reason above, the manifest should be persisted to either the usual `.openzeppelin` directory, or to somewhere in `/tmp` but without cleaning it up automatically afterwards.  I have a slight preference to placing it in `.openzeppelin` just so that it is easier to find if the user is running scripts against a `hardhat node` and wants to check the manifest file for debugging purposes.
  - Another option would be to have a subfolder in .openzeppelin for these files to avoid cluttering the main `.openzeppelin` folder, but that would require users to update their pattern in .gitignore since we [currently recommend](https://docs.openzeppelin.com/upgrades-plugins/1.x/network-files) ignoring just `.openzeppelin/unknown-*.json`

ericglau:
Given the risk of cluttering the `.openzeppelin` directory mentioned above, I think writing to `/tmp` should be included in this PR.  We could include debug log messages to help determine where the manifest file is being written to.

frangio:
Just to put this out there... we could consider a subdirectory `.openzeppelin/dev/`. But I think I prefer using `/tmp` because it removes the need to set up .gitignore for these files.

packages/core/src/manifest.ts
```diff
@@ -38,27 +40,82 @@ function defaultManifest(): ManifestData {
...
+const MANIFEST_DEFAULT_DIR = '.openzeppelin';
+const MANIFEST_TEMP_DIR = 'openzeppelin';
+
+async function getDevelopmentInstanceId(provider: EthereumProvider, chainId: number): Promise<string | undefined> {
+  let hardhatMetadata;
+
+  try {
+    hardhatMetadata = await getHardhatMetadata(provider);
+  } catch (e: unknown) {
+    if (e instanceof Error && (e.message.includes('Method') || e.message.includes('method'))) {
```
frangio, line 52: Hm, why don't we just skip to `return undefined` for all errors?

packages/core/src/manifest.ts
```diff
@@ -38,27 +40,82 @@ function defaultManifest(): ManifestData {
...
+
+async function makeAndGetTempDir() {
+  const tempDir = path.join(os.tmpdir(), MANIFEST_TEMP_DIR);
+  await fs.mkdir(tempDir, { recursive: true });
+  return tempDir;
+}
 
 export class Manifest {
   readonly chainId: number;
+  readonly devInstanceId: string | undefined;
```
frangio, line 85:
```suggestion
  readonly devInstanceId?: string;
```

packages/core/src/manifest.ts
```diff
@@ -38,27 +40,82 @@ function defaultManifest(): ManifestData {
...
+
+    return new Manifest(chainId, devInstanceId, dir);
   }
 
-  constructor(chainId: number) {
+  constructor(chainId: number, devInstanceId?: string, dir?: string) {
     this.chainId = chainId;
+    this.devInstanceId = devInstanceId;
+    this.manifestDir = dir ?? MANIFEST_DEFAULT_DIR;
+    debug('manifest dir:', this.manifestDir);
```
frangio, line 104:
```suggestion
    debug('manifest directory:', this.manifestDir);
```

packages/core/src/manifest.ts
```diff
@@ -38,27 +40,82 @@ function defaultManifest(): ManifestData {
...
+
+    const fallbackName = `unknown-${getSuffix(chainId, devInstanceId)}`;
```
frangio, line 106: I'd like to start using a name other than `unknown` for these files. One option would be to use `dev-...`, but if we know these are Hardhat we could go with `hardhat-...`.

packages/core/src/manifest.ts
```diff
@@ -38,27 +40,82 @@ function defaultManifest(): ManifestData {
...
+    this.fallbackFile = path.join(this.manifestDir, `${fallbackName}.json`);
+
+    const networkName = networkNames[this.chainId];
+    if (networkName !== undefined && devInstanceId !== undefined) {
+      throw new Error(
+        `Broken invariant: development instance id ${devInstanceId} not expected for named network ${networkName}`,
```
frangio, line 112: I don't think this is an invariant. Users can customize the chain id of their local Hardhat network and sometimes they do, in particular they make it match a known network. This may be used when forking that chain to test.

In those cases, we should use a temporary manifest file rather than writing to the real network manifest file.

Forking is important, and I would propose to tackle it separately, but this PR risks breaking current users' testing workflows.

Note that `hardhat_metadata` includes a `forkedNetwork` field. Maybe it's as simple as looking up the forked network manifest and copying it into the temporary manifest file as the starting point?

packages/core/src/manifest.ts
```diff
@@ -38,27 +40,82 @@ function defaultManifest(): ManifestData {
   };
 }
 
-const manifestDir = '.openzeppelin';
+const MANIFEST_DEFAULT_DIR = '.openzeppelin';
+const MANIFEST_TEMP_DIR = 'openzeppelin';
```
frangio, line 44: Something more specific might be safer:

```suggestion
const MANIFEST_TEMP_DIR = 'openzeppelin-upgrades';
```

packages/core/src/manifest.ts
ericglau: Handled forking as well.
If using a forked network, reads from the real manifest unless a dev manifest already exists, then writes to dev manifest location.
If using a dev instance but non forked network, only reads from the dev manifest (even if a real manifest for the same chain id exists).

packages/core/src/manifest.ts
```diff
@@ -58,58 +62,80 @@ async function getDevelopmentInstanceId(provider: EthereumProvider, chainId: num
...
   readonly file: string;
-  readonly manifestDir: string;
-  private readonly fallbackFile: string;
+  readonly fallbackFile: string;
+
+  private readonly devInstanceMetadata?: DevInstanceMetadata;
+  readonly devDir?: string;
+  readonly devFile?: string;
+
+  private readonly forked: boolean = false;
```
frangio, line 98:
```suggestion
  private readonly forked = false;
```

packages/core/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Support storage gaps named with `__gap_*`. ([#732](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/732))
+- Support Hardhat tests in --parallel mode when using Hardhat version 2.12.3 or later. ([#726](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/726))
```
frangio, line 6: Let's add a new entry mentioning better support for forking.

frangio:
There is a small problem with the way forking was implemented. In the case of dev networks, `this.file` may be read without locking it, what is locked instead is the "dev manifest" rather than the original manifest, but this isn't the way locking should work. The risk is that the file can be read corrupted if another process is writing to it at the same time. I say this is a "small" problem because in this particular scenario I don't think this is a real practical risk... but we've been dilligent about locking until now.

In my mind this would be fixed by a slightly different approach where a `Manifest` instance for a forked network has a `parent: Manifest` instance and we use `parent.read()` when the dev network file doesn't exist yet. You may have other ideas though.

packages/core/src/manifest.ts
ericglau: Changed it to `private readonly forked: boolean` and assigned the initial value in the constructor.  The `boolean` is needed to ensure a compile time error if we do not definitively set it in the constructor.

packages/core/src/manifest.test.ts
```diff
@@ -94,26 +95,293 @@ test.serial('rename manifest', async t => {
...
+  const instanceId = '0xccc';
+  const devInstanceMetadata = { networkName: 'hardhat', instanceId: instanceId, forkedNetwork: { chainId: forkedId } };
+
+  const realManifest = new Manifest(forkedId);
+  const devManifest = new Manifest(devId, devInstanceMetadata, os.tmpdir());
+
+  try {
+    await realManifest.lockedRun(async () => {
+      await devManifest.read();
+    });
```
ericglau, line 212: @frangio This test increases the overall `core` test time from 6 seconds to 11 seconds since it needs to wait for the lock timeout. Wondering if you have any alternative suggestions.

frangio: We can add a way to customize the lockfile parameters used by `Manifest`. If we pass `retries: 0` it shouldn't add extra delay I think.

packages/core/src/manifest.ts
```diff
@@ -38,27 +41,103 @@ function defaultManifest(): ManifestData {
...
+    }
+    return new Manifest(chainId);
   }
 
-  constructor(chainId: number) {
+  constructor(chainId: number, devInstanceMetadata?: DevInstanceMetadata, osTmpDir?: string) {
     this.chainId = chainId;
+    this.devInstanceMetadata = devInstanceMetadata;
+
+    let forkedChainId = undefined;
```
frangio, line 113: This variable is not needed outside the `if` statement as far as I can tell.

packages/core/src/manifest.ts
```diff
@@ -150,6 +229,9 @@ export class Manifest {
       return validateOrUpdateManifestVersion(data);
     } catch (e: any) {
       if (e.code === 'ENOENT') {
+        if (this.parent !== undefined) {
+          return await this.parent.read();
+        }
         return defaultManifest();
```
frangio, line 235: This is subjective, I personally prefer like this:

```suggestion
        } else {
          return defaultManifest();
        }
```

packages/core/src/manifest.ts
```diff
@@ -38,27 +41,103 @@ function defaultManifest(): ManifestData {
...
   private locked = false;
 
   static async forNetwork(provider: EthereumProvider): Promise<Manifest> {
-    return new Manifest(await getChainId(provider));
+    const chainId = await getChainId(provider);
+    const devInstanceMetadata = await getDevInstanceMetadata(provider, chainId);
+    if (devInstanceMetadata !== undefined) {
+      return new Manifest(chainId, devInstanceMetadata, os.tmpdir());
+    }
+    return new Manifest(chainId);
```
frangio, line 106:
```suggestion
    } else {
      return new Manifest(chainId);
    }
```

packages/core/src/manifest.ts
```diff
@@ -180,7 +262,10 @@ export class Manifest {
   }
 
   private async lock() {
-    const lockfileName = path.join(manifestDir, `chain-${this.chainId}`);
+    const lockfileName = path.join(
+      this.devDir ?? MANIFEST_DEFAULT_DIR,
```
frangio, line 266: Might be better to keep a non-optional `dir` variable in both cases so we use it consistently here and in the constructor.

```suggestion
      this.dir,
```

frangio (approved):
Minor comments above. Otherwise this is good to go.

packages/core/src/manifest.ts
ericglau: Changed it save the chain id suffix (including instance id if applicable) as an instance variable, since that is needed to determine the lock file name.

Issue #723, opened 2023-01-24
Compilation warnings in test contracts


ericglau:
From this repository's `packages/core` directory, run `yarn`.  The following compilation warnings occur:
```
$ tsc -b && yarn compile:contracts
$ hardhat compile && node scripts/copy-build-info.js
Warning: Return value of low-level calls not used.
 --> contracts/test/ignore-errors/Constructors.sol:7:9:
  |
7 |         msg.sender.delegatecall(data);
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


Warning: Return value of low-level calls not used.
  --> contracts/test/ignore-errors/Constructors.sol:25:9:
   |
25 |         msg.sender.delegatecall('');
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^


Warning: Return value of low-level calls not used.
  --> contracts/test/ignore-errors/Constructors.sol:46:9:
   |
46 |         msg.sender.delegatecall('');
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^


Warning: Return value of low-level calls not used.
  --> contracts/test/ignore-errors/Constructors.sol:79:9:
   |
79 |         msg.sender.delegatecall('');
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^


Warning: Return value of low-level calls not used.
 --> contracts/test/ignore-errors/Modifiers.sol:7:9:
  |
7 |         msg.sender.delegatecall(data);
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


Compiled 79 Solidity files successfully
```

These contracts are used for tests only, but either find a way to ignore these specific warnings, or update the test cases to not cause these.

Ensure that the tests still pass when running `yarn test`.

frangio:
See my Hardhat plugin: [hardhat-ignore-warnings](https://github.com/frangio/hardhat-ignore-warnings)

sujantkumarkv:
Hey, these are my findings:

The warnings are related to `delegatecall`, it's used to execute code from another contract, and returns the value that is returned by the called contract. However, the return value here is not being used, hence causing the warnings.

To fix this, we can either store the return value of `delegatecall` in a variable and use it somehow or we can use the "call" function instead which allows for code execution from another contract, but it does not return a value.

Is a return value something that's necessary for the tests? i'll proceed based on your inputs.

sujantkumarkv:
[...]

PR #730, opened 2023-01-28
Removes compilation warnings in test contracts

sujantkumarkv:
Fixes [#723](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/723): 
**"Compilation warnings in test contracts"** 

- The `delegatecall` function which executes code from another contract, gives us value returned by the called contract but the code **wasn't using it**, thus causing the warnings; `delegatecall` was returning a tuple of `(bool, bytes)`. 

- To fix this, the data was received in a tuple and having a `require` condition as shown; 
`(bool result, )= msg.sender.delegatecall(data);`
`require(result == true, "delegatecall failed");`

- The tuple above was earlier `(bool result, bytes memory resultData)` but `resultData` isn't necessary for usage; it also caused another warning of it being **unused**, therefore skipping it was the best solution here.

PS:
- A screenshot showing **_no warnings_** is attached; it also passed all tests with `yarn test`

Screenshot 2023-01-28 at 7 18 14 PM

### DIFF
--- a/packages/core/contracts/test/Validations.sol
+++ b/packages/core/contracts/test/Validations.sol
@@ -48,8 +48,7 @@ contract HasSelfDestruct {
 
 contract HasDelegateCall {
   function d() public {
-    (bool s, ) = msg.sender.delegatecall("");
-    //s;
+    (bool s,) = msg.sender.delegatecall("");
     require(s, "delegatecall failed");
   }
 }
--- a/packages/core/contracts/test/Validations.sol
+++ b/packages/core/contracts/test/Validations.sol
@@ -49,7 +49,8 @@ contract HasSelfDestruct {
 contract HasDelegateCall {
   function d() public {
     (bool s, ) = msg.sender.delegatecall("");
-    s;
+    //s;
+    require(s, "delegatecall failed");
   }
 }
 
--- a/packages/core/contracts/test/Validations.sol
+++ b/packages/core/contracts/test/Validations.sol
@@ -48,8 +48,8 @@ contract HasSelfDestruct {
 
 contract HasDelegateCall {
   function d() public {
-    (bool s,) = msg.sender.delegatecall("");
-    require(s, "delegatecall failed");
+    (bool s, ) = msg.sender.delegatecall("");
+    s;
   }
 }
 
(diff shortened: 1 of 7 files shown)


packages/core/contracts/test/ignore-errors/Constructors.sol
```diff
@@ -4,7 +4,8 @@ pragma solidity ^0.8.9;
 abstract contract UnsafeParent {
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor(bytes memory data) {
-        msg.sender.delegatecall(data);
+       (bool result,)= msg.sender.delegatecall(data);
+        require(result == true, "delegatecall failed");
```
frangio, line 8: Please apply this suggestion in the other contracts as well.

```suggestion
       (bool result,) = msg.sender.delegatecall(data);
        require(result, "delegatecall failed");
```

sujantkumarkv: Sure. 

- Some other files in `packages/core/` were resolving this by simply writing `s;` of returned boolean value `s`. I have added similar logic in those files too for code consistency which also gives a helpful message `"delegatecall failed"` when the function call fails.

The commit changes, [here.](https://github.com/OpenZeppelin/openzeppelin-upgrades/commit/7ffa8d86cd6416019189decb73b1a9ad59d0f350
)

PS: 
Screenshot 2023-01-29 at 10 47 45 AM

packages/core/contracts/test/Validations.sol
```diff
@@ -49,7 +49,8 @@ contract HasSelfDestruct {
 contract HasDelegateCall {
   function d() public {
     (bool s, ) = msg.sender.delegatecall("");
-    s;
+    //s;
```
frangio, line 52: Left over comment.

frangio:
My suggestion above hasn't been applied in all contracts. Can you review that? Please note the whitespace.

sujantkumarkv:
> My suggestion above hasn't been applied in all contracts. Can you review that? Please note the whitespace.

Your suggestion of only the order of whitespaces right? I'll do so. Is there something else I'm missing?

ericglau (approved):
I've changed it to follow the existing examples of
```
    (bool s, ) = msg.sender.delegatecall("");
    s;
```
There is no need to report delegatecall failures since these test contracts are intended for testing upgrade safety validations and are not intended to be run.

Thanks for looking into this issue!

PR #731, opened 2023-01-31
Improve detection of storage gap usage

### LINKED ISSUE #703: Cannot consume entire gap and add new gap at the same time
Consider the following storage layouts for V1, V2, V3:

```
pragma solidity ^0.8.4;

import "@openzeppelin/contracts-upgradeable/token/ERC721/ERC721Upgradeable.sol";

contract V1 is ERC721Upgradeable {
    uint256[50] private __gap;

    // functions
}
```

```
pragma solidity ^0.8.4;

import "@openzeppelin/contracts-upgradeable/token/ERC721/ERC721Upgradeable.sol";
import "@openzeppelin/contracts-upgradeable/token/ERC721/extensions/ERC721URIStorageUpgradeable.sol";

contract V2 is ERC721Upgradeable, ERC721URIStorageUpgradeable {

    // functions
}
```

```
pragma solidity ^0.8.4;

import "@openzeppelin/contracts-upgradeable/token/ERC721/ERC721Upgradeable.sol";
import "@openzeppelin/contracts-upgradeable/token/ERC721/extensions/ERC721URIStorageUpgradeable.sol";

contract V3 is ERC721Upgradeable, ERC721URIStorageUpgradeable {
    uint256 abc;
    uint256[49] private __gap;

    // functions
}
```

This sequence should be allowed because V1 has a 50 slot gap, V2 adds ERC721URIStorageUpgradeable in the inheritance list which consumes the 50 slot gap, and V3 adds an additional variable and a new gap at the end of the storage layout.

Upgrading from V1 to V2 to V3 works.

However, if we upgrade directly from V1 to V3, Hardhat Upgrades gives the following error:
```

contracts/V3.sol:9: Upgraded `__gap` to an incompatible type
  - Bad storage gap resize from 50 to 49
    Size decrease must match with corresponding variable inserts
  > Set __gap array to size -1
      at assertStorageUpgradeSafe (node_modules/@openzeppelin/upgrades-core/src/storage/index.ts:35:11)
      at Proxy.validateUpgrade (node_modules/@openzeppelin/hardhat-upgrades/src/validate-upgrade.ts:44:33)
``` 

Similarly, if V3 had gap size 50 and no `abc`, Hardhat Upgrades gives the following error, which is correct but could be more flexible to treat the original gap as consumed and allow the new gap:
```

contracts/V3.sol:9: Layout changed for `__gap` (uint256[50] -> uint256[50])
  - Slot changed from 151 to 201
  > Set __gap array to size 0
      at assertStorageUpgradeSafe (node_modules/@openzeppelin/upgrades-core/src/storage/index.ts:35:11)
      at Proxy.validateUpgrade (node_modules/@openzeppelin/hardhat-upgrades/src/validate-upgrade.ts:44:33)
```

ericglau:
Fixes #703 
Fixes LIB-657

Support consuming a gap (partial or entire gap) and adding new gap at the same time by adjusting the costs for the levenshtein comparison.


frangio (commented):
This looks good but I wish we had a principled way of explaining why we're setting the costs to this.

frangio (approved):

ericglau:
> This looks good but I wish we had a principled way of explaining why we're setting the costs to this.

Shrink gap is considered to be a safe operation, so if there is a change that could be considered a layout change, type change, or shrink gap, then we want to prioritize shrink gap (by giving it the lowest cost).

PR #732, opened 2023-01-31
Support storage gaps named with `__gap_*`

### LINKED ISSUE #657: Support a prefix for gaps instead of only `__gap`
#564 was merged recently which looks for `__gap` variables and does special handling for them. It would be very useful to match a `__gap_*` _prefix_ as well. This allows the use of multiple gaps within a single contract as well as give them more meaningful names like `__gap_purpose`. As an example, I like to divide my longer contracts into somewhat independent "sections" based on purpose and give each section its own gap so it can evolve independently as needed.

ericglau:
Fixes #657 
Fixes LIB-658

Support storage gaps named as either `__gap` or `__gap_*`


frangio (approved):
Looks good!

PR #737, opened 2023-02-03
Use proxy transactionHash for implementation contract abstraction in Truffle

llwslc:
The implementation contract transactionHash is not set to the proxy contract transactionHash in the build/contracts/\<ContractName\>.json file.

build:
```
"networks": {
    "5777": {
      "events": {},
      "links": {},
      "address": "0xd6Fd7B733e189109Ba0874B8c886Cc509a1b2C85",
      "transactionHash": "0xef8418e2bfea6f4175e86a707154ce63508d458c363074f3161fc92bdf4cd0ec"
    }
  },
```

.openzeppelin:
```
  "proxies": [
    {
      "address": "0xd6Fd7B733e189109Ba0874B8c886Cc509a1b2C85",
      "txHash": "0xea38585fa68592eec02faec61331e91d101be13bcdd54010bb8016a41d3ecde4",
      "kind": "transparent"
    }
  ],
  "impls": {
    "851eb7f66375439b9eb97824f366c9260c3472be98d739e29eccd0f291c5aaf9": {
      "address": "0xBb2100DeA2AfE3fc661e625E6F6673ca36bF0198",
      "txHash": "0xef8418e2bfea6f4175e86a707154ce63508d458c363074f3161fc92bdf4cd0ec",
      ...
    }
  }
```

### DIFF
--- a/packages/plugin-truffle/CHANGELOG.md
+++ b/packages/plugin-truffle/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Use proxy transaction hash for implementation contract abstraction. ([#737](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/737))
+
 ## 1.17.0 (2022-09-27)
 
 - Include solc version in storage layouts in manifest files. ([#662](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/662))
--- a/packages/plugin-truffle/src/deploy-beacon-proxy.ts
+++ b/packages/plugin-truffle/src/deploy-beacon-proxy.ts
@@ -82,6 +82,7 @@ export async function deployBeaconProxy(
   await manifest.addProxy(proxyDeployment);
 
   attachTo.address = proxyDeployment.address;
+  attachTo.transactionHash = proxyDeployment.txHash;
   const contract = new attachTo(proxyDeployment.address);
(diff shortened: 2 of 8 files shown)


ericglau:
@llwslc Thank you for this PR.  I've pushed a similar change here for beacon proxies as well.

Can you please provide the following info to help us better understand your use case?
1. What do you use the transaction hash for?
2. If you upgrade a proxy to an implementation contract V2, what do you expect to see in `build/contracts/V2.json`'s `address` and `transactionHash`?

llwslc:
1. When I looked at the `transactionHash` in `v1.json` in the block explorer, I found that the address  in `v1.json` was not the address shown in the page. I then reran 'truffle migrate' and found that the address  in `v1.json` had changed, but the `transactionHash` had not changed. This doesn't make sense.
2. I just used the `upgradeProxy` function, and I think the `transactionHash` in `v2.json` should also use the `transactionHash` of the proxy contract. `Address` and `transactionHash` should keep matching. `Address` is the proxy contract address, and `transactionHash` is the impl contract transactionHash, which has some ambiguity.

These are my ideas, thanks.

ericglau:
Thanks for the feedback.  I've made an additional change so that when using `upgradeProxy`, the `v2.json` will have the `transactionHash` of the transaction that upgraded the proxy.

ericglau (approved):

PR #744, opened 2023-02-14
add Foundry's "anvil" as a "development network"

AllTradesz:
And so avoid spurious "clashes" when upgrading proxy implementation contracts. Pull less hair.  👍 
Tested. Works now so much better.  😃


ericglau (approved):
Thanks for the PR!

PR #751, opened 2023-02-24
Remove test contracts from source code verification

### LINKED ISSUE #750: Remove test contracts from solc input
The Hardhat compilation artifact for the proxies, particularly the `artifacts/build-info.json` file that gets included with the `upgrades-core` package, contains source code for [test contracts](https://github.com/OpenZeppelin/openzeppelin-upgrades/tree/master/packages/core/contracts/test) as well.

This file contains the solc input that is used to verify the proxy when running the `verify` task.

The result is that these tests files would be included in the Etherscan verified source code for the proxy, when in fact these test files are not actually being used by the proxy.  This is misleading, especially since these test files have unsafe code that are used as part of test cases.

We should either change the compilation so that Hardhat does not include these tests in the same `build-info` file as the proxies, or remove the test contracts from the `build-info` file that gets published with the npm package.

ericglau:
Fixes #750

When compiling contracts in the `core` package, marks any contracts that are not part of `@openzeppelin/contracts/proxy` to be compiled separately.

Note that dependencies of `@openzeppelin/contracts/proxy` would still be compiled along with `@openzeppelin/contracts/proxy`

This results in the `build-info` file used for source code verification to only have the proxy contracts and their dependencies.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,7 @@ module.exports = {
   extends: ['eslint:recommended', 'plugin:prettier/recommended'],
   env: {
     node: true,
-    es6: true,
+    es2022: true,
   },
   plugins: ['unicorn'],
   rules: {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,6 +6,7 @@ module.exports = {
   extends: ['eslint:recommended', 'plugin:prettier/recommended'],
   env: {
     node: true,
+    es6: true,
   },
   plugins: ['unicorn'],
   rules: {
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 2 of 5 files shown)


ericglau:
@frangio This will omit the following files from the `build-info` file which is used for proxy verification:
```
      "@openzeppelin/contracts-upgradeable/proxy/beacon/IBeaconUpgradeable.sol"
      "@openzeppelin/contracts-upgradeable/proxy/ERC1967/ERC1967UpgradeUpgradeable.sol"
      "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol"
      "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol"
      "@openzeppelin/contracts-upgradeable/utils/AddressUpgradeable.sol"
      "@openzeppelin/contracts-upgradeable/utils/StorageSlotUpgradeable.sol"
      "contracts/import.sol"
      "contracts/test/*"
```

The `@openzeppelin/contracts-upgradeable` files are not used by the proxies themselves, but are only referenced by some test cases.

The `contracts/import.sol` does not look relevant for proxy verification either.

frangio (approved):

PR #759, opened 2023-03-14
Update proxy fallback solidity code snippet for solc 0.8.0

CJ42:
In the following page for the OpenZeppelin upgrade docs --> https://docs.openzeppelin.com/upgrades-plugins/1.x/proxies

The Solidity code snippet for the proxy forwarding (delegatecall in fallback function) does not work with the new solc version 0.8.x

image

The assembly opcodes now use the syntax with parentheses like `calldatasize()` for instance.

I am making this PR to update the docs.
Only part where I am not sure is for retrieving the address of the implementation contract from the contract state.

I use in assembly `sload(_impl.slot)` but maybe this is probably not robust because we don't know what else in the storage slot packed alongside the `_impl` address.

### DIFF
--- a/docs/modules/ROOT/pages/proxies.adoc
+++ b/docs/modules/ROOT/pages/proxies.adoc
@@ -29,9 +29,9 @@ The most immediate problem that proxies need to solve is how the proxy exposes t
 
 [source,solidity]
 ----
-// this code is for "illustration" purposes. To implement a similar functionality,
-// you should use the `Proxy` contract from the `@openzeppelin/contracts` library.
-// https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/Proxy.sol
+// This code is for "illustration" purposes. To implement this functionality in production it
+// is recommended to use the `Proxy` contract from the `@openzeppelin/contracts` library.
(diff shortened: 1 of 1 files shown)


CJ42:
An alternative would be to have the full code snippet of the contract as follow:

```solidity
contract Proxy {

    address private _impl;

    constructor(address implementation_) {
        _impl = implementation_;
    }

    fallback() external {
        assembly {
            let ptr := mload(0x40)

            // (1) copy incoming call data
            calldatacopy(ptr, 0, calldatasize())

            // (2) forward call to logic contract
            let result := delegatecall(gas(), sload(_impl.slot), ptr, 4, 0, 0)
            let size := returndatasize()

            // (3) retrieve return data
            returndatacopy(ptr, 0, size)

            // (4) forward return data back to caller
            switch result
            case 0 { revert(ptr, size) }
            default { return(ptr, size) }
        }
    }
}
```

frangio:
Thanks @CJ42! We should be using the code from OpenZeppelin Contracts, modified with the comments we have here.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.8.2/contracts/proxy/Proxy.sol#L22-L45C6

We should probably also add a note that this code is for "illustration" purposes and people should use the library contracts instead.

Could you make these changes in this PR?

PR #763, opened 2023-03-23
Enable using OpenZeppelin Platform for deployments

ericglau:
Remaining:
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/core/scripts/copy-build-info.js
+++ b/packages/core/scripts/copy-build-info.js
@@ -47,10 +47,48 @@ assert(
     readJSON('artifacts/@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol/ProxyAdmin.dbg.json').buildInfo,
 );
 
-const buildInfo = readJSON(jsonRelativePath);
-const reducedInfo = { solcLongVersion: buildInfo.solcLongVersion, input: buildInfo.input };
+let buildInfo = readJSON(jsonRelativePath);
 
-const sources = reducedInfo.input.sources;
+// Keep only relevant sections of output.contracts
+const contractFiles = buildInfo.output.contracts;
+for (const contractFile in contractFiles) {
+  const contractNames = contractFiles[contractFile];
+  for (const contractName in contractNames) {
+    contractNames[contractName] = {
+      abi: contractNames[contractName].abi,
+      evm: contractNames[contractName].evm,
+      metadata: contractNames[contractName].metadata, // metadata is needed to determine the license type
+    };
+  }
+}
+
+// Adjust correponding output selection
+const input = buildInfo.input;
+
+const origSelection = input.settings.outputSelection['*']['*'];
+assert(origSelection.includes('abi'));
+assert(origSelection.includes('evm.bytecode'));
+assert(origSelection.includes('evm.deployedBytecode'));
+assert(origSelection.includes('evm.methodIdentifiers'));
+assert(origSelection.includes('metadata'));
+
+input.settings.outputSelection['*'] = {
+  '*': ['abi', 'evm.bytecode', 'evm.deployedBytecode', 'evm.methodIdentifiers', 'metadata'],
+};
+
+// Keep only relevant sections of build info
+buildInfo = {
+  _format: buildInfo._format,
+  id: buildInfo.id,
+  solcVersion: buildInfo.solcVersion,
+  solcLongVersion: buildInfo.solcLongVersion,
+  input: input,
+  output: {
+    contracts: contractFiles,
+  },
+};
+
+const sources = input.sources;
 
 // Assert that all deployable proxy artifacts exist in ERC1967's build-info file
 assert(hasProperty(sources, '@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol'));
@@ -62,4 +100,4 @@ assert(hasProperty(sources, '@openzeppelin/contracts/proxy/transparent/ProxyAdmi
 // Assert that the build-info file does NOT contain test contracts
 assert(!hasPropertyStartsWith(sources, 'contracts/test'));
 
-writeJSON('artifacts/build-info.json', reducedInfo);
+writeJSON('artifacts/build-info.json', buildInfo);
--- a/packages/core/src/deployment.ts
+++ b/packages/core/src/deployment.ts
@@ -89,7 +89,7 @@ async function validateCached<T extends Deployment>(
 ): Promise<T | undefined> {
   if (cached !== undefined) {
     try {
-      await validateStoredDeployment(cached, provider, type, opts, merge, getDeploymentResponse);
+      return await validateStoredDeployment(cached, provider, type, opts, merge, getDeploymentResponse);
     } catch (e) {
       if (e instanceof InvalidDeployment && (await isDevelopmentNetwork(provider))) {
         debug('ignoring invalid deployment in development network', e.deployment.address);
@@ -101,8 +101,9 @@ async function validateCached<T extends Deployment>(
         throw e;
       }
     }
+  } else {
+    return undefined;
   }
-  return cached;
 }
 
 async function validateStoredDeployment<T extends Deployment & DeploymentId>(
@@ -112,8 +113,10 @@ async function validateStoredDeployment<T extends Deployment & DeploymentId>(
   opts?: DeployOpts,
   merge?: boolean,
   getDeploymentResponse?: (deploymentId: string, catchIfNotFound: boolean) => Promise<DeploymentResponse | undefined>,
-) {
+): Promise<T> {
   const { txHash, deploymentId } = stored;
+  let deployment = stored;
+
   if (txHash !== undefined) {
     // If there is a deployment with txHash stored, we look its transaction up. If the
     // transaction is found, the deployment is reused.
@@ -127,6 +130,10 @@ async function validateStoredDeployment<T extends Deployment & DeploymentId>(
       const response = await getDeploymentResponse(deploymentId, true);
       if (response !== undefined) {
         foundDeployment = deploymentId;
+
+        // update the stored tx hash
+        deployment = { ...stored, txHash: response.txHash };
+        debug('updating previous deployment to tx hash ', response.txHash);
       }
     }
 
@@ -134,20 +141,21 @@ async function validateStoredDeployment<T extends Deployment & DeploymentId>(
       debug('resuming previous deployment', foundDeployment);
       if (merge) {
         // If merging, wait for the existing deployment to be mined
-        waitAndValidateDeployment(provider, stored, type, opts, getDeploymentResponse);
+        waitAndValidateDeployment(provider, deployment, type, opts, getDeploymentResponse);
       }
     } else {
       // If the transaction is not found we throw an error, except if we're in
       // a development network then we simply silently redeploy.
       // This error should be caught by the caller to determine if we're in a dev network.
-      throw new InvalidDeployment(stored);
+      throw new InvalidDeployment(deployment);
     }
   } else {
-    const existingBytecode = await getCode(provider, stored.address);
+    const existingBytecode = await getCode(provider, deployment.address);
     if (isEmpty(existingBytecode)) {
-      throw new InvalidDeployment(stored);
+      throw new InvalidDeployment(deployment);
     }
   }
+  return deployment;
 }
 
 export interface DeploymentResponse {
--- a/packages/core/src/deployment.ts
+++ b/packages/core/src/deployment.ts
@@ -17,12 +17,9 @@ import { deleteDeployment, ManifestField } from './impl-store';
 
 const sleep = promisify(setTimeout);
 
-export interface BaseDeployment {
+export interface Deployment {
   address: string;
   txHash?: string;
-}
-
-export interface Deployment extends BaseDeployment {
   deploymentId?: string;
 }
 
--- a/packages/core/src/deployment.ts
+++ b/packages/core/src/deployment.ts
@@ -17,9 +17,12 @@ import { deleteDeployment, ManifestField } from './impl-store';
 
 const sleep = promisify(setTimeout);
 
-export interface Deployment {
+export interface BaseDeployment {
   address: string;
   txHash?: string;
+}
+
+export interface Deployment extends BaseDeployment {
   deploymentId?: string;
 }
 
--- a/packages/core/src/deployment.ts
+++ b/packages/core/src/deployment.ts
@@ -20,6 +20,9 @@ const sleep = promisify(setTimeout);
 export interface Deployment {
   address: string;
   txHash?: string;
+}
+
+export interface DeploymentId {
   deploymentId?: string;
 }
 
@@ -139,7 +142,7 @@ export interface DeploymentResponse {
 
 export async function waitAndValidateDeployment(
   provider: EthereumProvider,
-  deployment: Deployment,
+  deployment: Deployment & DeploymentId,
   type?: string,
   opts?: DeployOpts,
   getDeploymentResponse?: (deploymentId: string) => Promise<DeploymentResponse>,
@@ -173,7 +176,7 @@ export async function waitAndValidateDeployment(
         break;
       } else if (status === 'failed') {
         debug(`tx hash ${response.txHash} was reverted for deployment id ${deploymentId}`);
-        throw new InvalidDeployment({ address, txHash: response.txHash, deploymentId });
+        throw new InvalidDeployment({ address, txHash: response.txHash });
       } else if (status === 'submitted') {
         debug('waiting for deployment id to be completed', deploymentId);
         await sleep(pollInterval);
@@ -230,7 +233,7 @@ export async function waitAndValidateDeployment(
 }
 
 export class TransactionMinedTimeout extends UpgradesError {
-  constructor(readonly deployment: Deployment, type?: string, configurableTimeout?: boolean) {
+  constructor(readonly deployment: Deployment & DeploymentId, type?: string, configurableTimeout?: boolean) {
     super(
       `Timed out waiting for ${type ? type + ' ' : ''}contract deployment to address ${deployment.address} with ${
         deployment.deploymentId ? `deployment id ${deployment.deploymentId}` : `transaction ${deployment.txHash}`
--- a/packages/core/src/deployment.ts
+++ b/packages/core/src/deployment.ts
@@ -54,6 +54,7 @@ export interface DeployOpts {
  *   If not provided, invalid deployments will not be deleted in a dev network (which is not a problem if merge is false,
  *   since it will be overwritten with the new deployment).
  * @param merge whether the cached deployment is intended to be merged with the new deployment. Defaults to false.
+ * @param getDeploymentResponse a function to get the deployment response by id.
  * @returns the cached deployment if it should be used, otherwise the new deployment from the deploy function
  * @throws {InvalidDeployment} if the cached deployment is invalid and we are not on a dev network
  */
@@ -65,8 +66,9 @@ export async function resumeOrDeploy<T extends Deployment, U extends T = T>(
   opts?: DeployOpts,
   deployment?: ManifestField<T>,
   merge?: boolean,
+  getDeploymentResponse?: (deploymentId: string, catchIfNotFound: boolean) => Promise<DeploymentResponse | undefined>,
 ): Promise<T | U> {
-  const validated = await validateCached(cached, provider, type, opts, deployment, merge);
+  const validated = await validateCached(cached, provider, type, opts, deployment, merge, getDeploymentResponse);
   if (validated === undefined || merge) {
     const deployment = await deploy();
     debug('initiated deployment', 'transaction hash:', deployment.txHash, 'merge:', merge);
@@ -83,10 +85,11 @@ async function validateCached<T extends Deployment>(
   opts?: DeployOpts,
   deployment?: ManifestField<T>,
   merge?: boolean,
+  getDeploymentResponse?: (deploymentId: string, catchIfNotFound: boolean) => Promise<DeploymentResponse | undefined>,
 ): Promise<T | undefined> {
   if (cached !== undefined) {
     try {
-      await validateStoredDeployment(cached, provider, type, opts, merge);
+      await validateStoredDeployment(cached, provider, type, opts, merge, getDeploymentResponse);
     } catch (e) {
       if (e instanceof InvalidDeployment && (await isDevelopmentNetwork(provider))) {
         debug('ignoring invalid deployment in development network', e.deployment.address);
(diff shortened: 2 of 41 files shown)


packages/core/scripts/copy-build-info.js
```diff
@@ -47,10 +47,48 @@ assert(
...
+assert(origSelection.includes('evm.bytecode'));
+assert(origSelection.includes('evm.deployedBytecode'));
+assert(origSelection.includes('evm.methodIdentifiers'));
+assert(origSelection.includes('metadata'));
+
+input.settings.outputSelection['*'] = {
+  '*': ['abi', 'evm.bytecode', 'evm.deployedBytecode', 'evm.methodIdentifiers', 'metadata'],
+};
+
+// Keep only relevant sections of build info
```
frangio, line 79: What do we intend to remove? Can we specify that in the comment?

If I understand correctly we're removing only from the `output` object and keeping only `output.contracts` as filtered above, but also removing the AST for example.

I assume this is to reduce the package size? We could also specify that in a comment somewhere.

packages/core/scripts/copy-build-info.js
```diff
@@ -47,10 +47,48 @@ assert(
...
+buildInfo = {
+  _format: buildInfo._format,
```
frangio, line 81: I'm (very slightly) worried that we're modifying the file structure but preserving the `_format` string as is. Perhaps we should change it to like `${buildInfo._format}-slim1`.

packages/core/scripts/copy-build-info.js
```diff
@@ -47,10 +47,48 @@ assert(
...
+  for (const contractName in contractNames) {
+    contractNames[contractName] = {
+      abi: contractNames[contractName].abi,
+      evm: contractNames[contractName].evm,
+      metadata: contractNames[contractName].metadata, // metadata is needed to determine the license type
+    };
+  }
+}
+
+// Adjust correponding output selection
```
frangio, line 65: This is clever... but we should test that it actually works and that the end result can be verified. I don't trust the Solidity compiler to produce exactly the same bytecode with the modified settings. Perhaps we should: 1) check with the Solidity team if they think this should work, and 2) manually check that we can verify contracts like this (you may have already tried?).

packages/core/src/impl-store.ts
```diff
@@ -36,6 +43,7 @@ async function fetchOrDeployGeneric<T extends Deployment, U extends T = T>(
   deploy: () => Promise<U>,
   opts?: DeployOpts,
   merge?: boolean,
+  getDeploymentResponse?: (deploymentId: string, allowUndefined: boolean) => Promise<DeploymentResponse | undefined>,
```
frangio, line 46: This new param is missing documentation. Same for the other functions.

packages/core/src/impl-store.ts
```diff
@@ -180,11 +205,18 @@ async function checkForAddressClash(
       debug('deleting a previous deployment at address', updated.address);
       clash.set(undefined);
     } else {
-      throw new Error(
-        `The following deployment clashes with an existing one at ${updated.address}\n\n` +
-          JSON.stringify(updated, null, 2) +
-          `\n\n`,
-      );
+      const existing = clash.get();
+      // it's a clash if there is no deployment id or if deployment ids don't match
```
frangio, line 209: Can you describe for me in what scenario we would see this? Finding an existing entry with the same deployment id.

frangio (commented):
Very early comments, still a lot to go through.

docs/modules/ROOT/pages/platform-deploy.adoc
```diff
@@ -0,0 +1,87 @@
...
+  }
+}
+----
+
+[[usage]]
+== Usage
+
+When using the xref:api-hardhat-upgrades.adoc[Hardhat Upgrades API functions], enable OpenZeppelin Platform deployments using any of the ways below.
+
+NOTE: Only functions that have the `platform` option in their API reference support deployments through the OpenZeppelin Platform. If you enable the following but use functions that do not support `platform`, those functions will deploy using ethers.js.
```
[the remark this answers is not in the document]
ericglau: Scripts are meant to be reusable between a local dev network, testnets, and production.  The recommendation would be to set `deploy: true` in the config when using testnets/production (where platform will be used), so that the scripts don't need to be modified from when using a local network.

So there could be two modes:
- If using `platform.<function>` or using the option `{platform: true}`, then give an error if the function does not support platform.
- If using `deploy: true`, then just fall back if the function does not support platform, as described in the note.  If the function does support platform, but platform is not available (e.g. API key is not set), it would still give an error.

packages/core/src/impl-store.ts
ericglau: When calling resumeOrDeploy before checkForAddressClash, resumeOrDeploy may detect that the transaction hash changed for the deployment id.  In this case, it returns an updated entry for the deployment.

We then call checkForAddressClash with this updated entry.  The address and deployment ids match, but the transaction hash changed, so we don't want this to be treated as a clash.  We just want to update the manifest.

I'll add a test for this.

ericglau: Added test in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/763/commits/a9b5147abdd3256d974ff2a803ae5302232c0903

packages/core/scripts/copy-build-info.js
ericglau: I was able to verify a contract using this.

Would it be safer (although a bit incorrect) to keep the original output selection here, even if we remove it from the actual output?

packages/plugin-hardhat/src/index.ts
```diff
@@ -157,6 +132,71 @@ if (tryRequire('@nomiclabs/hardhat-etherscan')) {
...
+    beacon: {
+      getImplementationAddress: (beaconAddress: string) =>
+        getImplementationAddressFromBeacon(hre.network.provider, beaconAddress),
+    },
+    proposeUpgrade: makeProposeUpgrade(hre, platform),
+    verifyDeployment: makeVerifyDeploy(hre),
+    verifyDeploymentWithUploadedArtifact: makeVerifyDeployWithUploadedArtifact(hre),
+    getDeploymentArtifact: makeGetVerifyDeployArtifact(hre),
+    getDeploymentBuildInfo: makeGetVerifyDeployBuildInfo(hre),
+    getBytecodeDigest: makeGetBytecodeDigest(hre),
```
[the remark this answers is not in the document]
ericglau: These are cloned from the `hardhat-defender` package.  We could make these only available for `platform.<fn>` and not `upgrades.<fn>`

However, `deployContract` and `proposeUpgrade` should be considered separately since they take the `platform` option.

`upgrades.deployContract` without `platform` option is not allowed anyways and would fail.  While the syntax `upgrades.deployContract(...{platform: true})` would be consistent with what we allow for `deployProxy`, I no longer see any real benefit to allowing it for this case (other than having one less import/require in the user's code).

`upgrades.proposeUpgrade` without `platform` option could make sense.  This would mean deploy the new implementation using ethers.js, then submit the proposal on Platform (even though Platform is not part of the name).

---
To summarize:

Support:
```
platform.deployContract
platform.verifyDeployment
platform.verifyDeploymentWithUploadedArtifact
platform.getDeploymentArtifact
platform.getDeploymentBuildInfo
platform.getBytecodeDigest
platform.proposeUpgrade
upgrades.proposeUpgrade
```

Do not support:
```
upgrades.deployContract
upgrades.verifyDeployment
upgrades.verifyDeploymentWithUploadedArtifact
upgrades.getDeploymentArtifact
upgrades.getDeploymentBuildInfo
upgrades.getBytecodeDigest
```

Let me know if you agree.

docs/modules/ROOT/pages/platform-deploy.adoc
```diff
@@ -0,0 +1,87 @@
...
+    apiKey: process.env.API_KEY,
+    apiSecret: process.env.API_SECRET,
+  }
+}
+----
+
+[[usage]]
+== Usage
+
+When using the xref:api-hardhat-upgrades.adoc[Hardhat Upgrades API functions], enable OpenZeppelin Platform deployments using any of the ways below.
```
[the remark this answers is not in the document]
ericglau: Let me know what you think of this:
- If you want to use the same scripts on local devnets and Platform, use `upgrades.<fn>` with `deploy: true` in the config that you use with Platform.
- If you only run your scripts on Platform, use `platform.<fn>` for clarity.

packages/core/scripts/copy-build-info.js
ericglau: Any benefit to include an `oz` marker for easier identification? `${buildInfo._format}-oz-slim1`

docs/modules/ROOT/pages/platform-deploy.adoc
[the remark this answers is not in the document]
frangio: That sounds reasonable.

docs/modules/ROOT/pages/platform-deploy.adoc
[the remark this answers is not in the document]
frangio: I would say to use `platform.<fn>` when the user wants to make sure Platform is used and wants to see an error otherwise, but mostly they should use the first option you describe.

packages/plugin-hardhat/src/index.ts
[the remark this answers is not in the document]
frangio: What do you think of putting `platform` under `upgrades`?
So we would have
```
upgrades: {
  deployProxy: ...,
  ...,
  platform: {
    deployProxy: ..., // aliased with { platform: true }
    deployContract: ..., // only at this level
    ...
  }
}
```

packages/core/scripts/copy-build-info.js
frangio: Yeah I'd feel more confident in that. I think that + using the `-slim` format would be good.

packages/core/scripts/copy-build-info.js
frangio: Sounds good too.

packages/core/src/deployment.ts
```diff
@@ -94,58 +101,113 @@ async function validateCached<T extends Deployment>(
...
+    const existingBytecode = await getCode(provider, deployment.address);
     if (isEmpty(existingBytecode)) {
-      throw new InvalidDeployment(stored);
+      throw new InvalidDeployment(deployment);
     }
   }
+  return deployment;
+}
+
+export interface DeploymentResponse {
```
frangio, line 161: "Deployment response" doesn't seem like the right concept to me... I think it should be either `DeploymentStatus` or something like `RemoteDeployment` or even `PlatformDeployment`. This would apply as well to the function names.

I'm thinking we should go with the latter and use `remoteDeploymentId` in the manifest. `deploymentId` is very generic.

packages/core/src/impl-store.ts
frangio: I realize now that the entire function is missing documentation. :stuck_out_tongue: That's on me. Could be a good moment to add some though.

frangio: I realize now that the entire function is missing documentation. :stuck_out_tongue: That's on me. Could be a good moment to add some though.

packages/core/src/deployment.ts
```diff
@@ -217,3 +281,38 @@ export class InvalidDeployment extends Error {
...
+export async function isDeploymentCompleted(
+  address: string,
+  deploymentId: string,
+  getDeploymentResponse: (allowUndefined: boolean) => Promise<DeploymentResponse | undefined>,
+): Promise<boolean> {
+  debug('verifying deployment id', deploymentId);
+  const response = await getDeploymentResponse(false);
+  if (response === undefined) {
+    throw new Error(`Broken invariant: Response not found for deployment id ${deploymentId}`);
+  }
```
frangio, line 303: It seems like the only place where we use `allowUndefined = false` we check the result anyway. I would simplify `getDeploymentResponse` removing the `allowUndefined` argument, making it always return `undefined` if the deployment is not found.

In this function I would also make `getDeploymentResponse` a function that takes a deployment id, instead of hardcoding it at the callsite.

I think both of these changes should make invocation simpler because you would just pass along the `getDeploymentResponse` function.

packages/plugin-hardhat/contracts/DeployContract.sol
```diff
@@ -0,0 +1,22 @@
...
+contract NonUpgradeable {
+}
+
+contract IsInitializable is Initializable1 {
+}
+
+contract IsInitializableUpgradeable is Initializable2 {
+}
+
+contract IsUUPS is UUPSUpgradeable {
```
frangio, line 17:
```suggestion
// Insecure contract
contract IsUUPS is UUPSUpgradeable {
```

packages/core/src/deployment.ts
```diff
@@ -22,6 +22,10 @@ export interface Deployment {
   txHash?: string;
 }
 
+export interface DeploymentId {
+  deploymentId?: string;
```
frangio, line 26: What's the reason not to add this directly in `Deployment`?

packages/plugin-hardhat/src/utils/options.ts
```diff
@@ -32,16 +32,29 @@ type Initializer = {
...
-export type DeployProxyOptions = StandaloneOptions & Initializer;
+export type Platform = {
+  platform?: boolean;
+};
+
+export type PlatformSupportedOptions = Platform & {
+  verifySourceCode?: boolean;
+};
+
+export type DeployBeaconProxyOptions = DeployOpts & ProxyKindOption & Initializer & PlatformSupportedOptions;
```
frangio, line 43: What's the reason `DeployOpts` was added here but not in all `Deploy*Options` types?

packages/plugin-hardhat/src/prepare-upgrade.ts
```diff
@@ -23,8 +23,12 @@ export type PrepareUpgradeFunction = (
   opts?: PrepareUpgradeOptions,
 ) => Promise<DeployImplementationResponse>;
 
-export function makePrepareUpgrade(hre: HardhatRuntimeEnvironment): PrepareUpgradeFunction {
+export function makePrepareUpgrade(hre: HardhatRuntimeEnvironment, platformModule: boolean): PrepareUpgradeFunction {
   return async function prepareUpgrade(proxyOrBeacon, ImplFactory, opts: PrepareUpgradeOptions = {}) {
+    if (platformModule && opts.platform === undefined) {
+      opts.platform = true;
+    }
```
frangio, line 30: Similar comment about mutation here. Plus let's have an `enablePlatform` helper?

packages/plugin-hardhat/src/type-extensions.ts
```diff
@@ -5,5 +5,22 @@ import type { HardhatUpgrades } from '.';
...
   export interface HardhatRuntimeEnvironment {
     upgrades: HardhatUpgrades;
+    platform: HardhatUpgrades;
+  }
+}
+
+export interface HardhatPlatformConfig {
+  apiKey: string;
+  apiSecret: string;
+  deploy?: boolean;
```
frangio, line 15: This option name is a little confusing I think. Perhaps we should do `useDeploy` or `preferDeploy`.

packages/plugin-hardhat/src/deploy-proxy.ts
```diff
@@ -73,6 +84,13 @@ export function makeDeployProxy(hre: HardhatRuntimeEnvironment): DeployFunction
     const inst = ImplFactory.attach(proxyDeployment.address);
     // @ts-ignore Won't be readonly because inst was created through attach.
     inst.deployTransaction = proxyDeployment.deployTransaction;
+    if (opts.platform && proxyDeployment.deploymentId !== undefined) {
+      inst.deployed = async () => {
+        assert(proxyDeployment.deploymentId !== undefined);
+        await waitForDeployment(hre, opts, inst.address, proxyDeployment.deploymentId);
+        return inst;
```
frangio, line 91: Is there a chance we might want to update `inst.deployTransaction` here?

Should we also add a call to the original `inst.deployed()` at the end?

packages/plugin-hardhat/src/deploy-beacon-proxy.ts
```diff
@@ -72,16 +80,23 @@ export function makeDeployBeaconProxy(hre: HardhatRuntimeEnvironment): DeployBea
...
     const inst = attachTo.attach(proxyDeployment.address);
     // @ts-ignore Won't be readonly because inst was created through attach.
     inst.deployTransaction = proxyDeployment.deployTransaction;
+    if (opts.platform && proxyDeployment.deploymentId !== undefined) {
+      inst.deployed = async () => {
+        assert(proxyDeployment.deploymentId !== undefined);
+        await waitForDeployment(hre, opts, inst.address, proxyDeployment.deploymentId);
+        return inst;
+      };
+    }
```
frangio, line 99: Same here regarding updating `deployTransaction`. We might want to define a helper for this entire piece of code.

packages/plugin-hardhat/src/platform/verify-deployment.ts
frangio, line 1: What is the difference between `verifyDeploy` and `verifyDeployWithUploadedArtifact`? Do we need both? Maybe we can merge them.

I couldn't find docs for this file and I'm not sure what to make of this functionality.

packages/plugin-hardhat/src/platform/deploy.ts
```diff
@@ -0,0 +1,194 @@
...
+  const client = getPlatformClient(hre);
+
+  const constructorArgs = [...args] as (string | number | boolean)[];
+  const contractInfo = await getContractInfo(hre, factory, { constructorArgs, ...opts });
+  const network = await getNetwork(hre);
+  debug(`Network ${network}`);
+
+  const verifySourceCode = opts.verifySourceCode ?? true;
+  if (verifySourceCode) {
+    await registerEtherscanApiKey(hre, network, client.BlockExplorerApiKey);
```
frangio, line 67: As discussed offline we may want to reconsider doing this automatically.

packages/plugin-hardhat/src/platform/deploy.ts
```diff
@@ -0,0 +1,194 @@
...
+    for (const key of registeredKeys) {
+      if (key.network === network) {
+        return true;
+      }
+    }
+    return false;
+  }
+}
+
+function getContractPathAndName(fullyQualified: string) {
```
frangio, line 127:
```ts
import { parseFullyQualifiedName } from "hardhat/utils/contract-names";
```

packages/plugin-hardhat/src/platform/deploy.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  const metadata = JSON.parse(metadataString);
+
+  const license = metadata.sources[contractInfo.contractPath].license;
+  if (license === undefined) {
+    debug('License not found in metadata');
+    return undefined;
+  }
+
+  debug(`Found license from metadata: ${license}`);
```
frangio, line 192:
```suggestion
  if (license === undefined) {
    debug('License not found in metadata');
  } else {
    debug(`Found license from metadata: ${license}`);
  }
```

packages/plugin-hardhat/src/platform/propose-upgrade.ts
```diff
@@ -0,0 +1,113 @@
...
+    } else {
+      // try getting the implementation address so that it will give an error if it's not a transparent/uups proxy
+      await getImplementationAddress(hre.network.provider, proxyAddress);
+    }
+
+    const implFactory =
+      typeof contractNameOrImplFactory === 'string'
+        ? await hre.ethers.getContractFactory(contractNameOrImplFactory)
+        : contractNameOrImplFactory;
+    const contractName = typeof contractNameOrImplFactory === 'string' ? contractNameOrImplFactory : undefined;
```
frangio, line 62: So if you want verification you need to provide the contract name as opposed to the impl factory?

packages/plugin-hardhat/src/platform/propose-upgrade.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    let txResponse;
+    let newImplementation: string;
+
+    if (typeof prepareUpgradeResult === 'string') {
+      newImplementation = prepareUpgradeResult;
+    } else {
+      txResponse = prepareUpgradeResult;
+      // The txResponse is a create2 transaction, so the deployment address is not easily retrievable from it.
+      // Instead, run the prepareUpgrade again but this time adjust the options to request the address and ensure it does not attempt another deployment.
```
frangio, line 78: This seems hackish, do we have other options?

packages/plugin-hardhat/src/index.ts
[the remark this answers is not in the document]
ericglau: Removed everything related to `verifyDeployment` since that is handled automatically by the Platform deploy API.

I don't think `upgrades.platform.deployContract` makes sense since it is unrelated to upgradeable contracts.

Now since only `deployContract` and `proposeUpgrade` are new, I changed it to this:
```
upgrades: {
  deployProxy: ...,
  prepareUpgrade: ...,
  ...,
},
platform: {
  deployProxy: ..., // aliased with { platform: true }
  prepareUpgrade: ..., // aliased with { platform: true }
  ...,
  deployContract: ..., // only at this level
}
```

packages/plugin-hardhat/src/prepare-upgrade.ts
ericglau: The helper is named as `withPlatformDefaults`

packages/plugin-hardhat/src/platform/propose-upgrade.ts
ericglau: Removed this since verification is handled automatically by Deploy API

packages/plugin-hardhat/src/platform/verify-deployment.ts
ericglau: Removed since this is handled automatically by Deploy API

packages/plugin-hardhat/src/platform/deploy.ts
ericglau: Removed since this is expected to be handled automatically by the Platform

packages/core/src/deployment.ts
ericglau: There are some cases where we use `Required<Deployment>` to enforce that a deployment has a `txHash` defined.

For example, `deployProxy` previously used `Required<ProxyDeployment & DeployTransaction>` where `ProxyDeployment` extends `Deployment`.

We want the `deploymentId` to be optional, because it is only relevant when using Platform deploys.  So it cannot be part of `Deployment` if the above pattern continues to be used.

So now, `deployProxy` can use `Required<ProxyDeployment & DeployTransaction> & RemoteDeploymentId;`

packages/plugin-hardhat/src/utils/options.ts
ericglau: The other `Deploy*Options` types inherit `StandaloneOptions` which already has `DeployOpts`.  Added doc comments to clarify some of the inherited options.

docs/modules/ROOT/pages/platform-deploy.adoc
```diff
@@ -0,0 +1,87 @@
+= Using with OpenZeppelin Platform
+
+The Hardhat Upgrades package can use the OpenZeppelin Platform for deployments instead of ethers.js.
```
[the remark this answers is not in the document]
ericglau: Added a brief note but we should plan to improve this when there is more documentation to link to.

docs/modules/ROOT/pages/platform-deploy.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+[[configuration]]
+== Configuration
+
+Create a deployment environment on the OpenZeppelin Platform and provide the Team API Key and secret in your `hardhat.config.js` or `hardhat.config.ts` file under `platform`:
```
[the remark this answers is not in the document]
ericglau: We should plan to improve this when there is more documentation to link to

packages/plugin-hardhat/src/platform/propose-upgrade.ts
ericglau: Improved to use the deployed impl object (containing both txResponse and impl address) directly.

packages/plugin-hardhat/src/utils/deploy-impl.ts
```diff
@@ -116,17 +116,15 @@ async function deployImpl(
...
+    remoteDeploymentId => getRemoteDeployment(hre, remoteDeploymentId),
   );
 
   let txResponse;
   if (opts.getTxResponse) {
     if ('deployTransaction' in deployment) {
       txResponse = deployment.deployTransaction;
     } else if (deployment.txHash !== undefined) {
-      txResponse = hre.ethers.provider.getTransaction(deployment.txHash);
+      txResponse = await hre.ethers.provider.getTransaction(deployment.txHash);
```
frangio, line 127: This error was possible because this function returns `Promise<any>`. A cautionary tale! Is there any chance we could get rid of that `any`?

frangio (approved):
Changes look good!

PR #764, opened 2023-03-24
optimize isConstructor gas usage

zhiqiangxu:
It seems no particular reason to use `extcodesize` .


zhiqiangxu:
In constructor, `codesize` actually returns the size of `initcode`.

PR #766, opened 2023-03-27
Add sepolia testnet in networkNames

marc-aurele-besner:
## Add sepolia testnet in networkNames

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


ericglau:
Thanks for this PR!  I agree that Sepolia should be added, but Zhejiang is a short-lived testnet only for testing withdrawals and would be shut down soon.  Can you please update the PR to only include Sepolia?

marc-aurele-besner:
> Thanks for this PR! I agree that Sepolia should be added, but Zhejiang is a short-lived testnet only for testing withdrawals and would be shut down soon. Can you please update the PR to only include Sepolia?

Sure :+1:

ericglau (approved):

Issue #752, opened 2023-02-27
Verify task should fall back to hardhat-etherscan if proxy bytecode does not match


ericglau:
When calling the Hardhat Upgrades `verify` task with a proxy address, the task checks if the existing proxy's bytecode matches with the expected artifact bytecode that would normally be deployed by the Hardhat Upgrades plugin.  If it does not match, the proxy verification fails with the error `Bytecode does not match with the current version of ERC1967Proxy in the Hardhat Upgrades plugin.`  We should enhance this to try falling back to using the regular `hardhat-etherscan` `verify` task instead of failing right away.

One reason the bytecode would not match is if the proxy was compiled and deployed by the user with their own project's compiler settings, for example if using a proxy factory as described in [this forum thread](https://forum.openzeppelin.com/t/unverified-erc1967-proxy-goerli/34875).  In this case, the regular `hardhat-etherscan` `verify` task would be able to verify this proxy.

PR #776, opened 2023-04-25
Fall back to hardhat-etherscan verify if bytecode does not match artifacts

ericglau:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/752

When verifying, if the contract does match with known artifacts from the plugin, fall back to regular hardhat verify to use source code in the user's project.

If the fall back verification passed for the given contract address, but there were issues while verifying related contracts (such as a linked implementation), print warnings for those instead of failing.


packages/plugin-hardhat/src/verify-proxy.ts
```diff
@@ -343,7 +384,71 @@ async function searchEvent(
...
+) {
+  try {
+    await attemptVerifyWithCreationEvent(hre, etherscanApi, address, possibleContractInfo, errorReport);
+    return true;
+  } catch (origError: any) {
+    if (origError instanceof BytecodeNotMatchArtifact || origError instanceof EventNotFound) {
+      // Try falling back to regular hardhat verify in case the source code is available in the user's project.
+      try {
+        await hardhatVerify(address);
+        console.log(`Successfully verified contract at ${address}.`);
```
frangio, line 421: Seems like this would be printed by `hardhatVerify` itself, no?

packages/plugin-hardhat/src/verify-proxy.ts
```diff
@@ -191,12 +218,14 @@ async function fullVerifyTransparentOrUUPS(
...
-        await verifyContractWithCreationEvent(
+        await verifyWithArtifactOrFallback(
           hre,
+          hardhatVerify,
           etherscanApi,
           adminAddress,
           [verifiableContracts.proxyAdmin],
-          errors,
+          errorReport,
+          false,
```
frangio, line 228: Can you explain why we're treating this one differently in this way?

packages/plugin-hardhat/src/verify-proxy.ts
```diff
@@ -343,7 +384,71 @@ async function searchEvent(
...
+    return true;
+  } catch (origError: any) {
+    if (origError instanceof BytecodeNotMatchArtifact || origError instanceof EventNotFound) {
+      // Try falling back to regular hardhat verify in case the source code is available in the user's project.
+      try {
+        await hardhatVerify(address);
+        console.log(`Successfully verified contract at ${address}.`);
+      } catch (fallbackError: any) {
+        if (fallbackError.message.toLowerCase().includes('already verified')) {
+          console.log(`Contract at ${address} already verified.`);
```
frangio, line 424: Again I would expect some sort of message of this kind to already be printed by hardhat-verify itself. Doesn't it do that?

packages/plugin-hardhat/src/verify-proxy.ts
```diff
@@ -343,7 +384,71 @@ async function searchEvent(
...
+            recordError(origError.message, errorReport);
+          }
+
+          recordError(`Failed to verify directly using hardhat verify: ${fallbackError.message}`, errorReport);
+          return;
+        }
+      }
+
+      // Since the contract was able to be verified directly, we don't want the task to fail so we should convert earlier errors into warnings for other related contracts.
+      // For example, the user provided constructor arguments for the verify command will apply to all calls of the regular hardhat verify,
```
frangio, line 439: >the user provided constructor arguments for the verify command will apply to all calls

What do you mean by this?

packages/plugin-hardhat/src/verify-proxy.ts
ericglau: It throws an error with 'Contract source code already verified', but we want to catch it (not fail) and log with the address for context.

packages/plugin-hardhat/src/verify-proxy.ts
ericglau: Added reasoning in comments.

packages/plugin-hardhat/src/verify-proxy.ts
ericglau: The `hardhat verify` command normally takes constructor args like:
```
npx hardhat verify <ADDRESS> --constructor-args arguments.js
```
where arguments.js contains an object with the constructor arguments that will be used for verification, for example:
```
module.exports = [
  "arg1",
  "arg2",
];
```

In our plugin, we override this task to verify both the implementation and the proxy consecutively.

When we verify the implementation, hardhat verify is called on the implementation with user-provided constructor args `arg1` and `arg2`

When we verify the proxy, normally we [infer the proxy's constructor args](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/83c5f37b7ef55201989f7027dcd1ac4990192602/packages/plugin-hardhat/src/verify-proxy.ts#L542) by looking at the last portion of the creation code (removing the known artifact bytecode as the prefix), and use that to verify with the Etherscan API.

But if we DON'T know the artifact bytecode, we cannot infer the proxy's constructor args.  So the user needs to provide them as well, then we call hardhat verify with user-provided constructor args which need to be the impl address and initializer data (when verifying ERC1967Proxy).  

So the user can either pass in `arg1` and `arg2`, or impl address and initalizer data.  They can't pass in both sets at the same time.  One of them will fail.

I expect the user's scenario may look like this:
1. run hardhat verify with implementation's constructor args. impl verifies, proxy fails (gets an error about wrong number of constructor parameters). summarize with an error.
2. run hardhat verify again with the proxy's constructor args. impl fails (even though it is already verified -- gets a warning about wrong number of constructor parameters), proxy succeeds. summarize with a warning.

frangio (approved):

PR #778, opened 2023-04-25
Support prepareUpgrade from an implementation address

### LINKED ISSUE #777: Support prepareUpgrade from an implementation address
`prepareUpgrade` performs upgrade safety validation and deploys a new implementation contract, but requires a beacon or a proxy address as an argument to represent the contract that is being upgraded from.

Improve this to also support an implementation contract's address, so new versions of that implementation can be deployed without having to know about any existing proxy deployments.  One use case for this is if the proxy deployments are owned by other people.  Then after `prepareUpgrade` is run, those other users can decide to upgrade to the newly prepared implementation.

See related [forum thread](https://forum.openzeppelin.com/t/questions-about-the-openzeppelin-upgrades-file/35867/8?u=ericglau).

ericglau:
Fixes #777

### DIFF
--- a/docs/modules/ROOT/pages/api-hardhat-upgrades.adoc
+++ b/docs/modules/ROOT/pages/api-hardhat-upgrades.adoc
@@ -359,7 +359,7 @@ await upgrades.validateUpgrade(Box, BoxV2);
 [source,ts]
 ----
 async function prepareUpgrade(
-  proxyOrBeacon: string | ethers.Contract,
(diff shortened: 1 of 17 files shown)


julianmrodri:
Wow that was fast. Awesome!!! We really need this.

frangio (approved):
LGTM

julianmrodri:
@ericglau thanks a lot for this. Just to plan accordingly, any expected date or timeframe for this to be released?

PR #784, opened 2023-05-01
Allow custom signer for admin functions

### LINKED ISSUE #782: TransferProxyAdminOwnership in hardhat plugin does not support dynamic signer
When calling TransferProxyAdminOwnership in the hardhat plugin, the function will attach the admin contract with the default hardhat account. I cannot use a dynamic signer to call the function in my script, which results in the error of 'caller is not owner'

ericglau:
Fixes #782 

Supports passing in an optional Signer parameter to `admin.changeProxyAdmin` and `admin.transferProxyAdminOwnership`

### DIFF
--- a/docs/modules/ROOT/pages/api-hardhat-upgrades.adoc
+++ b/docs/modules/ROOT/pages/api-hardhat-upgrades.adoc
@@ -442,7 +442,7 @@ Similar to `prepareUpgrade`. This method validates and deploys the new implement
 [source,ts]
 ----
 async function deployProxyAdmin(
-  Signer?: ethers.Signer,
+  signer?: ethers.Signer,
   opts?: {
     timeout?: number,
     pollingInterval?: number,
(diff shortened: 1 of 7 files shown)


docs/modules/ROOT/pages/api-hardhat-upgrades.adoc
```diff
@@ -474,6 +474,7 @@ Deploys a https://docs.openzeppelin.com/contracts/4.x/api/proxy#ProxyAdmin[proxy
 async function changeProxyAdmin(
   proxyAddress: string,
   newAdmin: string,
+  Signer?: ethers.Signer,
```
frangio, line 477: The docs should say lowercase `signer`, right?

frangio (approved):
This seems good, but note the comment above about docs.

picardcapt1212:
thanks a lot. i was searching for this👍

PR #786, opened 2023-05-05
Improve Platform functions

ericglau:
Fixes LIB-856

- Move `proposeUpgrade` to be in `platform` module only.
- Use `usePlatformDeploy` as the option name and Hardhat config option name for enabling/disabling Platform deployments.
- Add `walletId?` option for Platform deployments to customize the wallet id to use.
- Use new Platform Upgrade API for `platform.proposeUpgrade` which simplifies input options and can include returning a Safe App url.
- Add `platform.getDefaultApprovalProcess` as a convenience function for users to get the default approval process with its associated address (e.g. multisig or governor address) that they should assign as owner of their proxies.
- Improve license type handling: Convert no license or 'UNLICENSED' to 'None' for Etherscan verification. Catch invalid license type from API and wrap with more useful error message.

(Note: tests will fail in the meantime until new version of https://www.npmjs.com/package/platform-deploy-client is released that includes some required updates)


ericglau:
Replaced by https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/791

PR #799, opened 2023-05-18
Support `salt` option for deployments on OpenZeppelin Platform

ericglau:
When using Platform, deployments are performed using the CREATE2 opcode. This option allows users to provide the salt for the deployment. If not provided, Platform defaults to using a random salt.


frangio (approved):
LGTM

PR #800, opened 2023-05-18
Add forceDeployImplementation option

ericglau:
Adds an option to force the deployment of the implementation contract even if it was previously deployed with the same bytecode. This can be used with the `salt` option when deploying a proxy through OpenZeppelin Platform to ensure that the new proxy has a matching implementation deployed with the same salt.


ericglau:
Note that this introduces `forceDeployImplementation` as the opposite of `useDeployedImplementation`, but these options by themselves are not completely intuitive. 

An alternative could be to have a single option with different "modes" controlling whether the implementation is deployed, for example an option such as `deployImplementation?: 'force' | 'never' | 'onchange'`, with `onchange` being the default.  However, there is already a function called `deployImplementation` so this would be slightly confusing as well.

@frangio let me know what you think.

ericglau:
Replaced by #804

PR #804, opened 2023-05-23
Add `redeployImplementation` option

ericglau:
Adds the following option: `redeployImplementation`: (`"always" | "never" | "onchange"`) Determines whether the implementation contract will be redeployed. Defaults to `"onchange"`.
* If set to `"always"`, the implementation contract is always redeployed even if it was previously deployed with the same bytecode. This can be used with the `salt` option when deploying a proxy through OpenZeppelin Platform to ensure that the implementation contract is deployed with the same salt as the proxy.
* If set to `"never"`, the implementation contract is never redeployed. If the implementation contract was not previously deployed or is not found in the network file, an error will be thrown.
* If set to `"onchange"`, the implementation contract is redeployed only if the bytecode has changed from previous deployments.

Deprecates `useDeployedImplementation` which is equivalent to `redeployImplementation = "never"`

### DIFF
--- a/packages/plugin-hardhat/src/utils/deploy-impl.ts
+++ b/packages/plugin-hardhat/src/utils/deploy-impl.ts
@@ -99,6 +99,14 @@ async function deployImpl(
 ): Promise<DeployedImpl> {
   const layout = deployData.layout;
 
+  if (opts.useDeployedImplementation && opts.forceDeployImplementation) {
+    throw new UpgradesError(
+      'The useDeployedImplementation and forceDeployImplementation options cannot both be set to true at the same time',
+    );
+  }
+
+  const merge = opts.forceDeployImplementation;
+
   const deployment = await fetchOrDeployGetDeployment(
     deployData.version,
     deployData.provider,
@@ -119,7 +127,7 @@ async function deployImpl(
       return { ...deployment, layout };
     },
     opts,
-    undefined,
+    merge,
     remoteDeploymentId => getRemoteDeployment(hre, remoteDeploymentId),
   );
 
--- a/packages/plugin-hardhat/src/utils/deploy-impl.ts
+++ b/packages/plugin-hardhat/src/utils/deploy-impl.ts
@@ -99,13 +99,13 @@ async function deployImpl(
 ): Promise<DeployedImpl> {
   const layout = deployData.layout;
 
-  if (opts.useDeployedImplementation && opts.forceDeployImplementation) {
+  if (opts.useDeployedImplementation && opts.redeployImplementation !== undefined) {
     throw new UpgradesError(
-      'The useDeployedImplementation and forceDeployImplementation options cannot both be set to true at the same time',
+      'The useDeployedImplementation and redeployImplementation options cannot both be set at the same time',
     );
   }
 
-  const merge = opts.forceDeployImplementation;
+  const merge = opts.redeployImplementation === 'always';
 
   const deployment = await fetchOrDeployGetDeployment(
     deployData.version,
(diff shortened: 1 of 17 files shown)


packages/plugin-hardhat/src/utils/options.ts
```diff
@@ -25,7 +29,8 @@ export function withDefaults(opts: UpgradeOptions = {}): Required<UpgradeOptions
     constructorArgs: opts.constructorArgs ?? [],
     timeout: opts.timeout ?? 60e3,
     pollingInterval: opts.pollingInterval ?? 5e3,
-    useDeployedImplementation: opts.useDeployedImplementation ?? true,
+    useDeployedImplementation: opts.useDeployedImplementation ?? false,
```
frangio, line 32: I'm confused by a couple of things here.

1. Why would the default change?
2. Was the default really `true`? I thought our default was always the `onchange` behavior.

See my other comment. I think we should be returning an options object that is self-consistent. Can we return an object without the `useDeployedImplementation` field?

frangio: I see now that if the user set `useDeployedImplementation` in their config we want to refer to that in the error message. This makes sense.

Perhaps the way it's implemented currently is the best option, let me know what you think. My comment was because I would expect the deprecated option to be handled in this location instead of in the other places.

frangio:
Ideally, if the user specifies `salt`, we only redeploy if the existing implementation was deployed with another salt. But I understand that we may not want to implement this right now. What do you think about forcing the user to use `redeploy: 'always'` when they specify a `salt`? If we do this, we can change the default behavior in the future.

packages/plugin-hardhat/src/utils/deploy-impl.ts
```diff
@@ -99,18 +99,28 @@ async function deployImpl(
 ): Promise<DeployedImpl> {
   const layout = deployData.layout;
 
+  if (opts.useDeployedImplementation && opts.redeployImplementation !== undefined) {
+    throw new UpgradesError(
+      'The useDeployedImplementation and redeployImplementation options cannot both be set at the same time',
+    );
+  }
```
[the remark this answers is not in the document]
ericglau: Consider the following:
User specifies `useDeployedImplementation = true` but does not specify `redeployImplementation` in their script, so it is undefined.
If we treat use `withDefaults` to treat `redeployImplementation` as being `onchange`, then that will ALWAYS conflict with `useDeployedImplementation = true`.
So `useDeployedImplementation` and `redeployImplementation` need to be mutually exclusive.

The only benefit of allowing these to have integrity would be to allow `useDeployedImplementation = true` and `redeployImplementation = never` at the same time, but there is no need for this.

packages/plugin-hardhat/src/utils/options.ts
ericglau: Re: original default being `true` - that was a bug but had no consequence because we never actually used the default value from `withDefaults`.

In [deployImpl](https://github.com/ericglau/openzeppelin-upgrades/blob/305828bfcfd9f73baea259ee511064a838b69ddb/packages/plugin-hardhat/src/utils/deploy-impl.ts#L116), we use the `opts` param instead of `deployData.fullOpts`, where only `deployData.fullOpts` was populated by `withDefaults`.

I agree we could remove `useDeployedImplementation` from `withDefaults`, but this function's return signature is currently `Required<UpgradeOptions>` so we would need to get around that.

Technically the default of `redeployImplementation = onchange` seems kind of pointless as well -- we could equally dictate that `undefined` means "on change" and not allow an explicit `onchange` value.  It does help to make it clearer though.  But (similar to `useDeployedImplementation` above) is there any benefit in setting `redeployImplementation = onchange` in `withDefaults` if we never use this default value?

ericglau:
> What do you think about forcing the user to use redeploy: 'always' when they specify a salt

I don't think we should do this.  Users may want to run `proposeUpgrade` with a salt.  If they run it again with the same salt (for example, they want to propose an upgrade for another proxy), forcing `redeploy: 'always'` would result in a deployment error for the implementation because the resulting address would be the same.

frangio (approved):

packages/plugin-hardhat/src/utils/options.ts
ericglau: Leaving the defaults as-is but improved the way we consume `fullOpts` (coming from `withDefaults`) and `opts`

frangio (approved):

PR #807, opened 2023-05-31
Add CLI for upgrade safety checks

ericglau:
Adds a CLI to the upgrades-core package to perform upgrade safety checks on a build-info directory.

CLI documentation: https://docs.openzeppelin.com/upgrades-plugins/1.x/api-core


frangio (commented):
Halfway through.

packages/core/src/cli/validate/validate-upgrade-safety.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+export async function validateUpgradeSafety(
+  buildInfoDir?: string,
+  reportOpts: ReportOptions = {},
+  opts: ValidateUpgradeSafetyOptions = {},
+): Promise<SummaryReport> {
+  const fullReportOpts = withReportDefaults(reportOpts);
+
+  const buildInfoFiles = await getBuildInfoFiles(buildInfoDir);
+  const reports = validateBuildInfoContracts(buildInfoFiles, opts);
```
frangio, line 39: `validateBuildInfoContracts` is doing a lot of work here. Understanding what the command is doing requires going very "deep": `main > validateUpgadeSafety > validateBuildInfoContracts > getContractReports`.

I think we would benefit by extracting `getContractReports` out, and having something like this here:

```ts
  const sourceContracts = validateBuildInfoContracts(buildInfoFiles);
  const reports = getContractReports(sourceContracts, opts);
```

This will also put the contract reports and the summary report side by side, and may make it easier to understand how they relate.

packages/core/src/cli/validate/summary-report.ts
frangio, line 1: When I saw "summary report" I thought it was an alternative "abridged" report. But if I understand correctly now, it's just an aggregate of all the reports, right? Perhaps I would change the word "summary" to "aggregate", or "project report"?

packages/core/src/cli/validate/upgradeability-assessment.ts
```diff
@@ -0,0 +1,138 @@
...
+  } else {
+    throw new ReferenceContractNotFound(reference, origin.fullyQualifiedName);
+  }
+}
+
+function getAnnotationAssessment(contract: SourceContract): AnnotationAssessment {
+  const node = contract.node;
+
+  if ('documentation' in node) {
+    const doc = typeof node.documentation === 'string' ? node.documentation : node.documentation?.text ?? '';
```
frangio, line 88: This line comes from `src/validate/run.ts`. Let's extract it to a shared function used in both locations.

packages/core/src/cli/validate/upgradeability-assessment.ts
```diff
@@ -0,0 +1,138 @@
...
+    const initializable = inferInitializable(contractValidation);
+    return {
+      upgradeable: initializable || isUUPS,
+      uups: isUUPS,
+    };
+  }
+}
+
+function getReferenceContract(reference: string, origin: SourceContract, allContracts: SourceContract[]) {
+  const referenceContracts = allContracts.filter(c => c.fullyQualifiedName === reference || c.name === reference);
```
frangio, line 69: Does `allContracts` here consist of all contracts in the project/build or only the the upgradeable ones? I had some concerns about the linear search, but even if it's all contracts it's unlikely to be over 200 contracts so it's probably fine. Curious about your thoughts anyway.

packages/core/src/cli/validate/validate-upgrade-safety.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  return getSummaryReport(reports, fullReportOpts.suppressSummary);
+}
+
+export function withReportDefaults(cmdOpts: ReportOptions): Required<ReportOptions> {
+  return {
+    suppressSummary: cmdOpts.suppressSummary ?? false,
```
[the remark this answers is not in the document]
ericglau: This was intended to be useful for the high-level API instead of CLI, where the caller may not want a printed summary but intends to handle the `SummaryReport` in its own way.  For example, if it wants to parse the summary report and print errors in its own preferred format.  But I'm fine with removing this for simplicity.

packages/core/src/cli/validate/upgradeability-assessment.ts
```diff
@@ -0,0 +1,138 @@
...
+  /**
+   * The fully qualified name of the contract that referenced the missing contract.
+   */
+  readonly origin: string;
+
+  constructor(reference: string, origin: string) {
+    super(`Could not find contract ${reference} referenced in ${origin}.`);
+    this.reference = reference;
+    this.origin = origin;
+    this.name = 'ReferenceContractNotFound';
```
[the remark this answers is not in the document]
ericglau: If a caller is using the high level API and encounters this error, they may want to programmatically know (without parsing the error) what is the missing reference.  This allows the caller to try to resolve the issue by finding and adding the reference to the set of compiled contracts.

packages/core/src/cli/validate/upgradeability-assessment.ts
ericglau: It consists of all contracts in the build, because it is looking for the reference which may or may not look like an upgradeable contract.  We could change this to something like a map.

packages/core/src/cli/validate/upgradeability-assessment.ts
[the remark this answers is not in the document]
frangio: I meant specifically `this.name`.

packages/core/src/cli/validate/upgradeability-assessment.ts
frangio: If the number is in the hundreds we are probably fine to keep as an array with linear search. Using a map would require a big refactor probably.

frangio:
I pushed changes to the output, with a change in how we handle `console.log`. The reports don't log anything, they accumulate a string, and that gets printed at the end.

frangio (approved):

frangio (approved):

PR #808, opened 2023-06-01
Pin versioning for peer dependencies

pcaversaccio:
As title. `ethers` defaults to the breaking v6 and `@nomiclabs/hardhat-ethers` already has released a RC version for `3.0.0`, so this is a preemptive measure. You should consider adding the peer deps also in the `package.json`.


frangio (approved):
Thanks!

PR #816, opened 2023-06-13
Update dependencies to ethers v6-based versions

fvictorio:
Hi, we recently updated all of our official plugins to be based on ethers v6, and now I'm trying to get the most important community plugins to do the same.

This PR just makes the necessary changes in the dependencies, but of course the code needs to be updated. I gave it a try, but it was more work than I expected, so I think it's better to let you do it. But happy to help in any way I can :smile:


ericglau:
Thanks for the initial PR. This will be continued in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/817

PR #828, opened 2023-06-29
Fix typo and link in docs

martinvol:
there's also a broken link in 

`this information to validate that your contracts are [upgrade safe](#what-does-it-mean-for-a-contract-to-be-upgrade-safe).` 

but I couldn't figure out why easily.


frangio (approved):
Thanks!

PR #829, opened 2023-06-29
Migrate to hardhat-verify

ericglau:
Extends the `verify` task from [`@nomicfoundation/hardhat-verify`](https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-verify) instead of `@nomiclabs/hardhat-etherscan`


packages/plugin-hardhat/src/utils/etherscan-api.ts
```diff
@@ -42,18 +44,17 @@ export async function callEtherscanApi(etherscanApi: EtherscanAPIConfig, params:
  * Makes use of Hardhat Etherscan for handling cases when Etherscan API parameters are not present in config.
  */
 export async function getEtherscanAPIConfig(hre: HardhatRuntimeEnvironment): Promise<EtherscanAPIConfig> {
-  const endpoints = await hre.run('verify:get-etherscan-endpoint');
-  const etherscanConfig: EtherscanConfig = (hre.config as any).etherscan;
-  const key = resolveEtherscanApiKey(etherscanConfig.apiKey, endpoints.network);
-  return { key, endpoints };
+  const endpoints = await getCurrentChainConfig(hre.network, (hre.config as any).etherscan.customChains);
+  const key = resolveEtherscanApiKey((hre.config as any).etherscan.apiKey, endpoints.network);
```
frangio, line 48: Is the `etherscan` field guaranteed to be there?

```suggestion
  const endpoints = await getCurrentChainConfig(hre.network, (hre.config as any).etherscan?.customChains);
  const key = resolveEtherscanApiKey((hre.config as any).etherscan?.apiKey, endpoints.network);
```

packages/plugin-hardhat/src/utils/hardhat-etherscan/undici.ts
```diff
@@ -0,0 +1,33 @@
+import type * as Undici from 'undici';
```
frangio, line 1: We need to add this dependency explicitly.

packages/plugin-hardhat/src/utils/hardhat-etherscan/resolveEtherscanApiKey.ts
```diff
@@ -0,0 +1,36 @@
+import { UpgradesError } from '@openzeppelin/upgrades-core';
```
frangio, line 1: I assume other than the error type everything else in `utils/hardhat-etherscan` was vendored from `@nomiclabs/hardhat-etherscan`?

Let's put the [license file](https://github.com/NomicFoundation/hardhat/blob/%40nomiclabs/hardhat-etherscan%403.1.7/packages/hardhat-etherscan/LICENSE) in the directory.

packages/plugin-hardhat/src/utils/etherscan-api.ts
ericglau: It should be guaranteed because it is defined in `hardhat-verify`. But since `hardhat-verify` is an optional peer dependency, I think it is safer to explicitly check this for undefined.

frangio (approved):

PR #831, opened 2023-07-04
Replace ASTDereferencer with one from solidity-ast

frangio:
ASTDereferencer was moved to solidity-ast [a while ago](https://github.com/OpenZeppelin/solidity-ast/releases/tag/v0.4.26) so we can use it from there. It also contains some optimizations since https://github.com/OpenZeppelin/solidity-ast/commit/835dc06543df0bad9dc24a59a0486f9bab4bbccf.

This should unblock https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/783.


frangio:
@ericglau I went ahead and merged this because it's very straightforward.

Issue #767, opened 2023-04-04
Deployed contract address is overriden


andreas-papageorgiou:
It is unclear why the `deploy` function overrides the deployed contract address.

This prevents the usage of `hardhat-upgrades` with other frameworks/chains where contract addresses are predicted differently.

**Relevant code:**
https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/0ef9efb7447f29a46dc26ffe04a852ba2775d30e/packages/plugin-hardhat/src/utils/deploy.ts#L14-L25

ericglau:
This was done in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/487 to handle incorrect addresses returned from certain RPC endpoints on chains such as Polygon Mumbai.

Can you provide examples of frameworks/chains that you are using where this should be different?

andreas-papageorgiou:
We are working on integrating with VeChain

ericglau:
This override has been removed in [@openzeppelin/hardhat-upgrades@2.0.0](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fhardhat-upgrades%402.0.0) since it no longer seems necessary. 
 Note however that the plugin uses ethers.js for deployments, and ethers.js does [its own calculation of the contract deployment address](https://github.com/ethers-io/ethers.js/blob/3c17cf56b5164236108269ac1e36d66e2843cd1e/src.ts/contract/factory.ts#L112).

PR #832, opened 2023-07-04
Remove address override

ericglau:
Closes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/767

Remove overriding the contract address on deployments, since ethers.js's `deploy` function already calculates the address for the create operation.

### DIFF
--- a/packages/plugin-hardhat/CHANGELOG.md
+++ b/packages/plugin-hardhat/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Use `hardhat-verify` for proxy verification. ([#829](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/829))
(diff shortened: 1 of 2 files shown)


frangio (commented):
We used to do something slightly different. I can't see what the difference would be though.

https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/d666753894c7ce7f8d5988c5b10b683745cd7315/packages/plugin-hardhat/src/utils/deploy.ts#L12

frangio (approved):

ericglau:
> We used to do something slightly different. I can't see what the difference would be though.
> 
> https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/d666753894c7ce7f8d5988c5b10b683745cd7315/packages/plugin-hardhat/src/utils/deploy.ts#L12

ethers v6 needs a different approach:
- `deployTransaction` can be null in the returned type (but should not be null here since it was a deployment)
- `address` is not a field so we need to await `getAddress()`

Issue #823, opened 2023-06-20
The upgradeProxy fails with "Bytecode is not a valid hex string"


nikitaVagin95:
Hi OZ! 

I'm trying to update the proxy to the new implementation, which differs from the old one in only one function that returns the contract name + v2 to make sure everything is ok but I get the error. if I change the return string in the contract or change the compile version from ^0.8.18 to ^0.8.20, there is no error. 
I think the problem is in the regular expression in the hashBytecode in version.ts because the bytecode that causes the error still has a library placeholder after the replace func. 

here's the bytecode before and after the replace 
**before:**
[...]

PR #834, opened 2023-07-04
Trim bytecode whitespace when hashing

frangio:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/823

Note: This is backwards compatible because any whitespace that is present would cause the hex check to fail, rather than change the resulting hash.


frangio:
Hm, I'm having doubts that this is the issue behind Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/823. There should never be whitespace here.

PR #838, opened 2023-07-05
Allow assignment of immutable variables

frangio:
Assignment to immutable variables currently requires both overrides `state-variable-immutable` and `state-variable-assignment`. I believe the second one should be implicit. Assignment is a problem when it's in storage.

### DIFF
--- a/packages/core/src/validate/report.ts
+++ b/packages/core/src/validate/report.ts
@@ -35,8 +35,10 @@ const errorInfo: ErrorDescriptions<ValidationError> = {
     link: 'https://zpl.in/upgrades/error-004',
   },
   'state-variable-immutable': {
-    msg: e => `Variable \`${e.name}\` is immutable`,
-    hint: () => `Use a constant or mutable variable instead`,
+    msg: e => `Variable \`${e.name}\` is immutable and will be initialized on the implementation`,
+    hint: () =>
+      `If by design, annotate with '@custom:oz-upgrades-unsafe-allow state-variable-immutable'\n` +
+      `Otherwise, use a mutable variable instead`,
     link: 'https://zpl.in/upgrades/error-005',
   },
   'external-library-linking': {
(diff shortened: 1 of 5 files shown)


packages/core/src/validate/run.ts
```diff
@@ -386,19 +386,19 @@ function* getStateVariableErrors(
 ): Generator<ValidationErrorWithName> {
   for (const varDecl of contractDef.nodes) {
     if (isNodeType('VariableDeclaration', varDecl)) {
```
frangio, line 388: This looks odd but I just switch the two errors and chained them with `else if`.

packages/core/src/validate/report.ts
```diff
@@ -35,8 +35,10 @@ const errorInfo: ErrorDescriptions<ValidationError> = {
     link: 'https://zpl.in/upgrades/error-004',
   },
   'state-variable-immutable': {
-    msg: e => `Variable \`${e.name}\` is immutable`,
-    hint: () => `Use a constant or mutable variable instead`,
+    msg: e => `Variable \`${e.name}\` is immutable and will be initialized on the implementation`,
+    hint: () =>
+      `If by design, annotate with '@custom:oz-upgrades-unsafe-allow state-variable-immutable'\n` +
+      `Otherwise, use a mutable variable instead`,
```
ericglau, line 41: Why is "use a constant" no longer one of the suggestions?

ericglau (approved):

ericglau:
Can you also update the changelog in core?

packages/core/src/validate/report.ts
frangio: No strong reason, I've added it back.

ericglau (approved):

frangio:
@ericglau Let's merge this and release this when you can. I want to make the equivalent change in the transpiler (override flags are used to control transpilation).

ericglau:
@frangio Released in [@openzeppelin/upgrades-core@1.27.2](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fupgrades-core%401.27.2)

PR #844, opened 2023-07-12
Support user-defined value types in mappings

Xanewok:
This fixes the upgraded contract storage check failure when using a custom value types in mappings, e.g.
```
mapping(UserDefinedValueType => ...) mappingThatFails;
```

See https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/779#issuecomment-1527902368

I can confirm that this fixes the upgrade for a contract that uses such mapping (I'm using `upgradeProxy` from hardhat-upgrades).

cc @frangio who suggested the fix

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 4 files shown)


ericglau (approved):
Thanks!

PR #852, opened 2023-07-26
Add `txOverrides` option for overriding transaction parameters

### LINKED ISSUE #85: Override transaction options (from, gas, gasPrice) in create and upgrade proxy

passabilities:
* allows tx overrides object to be passed into the options parameter to use when submitting a tx

Fixes #85


ericglau:
Thanks for this PR!  This looks like a reasonable approach which would address https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/85.  Particularly, since `deploy.ts` in both Hardhat and Truffle are used for all deployments (including proxies, proxy-related contracts and implementations), setting the overrides there would cover the various deployment transactions.

I’ve made some changes to separate the option type for Hardhat and Truffle, since Truffle does not use ethers.js.
Also removed the .idea folder since that is environment-specific and would be more suitable in a local configuration.

There are a few more changes that I am still planning to make:
- The Truffle option still needs a more specific type.
- Any transactions other than deployments (for example, upgrading a proxy) also need to respect the new options.
- Add tests.
- Add documentation.
- Add changelog entry.

packages/plugin-hardhat/src/upgrade-beacon.ts
```diff
@@ -27,7 +27,9 @@ export function makeUpgradeBeacon(hre: HardhatRuntimeEnvironment, platformModule
 
     const UpgradeableBeaconFactory = await getUpgradeableBeaconFactory(hre, getSigner(ImplFactory.runner));
     const beaconContract = attach(UpgradeableBeaconFactory, beaconAddress);
-    const upgradeTx = await beaconContract.upgradeTo(nextImpl);
+
+    const overrides = opts.txOverrides ? [opts.txOverrides] : [];
+    const upgradeTx = await beaconContract.upgradeTo(nextImpl, ...overrides);
```
frangio, line 32: What happens if you just write `upgradeTo(nextImpl, overrides)`?

ericglau: That would cause an error because then `overrides` appears as an array and Ethers tries to use it as part of the actual function arguments.

We need to spread the array because its element `opts.txOverrides` is the actual overrides object, and if that is not present, then it will be excluded.

frangio: Sorry my question was actually if we write `upgradeTo(nextImpl, opts.txOverrides)` :slightly_smiling_face:

ericglau: That doesn't work, but `upgradeTo(nextImpl, opts.txOverrides ?? {})` works.  Let's use this.

frangio: I'm not sure that these options will work with Truffle, I recall it working weird if you pass an empty object.

frangio (approved):

packages/plugin-hardhat/src/upgrade-beacon.ts
ericglau: Right, these don't work for Truffle.  Will leave them as-is for both Hardhat and Truffle, for consistency.

PR #856, opened 2023-08-01
CLI: Support `contract` and `reference` options

### LINKED ISSUE #855: CLI: Provide a way to specify specific contract and reference
The current upgrades-core CLI validates the entire build-info directory.  However, callers may want to check upgrade safety for a specific contract, and optionally specify which reference contract to use for the storage layout comparison, without having to annotate contracts.  For example, this would be useful for pre-deployment or pre-upgrade checks where the caller knows the specific contracts they are interested in.

Provide a way to specify a single contract and optional "reference" contract, with the following additional differences:
- Should not require `@custom:oz-upgrades-from` or `@custom:oz-upgrades` annotations.  The presence of the new options imply that the user wants to perform a check for a specific contract or pair of contracts.

ericglau:
Fixes #855


packages/core/src/cli/cli.test.ts
frangio, line 1: There is no test for passing in a fully qualified name.

packages/core/src/cli/validate.ts
```diff
@@ -76,8 +85,13 @@ export function getFunctionArgs(parsedArgs: minimist.ParsedArgs, extraArgs: stri
     throw new Error('The validate command takes only one argument: the build info directory.');
   } else {
     const buildInfoDir = extraArgs.length === 1 ? undefined : extraArgs[1];
+    const contract = parsedArgs['contract'];
+    const reference = parsedArgs['reference'];
+    if (reference !== undefined && contract === undefined) {
```
frangio, line 90: It's possible to pass `--contract --reference Foo` and it results in `contract: '', reference: 'Foo'`. This sounds like something we would want to catch in this function.

Just `--contract --reference` is also probably valid.

packages/core/src/cli/validate/upgradeability-assessment.ts
```diff
@@ -35,49 +18,34 @@ export interface UpgradeabilityAssessment {
...
-      referenceContract: referenceContract,
-      uups: isReferenceUUPS || isUUPS,
-    };
-  } else {
-    const initializable = inferInitializable(contractValidation);
-    return {
-      upgradeable: initializable || isUUPS,
-      uups: isUUPS,
-    };
+  let referenceContract = overrideReferenceContract ?? undefined;
```
frangio, line 30: I don't see how this could be necessary?

```suggestion
  let referenceContract = overrideReferenceContract;
```

frangio (approved):

PR #859, opened 2023-08-09
[AY]: Allow override of proxy admin

arijoon:
In some instances a single proxy admin per network does not suffice but implementation addresses can be shared.
Due to this issue we must be able to override the admin

For example in our specific usecase, we have difference accounts for difference environments, i.e. multiple admin addresses per each network, this requires multiple proxy admins per each network however the current setup only allows a single admin. The proposed solution is not the most optimal but allows the most flexibility as the applications can simply override the admin to their desired value, it would be best if this value can be passed down the function calls but that would result in a lot of changes and I was not sure if it would be desired by the maintainers


ericglau:
Thank you @arijoon for this PR.  The plugins do not currently make use of environment variables, so this approach would deviate from the current design and user experience.

Also note that in the upcoming OpenZeppelin Contracts 5.0, transparent proxies always deploy a new immutable proxy admin (see https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/transparent/TransparentUpgradeableProxy.sol)

Due to the above, I suggest the following instead:

1. Change the manifest file's `admin` section to also allow storing an array of admin addresses. e.g. in manifest.ts, the interface of `ManifestData` could have `admin?: Deployment | Deployment[]`;

2. Enable importing an existing proxy admin, which will append its address to the `admin` array.  This could be done in a new `importProxyAdmin` function.  This should validate whether the given address has bytecode.

3. It would also need to update the scenario of importing a transparent proxy from the `forceImport` function, during which it should add the associated proxy admin to the manifest as well.  Currently it would give an error if the imported transparent proxy's admin does not match what is in the manifest.

Please let me know if this satisfies your requirements, and whether you would be interested in doing these suggested changes. Thanks.

arijoon:
> 
> Please let me know if this satisfies your requirements, and whether you would be interested in doing these suggested changes. Thanks.

Thanks for a swift review. The proposed changes could work, my only worry is the reliance on `getAdmin` which is expecting a single admin address. Any flow that uses this address would not function correctly when there is a list of admins. Do you think this would be an issue or it would be safe to completely eliminate the reliance on a single admin? (maybe I should add an admin argument to the deployment options as well? basically we already deploy as you mentioned like v5.0 by always creating a new proxy admin for each transparentProxy due to our authorization needs)
In an ideal scenario there''ll be a mapping between the proxy and its admin, so there'll no longer be a getAdmin but getAdmin(address proxy). I can also add it this way but again it would deviate from current flows (maybe 2 flows again by deployment options, if mapping exists for proxy use it, if not rely on the global admin set, for new deployments use a new option argument to decide whether a new proxy admin should be deployed or existing one reused)

On a side note is there an objection to env variables overall or just in this manner as all envs vars are basically inputs to the program

ericglau:
Your assessment makes sense.  When working with an existing proxy, we could even ignore the admin from the manifest, and just read the proxy's admin from the ERC1967 admin slot.  Then `getAdmin()` would only be used when deploying new proxies.

So the proposal would be changed to the following:

1. Keep only a single admin in the manifest's `admin` field (as it currently is).  This will be used by default when deploying transparent proxies.

2. When modifying an existing proxy through `upgradeProxy`, `admin.changeProxyAdmin`, or `admin.transferProxyAdminOwnership`, use [getAdminAddress](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/7eea0d375813722ba91226934e968130bcb84ffa/packages/core/src/eip-1967.ts#L9) to get the specific proxy's admin instead of `getAdmin()`.

3. For the `forceImport` function when importing a transparent proxy, change the error to a warning if the imported transparent proxy's admin does not match what is in the manifest.

If you like, you can also add an option when deploying a transparent proxy to always deploy a new admin.  However, this option would be removed when we migrate the plugin to use proxies from v5.0, because it will always deploy a new admin.  So it's up to you whether you think it is worth the effort in the meantime.

> On a side note is there an objection to env variables overall or just in this manner as all envs vars are basically inputs to the program

It's mainly for this situation because I see environment variables as something that work best when set for the entire program.  But if a script is using multiple transparent proxies with multiple admins, the environment variable would need to be changed multiple times.

arijoon:
Thanks @ericglau , this was easier than I had anticipated. I dropped the change for proxy address in deployOpts if it's no longer needed in v5 anyways since we already have a setup that deletes this property before new deployments to force a new admin. Let me know if more tests are needed, I've modified the existing ones to match the current behavior

packages/plugin-hardhat/src/admin.ts
```diff
@@ -36,8 +36,12 @@ export function makeChangeProxyAdmin(hre: HardhatRuntimeEnvironment, platformMod
...
-      throw new Error('Proxy admin is not the one registered in the network manifest');
-    } else if (manifestAdminAddress !== newAdmin) {
+      console.log(
+        `Proxy admin ${proxyAdminAddress} is not the one registered in the network manifest ${manifestAdminAddress}`,
+      );
+    }
+
+    if (proxyAdminAddress !== newAdmin) {
       const overrides = opts.txOverrides ? [opts.txOverrides] : [];
       await admin.changeProxyAdmin(proxyAddress, newAdmin, ...overrides);
```
ericglau, line 46: This should no longer use the admin from the manifest, but should use the proxy's current admin.  For example, you can remove lines 34-35, then change this to:
```suggestion
      const AdminFactory = await getProxyAdminFactory(hre, signer);
      const admin = attach(AdminFactory, proxyAdminAddress); 
      await admin.changeProxyAdmin(proxyAddress, newAdmin, ...overrides);
```

packages/plugin-hardhat/src/admin.ts
```diff
@@ -36,8 +36,12 @@ export function makeChangeProxyAdmin(hre: HardhatRuntimeEnvironment, platformMod
     const proxyAdminAddress = await getAdminAddress(hre.network.provider, proxyAddress);
 
     if (manifestAdminAddress !== proxyAdminAddress) {
-      throw new Error('Proxy admin is not the one registered in the network manifest');
-    } else if (manifestAdminAddress !== newAdmin) {
+      console.log(
+        `Proxy admin ${proxyAdminAddress} is not the one registered in the network manifest ${manifestAdminAddress}`,
+      );
+    }
```
ericglau, line 42: This message can just be deleted, since the manifest's admin is not really relevant anymore for this scenario.

packages/plugin-hardhat/test/transparent-admin-validation.js
```diff
@@ -12,10 +13,16 @@ test.before(async t => {
...
+  const AdminFactory = await getProxyAdminFactory(hre, signer);
+  const deployedAdmin = await AdminFactory.deploy();
+  const deployedAdminAddress = await deployedAdmin.getAddress();
+  await upgrades.admin.changeProxyAdmin(await greeter.getAddress(), deployedAdminAddress);
+
+  // Old admin signer cannot upgrade this
+  await t.throwsAsync(() => upgrades.upgradeProxy(greeter, GreeterV2));
+
+  const GreeterV3 = Greeter.connect(signer);
+  await upgrades.upgradeProxy(greeter, GreeterV3);
```
ericglau, line 27: Can you add a test for running `changeProxyAdmin` again?  For example, add this here:

```
  // New admin signer can change the admin again
  const deployedAdmin2 = await AdminFactory.deploy();
  const deployedAdminAddress2 = await deployedAdmin2.getAddress();
  await upgrades.admin.changeProxyAdmin(await greeter.getAddress(), deployedAdminAddress2, signer);
  ```

ericglau (commented):
Thanks a lot for these changes.  Just a few comments.

Also, this currently only covers Hardhat.  Would you be able to make the same changes for Truffle?

arijoon:
Applied your suggestions and added the same changes to truffle

ericglau (approved):
Thank you!

PR #876, opened 2023-09-07
Add validations for namespaced storage layout

ericglau:
Adds upgrade safety validations for implementations that use [ERC-7201: Namespaced Storage Layout](https://eips.ethereum.org/EIPS/eip-7201).


frangio (commented):
Still reviewing but I want to leave my comments so far.

packages/plugin-hardhat/ava.config.js
```diff
@@ -4,4 +4,5 @@ module.exports = {
   ignoredByWatcher: ['**/*.ts', '.openzeppelin'],
   serial: true,
   failWithoutAssertions: false,
+  snapshotDir: '.',
```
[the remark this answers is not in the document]
ericglau: This causes the snapshots to be placed flat in the same directory as the tests (in the `test` directory).  Without this, the snapshots would go into a `snapshots` subdirectory (`test/snapshots`).

This was done to be consistent with `core` where the snapshots were placed alongside the tests.  The reason this behaves differently by default is due to https://github.com/avajs/ava/discussions/2765#discussioncomment-849837

Although I would be fine with reverting to ava's default behaviour as well.

packages/core/src/validate/run.ts
```diff
@@ -195,6 +224,71 @@ export function validate(solcOutput: SolcOutput, decodeSrc: SrcDecoder, solcVers
...
+    }
+  }
+}
+
+function getNamespacedCompilationContext(
+  source: string,
+  contractDef: ContractDefinition,
+  namespacedOutput?: SolcOutput,
+) {
+  if (namespacedOutput === undefined || contractDef.canonicalName === undefined) {
```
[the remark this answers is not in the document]
ericglau: For `UserDefinedValueTypeDefinition`, canonical name includes the contract name e.g. `<contract>.<uvdt>`, but I'm not sure what's the difference (if any) for `ContractDefinition`

However, this code is reachable even if users are on < 0.8.9 and not using namespaces.  In this case, `canonicalName` is undefined for all contracts so we should effectively skip (by returning undefined as currently written).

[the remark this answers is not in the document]
frangio: My guess is that canonicalName is the "scoped" name, but since contracts can't be nested they will always be defined at the root and `name` will be equal to `canonicalName`. It's nice not to encode the assumption of not being nested so it's good to use `canonicalName`.

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,396 @@
...
+ * functions and constructors.
+ *
+ * Also sets the outputSelection to only include storageLayout and ast, since the other outputs are not needed.
+ *
+ * @param input The original solc input.
+ * @param output The original solc output.
+ * @returns The modified solc input with storage layout that includes namespaced type information.
+ */
+export function makeNamespacedInputCopy(input: SolcInput, output: SolcOutput): SolcInput {
+  const modifiedInput: SolcInput = JSON.parse(JSON.stringify(input));
```
frangio, line 323: `JSON.parse(JSON.stringify(...))` is a quick hack that works, but for this use case I'm pretty sure it will be particularly inefficient. The intuition is that there are a lot of strings in this object (all of the Solidity source code) that stringify will copy into a new string before they are parsed again. If we used some procedure that copied the object in memory, the strings can just be copied by reference because JavaScript strings are immutable, so it should be a lot cheaper.

The native way to do a deep copy is `structuredClone`. It's not available in Node 16, which we still support, but we can use a polyfill: https://github.com/ungap/structured-clone

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,396 @@
...
+    const source = modifiedInput.sources[sourcePath];
+    if (source.content === undefined) {
+      continue;
+    }
+
+    // Collect all contract definitions
+    const contractDefs = [];
+    for (const contractDef of findAll('ContractDefinition', output.sources[sourcePath].ast)) {
+      contractDefs.push(contractDef);
+    }
```
frangio, line 344:
```suggestion
    const contractDefs = [...findAll('ContractDefinition', output.sources[sourcePath].ast)];
```

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,396 @@
...
+      const contractDef = contractDefs[i];
+      const nodes = contractDef.nodes;
+      for (let j = nodes.length - 1; j >= 0; j--) {
+        const node = nodes[j];
+        if (
+          isNodeType('FunctionDefinition', node) ||
+          isNodeType('ModifierDefinition', node) ||
+          (isNodeType('VariableDeclaration', node) && node.mutability === 'immutable')
+        ) {
+          const orig = Buffer.from(source.content);
```
frangio, line 357: Each of these operations copies the string contents, so let's create the buffer outside of the loop and store it in the modified input object after the loop.

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,396 @@
...
+    // Iterate backwards so we can delete source code without affecting remaining indices
+    for (let i = contractDefs.length - 1; i >= 0; i--) {
+      const contractDef = contractDefs[i];
+      const nodes = contractDef.nodes;
+      for (let j = nodes.length - 1; j >= 0; j--) {
+        const node = nodes[j];
+        if (
+          isNodeType('FunctionDefinition', node) ||
+          isNodeType('ModifierDefinition', node) ||
+          (isNodeType('VariableDeclaration', node) && node.mutability === 'immutable')
```
frangio, line 355: I'm realizing this is more complex than I said initially. You probably ran into this. If we delete functions, we can no longer have references to those functions, so we have to delete every possible place where a reference would show up, i.e. every possible expression. I would delete all `VariableDeclaration` nodes (ignoring the mutability), because they can contain an initialization expression. The other thing that isn't covered by your code here is initialization in the parents list:

```solidity
contract A is B(expr()) {
}
```

This is in `contractDef.baseContracts[i].arguments`, but this is an array and we don't have bounds that include parenthesis. What I would do is take each one of `contractDef.baseContracts[i]` and replace all text within the node bounds with the value of `contractDef.baseContracts[i].baseName`.

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,396 @@
...
+          const [start, length] = node.src.split(':').map(Number);
+          let end = start + length;
+
+          // If the next character is a semicolon (e.g. for immutable variables), delete it too
+          if (end + 1 < orig.length && orig.subarray(end, end + 1).toString() === ';') {
+            end += 1;
+          }
+
+          // Delete the source code segment
+          const buf = Buffer.concat([orig.subarray(0, start), orig.subarray(end)]);
```
frangio, line 368: You had originally suggested to fill with spaces. It may be more efficient and result in less memory copying.

[`buf.fill`](https://nodejs.org/api/buffer.html#buffillvalue-offset-end-encoding)

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,396 @@
...
+
+          source.content = buf.toString();
+        } else if (isNodeType('StructDefinition', node)) {
```
frangio, line 371: I see that this part is going to break if we do what I said above about filling with spaces. What we could do is to build a contract like `contract ___ is ${contractDef.name} { ${structName} ${variableName}; }` and append it at the end of the file outside the for loop...

Open to other ideas. To be honest, we should optimize prematurely, but avoiding copying the file contents over and over again seems like a no brainer.

frangio: Ok so this idea of appending a new contract at the end of the file is probably not going to work because our goal is to add the namespace struct into the types of the `storageLayout` output for the original contract. So we have to somehow add the variable inside the original contract.

An alternative to copying the contents over and over again is to only do it at the end when we convert the source back to a string. We can construct an array of... something and then at the end concatenate it all into a string.

packages/core/src/storage/namespace.ts
ericglau: Changed it to record modifications including original positions, then perform a copy at the end.  This way there is no need to fill with spaces or iterate backwards.

packages/core/src/storage/namespace.ts
ericglau: Changed to copy parts of the solc input by reference, and construct the resulting solc input object using those references and the generated portions.

packages/core/src/utils/make-namespaced.ts
```diff
@@ -0,0 +1,133 @@
...
+          const doc = node.documentation;
+          if (doc) {
+            modifications.push(getDelete(doc, orig));
+          }
+          modifications.push(getDelete(node, orig));
+        } else if (isNodeType('StructDefinition', node)) {
+          const storageLocationArg = getStorageLocationArg(node);
+          if (storageLocationArg !== undefined) {
+            const structName = node.name;
+            const variableName = `$${structName}`;
```
frangio, line 67: Do we rely on the variable name at any point? This one seems too easy to clash. I'm also wondering how this whole process would interact with `hardhat-exposed`, but it might interfere because it uses this naming convention too.

I would append some random numbers at the end to ensure there are no clashes.

```suggestion
            const variableName = `$${structName}_${(Math.random() * 1e6).toFixed(0)}`;
```

packages/core/src/utils/make-namespaced.ts
```diff
@@ -44,7 +44,16 @@ export function makeNamespacedInput(input: SolcInput, output: SolcOutput): SolcI
...
+          const baseName = inherit.baseName;
+
+          let name;
+          if (isNodeType('UserDefinedTypeName', baseName)) {
+            name = (baseName.name ?? baseName.pathNode?.name)!;
+          } else {
+            name = baseName.name;
+          }
+
+          modifications.push(getReplace(inherit, name));
```
frangio, line 56: This can be a lot simpler:
```suggestion
          assert(inherit.baseName.name !== undefined);
          modifications.push(getReplace(inherit, inherit.baseName.name));
```

See https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/876/files/db69587e61da520fb1949c86950bd029f0cdc17e#r1332135132 if you haven't seen it yet.

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,305 @@
...
+  origContractDef: ContractDefinition,
+) {
+  for (const node of context.contractDef.nodes) {
+    if (isNodeType('StructDefinition', node)) {
+      const storageLocationArg = getStorageLocationArg(node);
+      if (storageLocationArg !== undefined) {
+        const origSrc = decodeSrc(getOriginalStruct(node.canonicalName, origContractDef));
+
+        if (namespaces[storageLocationArg] !== undefined) {
+          namespaces[storageLocationArg].srcs.push(origSrc);
```
frangio, line 146: Is there a reason why we don't just throw an error at this point, and instead do that later at `checkAndSaveLayout`?

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,305 @@
...
+}
+
+function getStorageLocation(doc: string) {
+  const storageLocationArgs = getAnnotationArgs(doc, 'storage-location');
+  if (storageLocationArgs.length !== 1) {
+    throw new Error('@custom:storage-location annotation must have exactly one argument');
+  }
+  const storageLocation = storageLocationArgs[0];
+  return storageLocation;
+}
```
frangio, line 180: I would merge these two functions.

Additionally, I think I would directly call `getAnnotationArgs` without first checking `hasAnnotationTag`? If it returns an empty array, there is no annotation, and if it returns more than 1 element we throw the error.

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,305 @@
...
+ * Gets the storage items for the given struct node.
+ * Includes loading recursive type information, and adds slot and offset if they are available in the given compilation context's layout.
+ */
+function getNamespacedStorageItems(
+  node: StructDefinition,
+  decodeSrc: SrcDecoder,
+  layout: StorageLayout,
+  context: CompilationContext,
+  origContractDef: ContractDefinition,
+) {
```
frangio, line 192: It's not strictly necessary but for these complex functions especially I would declare the return type explicitly:
```suggestion
): StorageItem[] {
```

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,305 @@
...
+  }
+  const storageLocation = storageLocationArgs[0];
+  return storageLocation;
+}
+
+/**
+ * Gets the storage items for the given struct node.
+ * Includes loading recursive type information, and adds slot and offset if they are available in the given compilation context's layout.
+ */
+function getNamespacedStorageItems(
```
frangio, line 186: I want to discuss this function. I think I expected this to be less complex because StructMember and StorageItem are very similar types so my intuition was that if we first build the layout we should get the list of items "for free".

packages/plugin-hardhat/src/index.ts
```diff
@@ -110,13 +113,45 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE, async (args: RunCompilerArgs, hre, runSup
   const { isFullSolcOutput } = await import('./utils/is-full-solc-output');
   if (isFullSolcOutput(output)) {
     const decodeSrc = solcInputOutputDecoder(args.input, output);
-    const validations = validate(output, decodeSrc, args.solcVersion);
+
+    const namespacedInput = makeNamespacedInput(args.input, output);
+    const { output: namespacedOutput } = await runSuper({ ...args, input: namespacedInput });
```
frangio, line 118:
```suggestion
    const { output: namespacedOutput } = await runSuper({ ...args, quiet: true, input: namespacedInput });
```


?

packages/plugin-hardhat/src/index.ts
```diff
@@ -110,13 +113,45 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE, async (args: RunCompilerArgs, hre, runSup
...
     await writeValidations(hre, validations);
   }
 
   return { output, solcBuild };
 });
 
+/**
+ * Checks for compile errors in the modified contracts for namespaced storage.
+ * If errors are found, throws an error with the compile error messages, and logs
+ * the modified contract source code as debug.
```
frangio, line 131: I wonder if we should log the input source code instead, it would probably give us more information to debug, and we can recover the namespaced input from it.

packages/plugin-hardhat/src/index.ts
```diff
@@ -110,13 +113,45 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE, async (args: RunCompilerArgs, hre, runSup
...
+  const errors = [];
+  if (namespacedOutput.errors !== undefined) {
+    for (const error of namespacedOutput.errors) {
+      if (error.severity === 'error') {
+        const msg = error.formattedMessage;
+
+        debug('Compile error in modified contract for namespaced storage.');
+        debug('Error:', msg);
+        if (error.sourceLocation !== undefined && error.sourceLocation.file in namespacedInput.sources) {
+          debug('Modified contract source:', namespacedInput.sources[error.sourceLocation.file].content);
```
frangio, line 143: This will be better I think:

```suggestion
          debug('Modified contract source:', JSON.stringify(namespacedInput.sources[error.sourceLocation.file].content));
```

packages/plugin-hardhat/src/index.ts
```diff
@@ -35,6 +37,7 @@ import {
   GetVerifyDeployBuildInfoFunction,
   GetBytecodeDigestFunction,
 } from './defender-v1/verify-deployment';
+import debug from './utils/debug';
```
frangio, line 40: Let's add this import inside the function where it's used:

```ts
const debug = await import('./utils/debug');
```

Originally it was a goal to not have any imports at the top level in this file (other than type imports). I think at some point this wasn't followed properly. We shouldn't have an import of `@upgrades/core` at the top level, and there are some of these import statements that should be written as `import type`.

See https://hardhat.org/hardhat-runner/docs/advanced/building-plugins#optimizing-your-plugin-for-better-startup-time

packages/plugin-hardhat/src/index.ts
ericglau: The purpose of the debug logging statements here are to allow us to ask the user to provide this info if they are comfortable with it.  If we want the original source code, we could ask for that instead.  In that case, we could even remove the debug messages from this function since the error itself would be thrown anyways.  What do you think?

packages/core/src/storage/namespace.ts
ericglau: Doing it later allows us to collect all of the conflicts and report them all at once (in case there are more than 2 of the same namespace)

packages/plugin-hardhat/src/index.ts
ericglau: Removed debug messages as mentioned in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/876#discussion_r1333201777

packages/core/src/storage/namespace.ts
ericglau: The challenge is that `getAnnotationArgs` would return an emtpy array if the annotation does not exist at all, or has just  `@custom:storage-location` with no argument.  Ideally we want to throw an error in the second case.

packages/plugin-hardhat/src/index.ts
ericglau: Removed top level imports

packages/core/src/storage/namespace.ts
frangio: Got it! Makes sense.

I would still merge them though.

packages/core/src/storage/namespace.ts
ericglau: We need to lookup the struct members from both the AST (to get `typeName`) and storage layout types (to get `offset` and `slot`).  Kept the same approach as before but refactored the code to be more readable.

packages/core/src/utils/make-namespaced.ts
```diff
@@ -0,0 +1,152 @@
...
+        modifications.push(makeReplace(node, orig, insertText));
+      } else if (isNodeType('UsingForDirective', node)) {
+        modifications.push(makeDelete(node, orig));
+      }
+    }
+
+    modifiedSources[sourcePath] = { ...source, content: getModifiedSource(orig, modifications) };
+  }
+
+  return { ...input, sources: modifiedSources, settings: { ...input.settings, outputSelection: OUTPUT_SELECTION } };
```
frangio, line 97: I'm thinking we might want to disable optimizations in the settings too. Since the output selection doesn't include any bytecode, I'm not sure if it will have any effect though. I'm checking with the Solidity team to confirm.

frangio: Update: No need to disable optimizations, it should apparently not run unless we request codegen outputs.

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,280 @@
...
+      item = {
+        ...item,
+        offset: layoutMember.offset,
+        slot: layoutMember.slot,
+      };
+    }
+
+    storageItems.push(item);
+
+    loadLayoutType(astMember.typeName, layout, context.deref);
```
frangio, line 213: Can we add a comment here explaining the role of this statement?

If I understand correctly:
- If context is namespaced, we have storage layout, and this will fill in enum members just like in `extractStorageLayout`.
- If context is original, this will add the namespace structs to the layout.

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,280 @@
...
+  const structType = findTypeWithLabel(types, `struct ${structCanonicalName}`);
+  const structMembers = structType?.members;
+  if (structMembers !== undefined) {
+    for (const structMember of structMembers) {
+      assert(typeof structMember !== 'string');
+      if (structMember.label === memberLabel) {
+        return structMember;
+      }
+    }
+    return undefined;
```
frangio, line 271:
```suggestion
```

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,280 @@
...
+function findLayoutStructMember(
+  types: Record<string, TypeItem<string>>,
+  structCanonicalName: string,
+  memberLabel: string,
+) {
+  const structType = findTypeWithLabel(types, `struct ${structCanonicalName}`);
+  const structMembers = structType?.members;
+  if (structMembers !== undefined) {
+    for (const structMember of structMembers) {
+      assert(typeof structMember !== 'string');
```
frangio, line 266: Needs import `isStructMembers`.

```suggestion
    assert(isStructMembers(structMembers));
    for (const structMember of structMembers) {
```

frangio (commented):
Sorry these are my final comments. All relatively minor I think.

packages/core/src/utils/make-namespaced.ts
```diff
@@ -0,0 +1,152 @@
...
+            if (storageLocation !== undefined) {
+              const structName = contractNode.name;
+              const variableName = `$${structName}_${(Math.random() * 1e6).toFixed(0)}`;
+              const insertText = ` ${structName} ${variableName};`;
+
+              modifications.push(makeInsertAfter(contractNode, insertText));
+            }
+          }
+        }
+      } else if (isNodeType('FunctionDefinition', node) || isNodeType('VariableDeclaration', node)) {
```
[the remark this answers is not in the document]
ericglau: We could delete the documentation, but why is it necessary?  These nodes are just being replaced by variables, so the documentation would not be orphaned.

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,280 @@
...
+  const astTypeMembers = getTypeMembers(node, { typeName: true });
+
+  const storageItems: StorageItem[] = [];
+
+  for (const astMember of astTypeMembers) {
+    let item: StorageItem = {
+      contract: context.contractDef.name,
+      label: astMember.label,
+      type: astMember.type,
+      src: decodeSrc({ src: getOriginalMemberSrc(node.canonicalName, astMember.label, origContractDef) }),
```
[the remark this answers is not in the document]
ericglau: What would be the benefit of storing the `astId`?  Especially since once the manifest is written, the manifest contents may no longer have any relation to solc output (since the solc output in the user's project may get deleted).

packages/core/src/storage/extract.ts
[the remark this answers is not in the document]
ericglau: Should we always prefer `assert/strict` unless there is a need for non-strictness?

We also have a `core/src/utils/assert.ts` utility.  In what scenarios would this util be preferred vs node.js assertions?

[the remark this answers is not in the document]
frangio: Tbh we should probably use that utility if we think it's a good idea. The thinking was to have a common error message for these unexpected errors.

packages/core/src/utils/make-namespaced.ts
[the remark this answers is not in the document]
frangio: Some NatSpec tags like `@param` will error when applied to a variable declaration.

packages/core/src/storage/namespace.ts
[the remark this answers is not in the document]
frangio: > Especially since once the manifest is written, the manifest contents may no longer have any relation to solc output

This is a good point. No need to store it.

packages/core/src/storage/namespace.ts
```diff
@@ -0,0 +1,280 @@
...
+      node.canonicalName,
+      astMember.label,
+    );
+
+    if (layoutMember?.offset !== undefined && layoutMember?.slot !== undefined) {
+      item = {
+        ...item,
+        offset: layoutMember.offset,
+        slot: layoutMember.slot,
+      };
```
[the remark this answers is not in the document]
frangio: But this is in an `if` block that checks those values aren't undefined. Additionally, when stringified to JSON all undefined values are removed.

packages/core/src/storage/extract.ts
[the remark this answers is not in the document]
frangio: But no need to change that in this PR.

frangio (approved):
LFG!

PR #880, opened 2023-09-19
Support implementations with upgradeTo or upgradeToAndCall

ericglau:
Allow UUPS implementations from OpenZeppelin Contracts 5.0 which only have `upgradeToAndCall`


Amxx (approved):
LGTM

frangio (approved):

Issue #881, opened 2023-09-22
Call `upgradeToAndCall` with no data


Amxx:
Since 5.0.0-rc.0, the UUPS proxy and proxy admin do not expose `upgradeTo` anymore. Instead, they expose `upgradeToAndCall` with support for empty `data`.

Currently the plugin relies on the `upgradeTo` function ... unless a `call` option is specified. There is no easy way to set that call option to empty bytes. A workaround it to use a call to a view function as an initializer, but that is not great.

PR #883, opened 2023-09-22
Use upgradeToAndCall depending on upgrade interface version

ericglau:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/881

This PR enables the use of UUPS implementations and proxy admins from OpenZeppelin Contracts 5.0.  Other 5.0 proxy kinds can be imported as well (added test cases).

- Call `upgradeToAndCall` depending on `UPGRADE_INTERFACE_VERSION`
  - when upgrading with a UUPS implementation
  - when upgrading a transparent proxy using a 5.0 proxy admin (which was imported via `forceImport`)
- Fixes inferring of UUPS proxies when only `upgradeToAndCall` is available
- Add note about `changeProxyAdmin` not being supported for 5.0 admins or proxies


packages/core/src/call-optional-selector.ts
```diff
@@ -0,0 +1,22 @@
+import { keccak256 } from 'ethereumjs-util';
+import { call, EthereumProvider } from './provider';
+
+export async function callOptionalSelector(provider: EthereumProvider, address: string, selector: string) {
```
frangio, line 4: "Selector" is the 4-byte hash. We can call this "signature".

packages/core/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Support getting `UPGRADE_INTERFACE_VERSION`, fix inferring of UUPS proxies with `upgradeToAndCall`. ([883](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/883))
```
frangio, line 5:
```suggestion
- Support new upgrade interface in OpenZeppelin Contracts 5.0. ([883](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/883))
```

frangio (commented):
Looks good, minor comments above.

frangio (approved):

packages/core/src/call-optional-signature.ts
```diff
@@ -0,0 +1,22 @@
...
+        e.message.includes('function selector was not recognized') ||
+        e.message.includes('invalid opcode') ||
+        e.message.includes('revert') ||
+        e.message.includes('execution error')
+      )
+    ) {
+      throw e;
+    } else {
+      return undefined;
+    }
```
Amxx, line 20: I think it would be clearer if we invert the condition.
```suggestion
    if (
      e.message.includes('function selector was not recognized') ||
      e.message.includes('invalid opcode') ||
      e.message.includes('revert') ||
      e.message.includes('execution error')
    ) {
      return undefined;
    } else {
      throw e;
    }
```

packages/core/src/call-optional-signature.ts
```diff
@@ -0,0 +1,22 @@
+import { keccak256 } from 'ethereumjs-util';
+import { call, EthereumProvider } from './provider';
+
+export async function callOptionalSignature(provider: EthereumProvider, address: string, signature: string) {
```
Amxx, line 4: This function is limited to function with no arguments, and the output is not parsed. That is good enough for now, but if we ever want to go further, I believe that ethers has some usefull primitives that are ts friendly.

ericglau: Currently the `upgrades-core` package currently does not pull in ethers as a dependency, we can consider it when there is a need.

PR #885, opened 2023-09-25
Add Truffle deprecation notices

ericglau:
Deprecate Truffle Upgrades due to https://consensys.io/blog/consensys-announces-the-sunset-of-truffle-and-ganache-and-new-hardhat


frangio (commented):
Maybe we should use `WARNING:` and `**Warning**` instead of "note"?

frangio (approved):

Issue #882, opened 2023-09-22
With contracts-upgradeable v5.0.0-rc.0 adminAddress is zero


SKYBITDev3:
When trying contracts-upgradeable v5.0.0-rc.0 and hardhat-upgrades, calling `forceImport` now fails because `adminAddress` is `0x0000000000000000000000000000000000000000`.

frangio:
The Upgrades Plugins haven't yet been updated for 5.0 proxies.

SKYBITDev3:
> The Upgrades Plugins haven't yet been updated for 5.0 proxies.

When will it be updated? I'm eager to try it.

frangio:
This is likely to happen a few weeks after we release 5.0.

ericglau:
@SKYBITDev3 Can you provide more information about what you mean by `adminAddress` is `0x0000000000000000000000000000000000000000`?  Where are you seeing that value being zero and how are you retrieving that value?

In https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/883 we are enabling the import and upgrading of 5.0 proxies, but we did not encounter any issues for the actual import.  If possible, can you share what your scripts look like for deploying and importing the proxies?

SKYBITDev3:
Here is my deployment script and failure location: https://github.com/SKYBITDev3/SKYBIT-Keyless-Deployment/blob/0e8051ba6bffb2a296ed2234f0429b8a5a3c1ba0/scripts/deployViaCREATE3-TESTERC20UG.js#L92

```
InvalidDeployment [Error]: No contract at address 0x0000000000000000000000000000000000000000 (Removed from manifest)
    at waitAndValidateDeployment (C:\Users\-\Documents\Projects\project\node_modules\@openzeppelin\upgrades-core\src\deployment.ts:251:15)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at fetchOrDeployGeneric (C:\Users\-\Documents\Projects\project\node_modules\@openzeppelin\upgrades-core\src\impl-store.ts:88:5)
    at fetchOrDeployAdmin (C:\Users\-\Documents\Projects\project\node_modules\@openzeppelin\upgrades-core\src\impl-store.ts:210:11)
    at simulateDeployAdmin (C:\Users\-\Documents\Projects\project\node_modules\@openzeppelin\hardhat-upgrades\src\utils\simulate-deploy.ts:19:32)
[...]

PR #887, opened 2023-09-27
Check for non-zero admin address when importing transparent proxy

ericglau:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/882

When importing a proxy, the proxy kind can be inferred based on the presence of upgrade function signatures in the implementation.

For transparent proxies, we import the admin address using the admin address slot.  During this import, we check whether the admin address has code.

If the proxy kind was inferred incorrectly, or if the user provided the wrong `kind` option (e.g. passed in 'transparent' but the proxy is actually UUPS), the admin address is the zero address.  This causes the error when we check for code at the admin address:
```
No contract at address 0x0000000000000000000000000000000000000000 (Removed from manifest)
```

Instead, we should throw an error if the admin address is zero, and tell the user to set the correct proxy kind.

NOTE: This assumes that a zero admin address is never a valid scenario for a transparent proxy.

### DIFF
--- a/packages/plugin-hardhat/src/force-import.ts
+++ b/packages/plugin-hardhat/src/force-import.ts
@@ -116,6 +116,7 @@ async function addAdminToManifest(
     throw new UpgradesError(
       `Proxy at ${proxyAddress} doesn't look like a transparent proxy`,
       () =>
+        `The proxy doesn't look like a transparent proxy because its admin address slot is empty. ` +
         `Set the \`kind\` option to the kind of proxy that was deployed at ${proxyAddress} (either 'uups' or 'beacon')`,
     );
   }
--- a/packages/plugin-hardhat/src/force-import.ts
+++ b/packages/plugin-hardhat/src/force-import.ts
@@ -111,7 +111,7 @@ async function addAdminToManifest(
 ) {
   const adminAddress = await getAdminAddress(provider, proxyAddress);
   if (isEmptySlot(adminAddress)) {
-    // Asser that the admin slot of a transparent proxy is not zero, otherwise the simulation below would fail due to no code at the address.
+    // Assert that the admin slot of a transparent proxy is not zero, otherwise the simulation below would fail due to no code at the address.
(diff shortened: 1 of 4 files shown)


packages/plugin-hardhat/src/force-import.ts
```diff
@@ -108,5 +110,14 @@ async function addAdminToManifest(
   opts: ForceImportOptions,
 ) {
   const adminAddress = await getAdminAddress(provider, proxyAddress);
+  if (isEmptySlot(adminAddress)) {
+    // Asser that the admin slot of a transparent proxy is not zero, otherwise the simulation below would fail due to no code at the address.
```
ernestognw, line 114:
```suggestion
    // Assert that the admin slot of a transparent proxy is not zero, otherwise the simulation below would fail due to no code at the address.
```

packages/plugin-hardhat/src/force-import.ts
```diff
@@ -108,5 +110,14 @@ async function addAdminToManifest(
...
+    // Note: Transparent proxies should not have the zero address as the admin, according to TransparentUpgradeableProxy's _setAdmin function.
```
ernestognw, line 115:
```suggestion
    // Note: Since contracts v5.0, transparent proxies should not have the zero address as the admin, according to TransparentUpgradeableProxy's _setAdmin function.
```

ernestognw (commented):
My 2 wei

packages/plugin-hardhat/src/force-import.ts
ericglau: This is also a restriction in 4.x though: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.9.3/contracts/proxy/ERC1967/ERC1967Upgrade.sol#L105

ernestognw: Ah I see, I thought it was because the admin is now immutable in Transparent Proxy, now I see this wasn't just reached in the past.

Solving

ernestognw (approved):

PR #895, opened 2023-10-12
Auto Verify After Deployment

passabilities:
* runs `verify:verify` hardhat script after deployments

### DIFF
--- a/packages/plugin-hardhat/src/utils/deploy-impl.ts
+++ b/packages/plugin-hardhat/src/utils/deploy-impl.ts
@@ -141,5 +141,10 @@ async function deployImpl(
     }
   }
 
+  await hre.run("verify:verify", {
+    address: deployment.address,
+    constructorArguments: opts.constructorArgs ?? []
+  });
+
   return { impl: deployment.address, txResponse };
 }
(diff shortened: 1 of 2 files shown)


ericglau:
@passabilities Thanks for suggesting this.  This seems like a useful feature to have, however I'm not sure whether this would work well in practice.

My main concerns are regarding latency, both for the plugin and from Etherscan.

First, depending on the scenario, we don't always wait for a deployment to be confirmed but often just return an ethers contract instance which is still pending deployment.  For example, after running `deployProxy`, users can call `contract.waitForDeployment()` on the returned contract instance to wait for the deployment.  For these cases, we would want to avoid having an internal waiting period (which would be needed before running Etherscan verify) and also avoid waiting for the verification itself.  Otherwise, the function would take too long to return.

Secondly, Etherscan does not immediately detect bytecode that was deployed to an address.  We can see this error message that could be thrown by the hardhat-verify plugin https://github.com/NomicFoundation/hardhat/blob/b06601b9d67bd1f33f5217e215d26eecde6ea86e/packages/hardhat-verify/src/internal/errors.ts#L154, where it suggests to wait perhaps 5 confirmations before running verification.  If we run verification immediately after a deployment, this kind of error could possibly occur quite frequently, whereas adding a delay would result in the first issue mentioned above.

Due to these reasons, it could be more reasonable to allow the user to just run verification in their own scripts as needed, as it is currently.  Let me know what you think.

PR #900, opened 2023-10-19
CLI: Add `--requireReference` option

ericglau:
Adds a `--requireReference` flag to make the CLI run in an upgrade-oritented mode, where it expects the contract specified by `--contract` to have a reference for storage layout comparisons.

This flag can be used when we want to ensure that storage layout checks are run.

When this option is enabled, either the `--reference <REFERENCE_CONTRACT>` flag must be provided as well, or the contract must have a `@custom:oz-upgrades-from <REFERENCE_CONTRACT>` annotation, so that storage layout comparisons can occur.


Amxx (approved):
LGTM

PR #906, opened 2023-10-20
CLI: Support Foundry contract naming conventions

### LINKED ISSUE #902: CLI: Support Foundry contract naming conventions
Contract names for the CLI currently need to be the actual contract name (e.g. `MyContract`), or the fully qualified name (e.g. `src/myPath/MyContract.sol:MyContract`).

To be more useful from Foundry, also support the following formats, according to the [`getCode` cheatcode](https://book.getfoundry.sh/cheatcodes/get-code):
- `MyContract.sol` if the filename and contract name are the same
- `MyContract.sol:Foo` 

Note: This does not include support for passing in Foundry artifact paths, as that is Foundry specific and not part of solc input.

ericglau:
Fixes #902

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - CLI: Add `--requireReference` option. ([#900](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/900))
 - CLI: Simplify summary message when using `--contract`. ([#905](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/905))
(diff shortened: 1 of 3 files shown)


ernestognw (approved):
I rewrote the tests because we shouldn't test `isMatch` but `findContract` instead. Otherwise there's no direct test connection between these tests and the actual CLI

LGTM

ericglau:
Cancelling since Foundry naming conventions are not standard, so should be handled by the caller on the Foundry side before calling the Upgrades CLI.

PR #911, opened 2023-10-24
Update to Node 18

ericglau:
Update to use Node 18, which is the current minimum supported LTS version of Node.js


ericglau:
Netlify preview has error:
```
3:38:05 PM: npm ERR! `node_gyp` is not a valid npm option
```

Amxx (approved):

PR #913, opened 2023-10-26
CLI: Throw error if `--requireReference` and `--unsafeSkipStorageCheck` are both enabled

ericglau:
The `requireReference` option is used for requiring a reference contract for storage compatibility checks, which contradicts with `unsafeSkipStorageCheck`.  Throw an error if both are enabled.


Amxx:
What is the argument for ignoring (silently) vs throwing an "incompatible options" error ?

ernestognw:
Same concern as @Amxx, the CLI arguments are contradictory and it's not obvious what the behavior would be.

ericglau:
Thanks, I agree with the comments.  Changed it to throw an error if both are enabled.

ernestognw (approved):
LGTM, thanks!

PR #918, opened 2023-11-08
Avoid duplicate file level functions and variables in namespace rewrite

### LINKED ISSUE #917: Hardhat compile error with `@openzeppelin/hardhat-upgrades`
I am developing a project on hardhat with foundry together, which uses the ERC6551 standard, which relies on account-abstraction. I have moved all [ERC6551](https://github.com/tokenbound/contracts) contracts to my `contracts/lib/erc6551` directory and installed [`@account-abstraction/contracts`](https://github.com/eth-infinitism/account-abstraction) as a dependency of ERC6551 contracts. It was done because ERC6551 contracts did not use the new 5th version of OpenZeppelin contracts and had some broken imports while initiating it via foundry libs. But when I ran compile on hardhat I received such an error:
```
An unexpected error occurred:

DeclarationError: Identifier already declared.
  --> @account-abstraction/contracts/core/Helpers.sol:32:5:
   |
32 |     uint256 constant _packValidationData = 0;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Note: The previous declaration is here:
  --> @account-abstraction/contracts/core/Helpers.sol:29:5:
   |
29 |     uint256 constant _packValidationData = 0;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^



```

The solidity code which provides such an error is located inside [`@account-abstraction/contracts`](https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/core/Helpers.sol#L73) lib:
```solidity
// I remove natspec comments from the code to be cleaner
// Also, need to mention that these functions are defined as 'free functions', which are not located inside the contracts

function _packValidationData(
    ValidationData memory data
) pure returns (uint256) {
    return
        uint160(data.aggregator) |
        (uint256(data.validUntil) << 160) |
        (uint256(data.validAfter) << (160 + 48));
}

function _packValidationData(
    bool sigFailed,
    uint48 validUntil,
    uint48 validAfter
) pure returns (uint256) {
    return
        (sigFailed ? 1 : 0) |
        (uint256(validUntil) << 160) |
        (uint256(validAfter) << (160 + 48));
}
```

I have tried such things to define what the problem is with compilation:
1. Reinstalled `@account-abstraction/contracts` lib, the error still was there.
2. Cloned `@account-abstraction/contracts` lib to check if it is not broken and could compile, also tried to compile it with my EVM compiler settings, lib compiled successfully.
3. Check if the installed version of `@account-abstraction/contracts` lib is the same as the ERC6551 dependency, it was the same and was `0.6.0`.
4. Find a fork of `@account-abstraction/contracts` which uses version 5 of OpenZeppelin contracts and tried to compile with it, got the same error.
5. Change env compiler versions: paris, london etc, did not help at all.

The last thing that I tried was to check hardhat's config properties one by one. I set up a new empty project with hardhat alone and installed some contract dependencies such as OpneZeppelin default and upgradeable contracts and `@account-abstraction/contracts` lib. Then tried to compile and the error was gone. Next was checking the hardhat configs, and I tried to remove `@openzeppelin/hardhat-upgrades` lib because a compile error left this message `Please report this at https://zpl.in/upgrades/report.` and it worked. The compilation was successful.

Is there an issue with `@openzeppelin/hardhat-upgrades` lib and does it override compilation settings somehow?

How I can use `@openzeppelin/hardhat-upgrades` lib in my situation as I need your lib to deploy and update a bunch of proxy contracts which I have in my project?

Repo with reproduction: https://github.com/vladikopl01/op-upgrades-compile-error

Amxx:
To extract storage layouts for namespaces, the Hardhat plugin makes a copy of the user's source code and makes modifications to this copy: it injects storage variables for namespaced structs and removes everything that is not needed (to speed up compilation).

File level functions and variables are not removed, but are replaced with dummy constants in case they are imported by another file.  However, overloaded functions would cause duplicate identifiers to exist, which causes compilation to fail when compiling this modified copy of source code.

We should keep track of identifiers to avoid duplicates in this case.

Fixes #917


Amxx:
TODO: we should add a default case for ast evolution (such as file-level event)

packages/core/src/utils/make-namespaced.ts
```diff
@@ -44,52 +44,74 @@ export function makeNamespacedInput(input: SolcInput, output: SolcOutput): SolcI
...
+          // - ImportDirective may import types used in structures with storage locations
+          // - PragmaDirective is necessary for compilation
+          // - StructDefinition may be used in structures with storage locations
+          // - UserDefinedValueTypeDefinition may be used in structures with storage locations
+          // - default: in case unexpected ast nodes show up (file-level events since 0.8.22)
+          break;
         }
-        // Replace with a dummy variable of arbitrary type
-        const name = node.name;
-        const insertText = `uint256 constant ${name} = 0;`;
```
ericglau, line 89: This was replaced by `uint256 constant` because the function definition or file level constant could be imported by another file, and we are not removing imports from the other file. I've added a testcase to reflect this.

Therefore we cannot outright delete it.  We cannot also just keep the original function or constant's source code, because they might reference deleted functions.

What we could consider is replace with a dummy variable but avoid duplicates if one already exists with the same name.

Amxx: That is a great point. I forgot about explicit imports :/

ericglau:
@Amxx I changed this to keep dummy variables for `ErrorDefinition`, `FunctionDefinition`, and `VariableDeclaration`, but avoids duplicate identifiers if one already exists due to overloading.  For now I think `FunctionDefinition` is really the only case where overloading could occur, but it could happen for `ErrorDefinition` when Solidity eventually supports overloading custom errors.  So for simplicity, I use the same code for all 3 of these cases.  Can you please review?

packages/core/src/utils/make-namespaced.ts
```diff
@@ -41,55 +40,112 @@ export function makeNamespacedInput(input: SolcInput, output: SolcOutput): SolcI
...
+          // Otherwise delete to avoid duplicate names, which can happen if there was overloading.
+          // This does not need to check all identifiers from the original contract, since the original compilation
+          // should have failed if there were conflicts in the first place.
+          const name = node.name;
+          if (!replacedIdentifiers.has(name)) {
+            const insertText = `uint256 constant ${name} = 0;`;
+            modifications.push(makeReplace(node, orig, insertText));
+            replacedIdentifiers.add(name);
+          } else {
+            modifications.push(makeDelete(node, orig));
```
ernestognw, line 131: Why are we deleting and not replacing it with another identifier instead? I saw we're replacing `UsingForDirective` with a dummy enum to avoid orphaning documentation.

ernestognw: Is it that variables don't have Natspec?

ernestognw (approved):
LGTM, just minor question

packages/core/src/utils/make-namespaced.ts
ericglau: Good point.  We are deleting the associated Natspec here, but there could still be standalone Natspec in the original source code which was not associated with anything.  Added some failing testcases in commit https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/918/commits/a5d1d8815a68296ae860a1aa32d3ef9e45c29da6 to demonstrate, which is the cause of the problem described in https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/917#issuecomment-1828565283.

I think we need to change the overall strategy to avoid deleting anything, but just replace them with dummy enums.  Invalid NatSpec tags seems to have no effect on enums.

ernestognw: > but there could still be standalone Natspec in the original source code which was not associated with anything.

This is a great point, thanks for noticing.

> I think we need to change the overall strategy to avoid deleting anything, but just replace them with dummy enums. Invalid NatSpec tags seem to have no effect on enums.

Yeah, I don't think this is trivial, but let's keep it in mind in case NatSpec becomes a problem, which I doubt.

ericglau: Changed it to not delete nodes but replace them with enums in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/918/commits/8badccedb36f428a298397bb1fb699a64071b0b5

ericglau:
Summary:

Changed to not delete any nodes and not delete documentation. Instead, replaces functions and other unneeded nodes with dummy enums, either with the original name (if the identifier could be imported in other files) or with a random identifier based on the AST node ID (if the identifier is no longer relevant or if it was originally an overloaded function).

This avoids any possibility of NatSpec getting orphaned or mixed up with different nodes than originally written, which could lead to compile errors.  Enums are able to compile with any NatSpec tag even if they are not relevant.

PR #919, opened 2023-11-09
Deploy proxy contracts from OpenZeppelin Contracts 5.0

### LINKED ISSUE #907: Deploy proxy contracts from OpenZeppelin Contracts 5.0
The proxy contracts deployed by the Hardhat Upgrades plugin are currently from OpenZeppelin Contracts v4.

Change the Hardhat Upgrades plugin to deploy proxy contracts from OpenZeppelin Contracts v5.

Currently, to use v5 proxy contracts, users need to deploy them without the plugin, then import them using [upgrades.forceImport](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades#force-import).

ericglau:
Targets release branch `hardhat-upgrades-v3` to allow for alpha version.

---

Deploys proxy contracts from OpenZeppelin Contracts 5.0, but continues to support importing and working with existing deployments from Contracts 4.x and below.

Updates Etherscan verification to verify proxy admins which are deployed by a transparent proxy rather than directly by the plugin.

`upgrades-core` will have proxy bytecodes for both 5.0 and 4.x proxies for backwards compatibility with older version of `hardhat-upgrades`.

Fixes #907 

- [x] Tested with local npm packages
- [x] Tested with Defender
- [x] Tested with Hardhat Verify


rsimonton:
@ericglau hello ser, is this blocked?  Anxiously awaiting this functionality, thanks.

packages/core/contracts/test/Proxiable.sol
```diff
@@ -1,7 +1,21 @@
...
+    function upgradeToAndCall(address newImplementation, bytes calldata data) external {
+        _authorizeUpgrade(newImplementation);
+        if (data.length > 0) {
+            (bool success, ) = address(this).call(data);
+            require(success, "upgrade call reverted");
+        }
+    }
+
+    function _authorizeUpgrade(address newImplementation) internal virtual;
+}
```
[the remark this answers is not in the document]
ericglau: This was done to remove the devDependency on `@openzeppelin/contracts-upgradeable` in core.  Simplified it further and clarified that it is only for tests to check the existence of the `upgradeToAndCall` function.

ericglau:
@rsimonton We're planning to have this available soon. We'll likely do an alpha release once this is merged so that you can start using it (it will be alpha for now since we still need to evaluate some breaking changes, particularly https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/912)

packages/plugin-hardhat/src/verify-proxy.ts
```diff
@@ -210,26 +214,57 @@ async function fullVerifyTransparentOrUUPS(
...
+      const deployedBytecode = await getCode(provider, adminAddress);
+      if (deployedBytecode !== artifact.deployedBytecode) {
+        throw new BytecodeNotMatchArtifact(
+          `Bytecode does not match with the current version of ${artifact.contractName} in the Hardhat Upgrades plugin.`,
+          artifact.contractName,
+        );
+      }
+
+      const constructorArgs = ethers.AbiCoder.defaultAbiCoder().encode(['address'], [owner]).replace(/^0x/, '');
+      await verifyContractWithConstructorArgs(etherscan, adminAddress, artifact, constructorArgs, errorReport);
```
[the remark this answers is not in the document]
ericglau: Good point. Changed it to look for the OwnershipTransferred event that should have been logged during construction and get the original owner from that.

packages/plugin-hardhat/test/uups-initial-owner.js
```diff
@@ -0,0 +1,17 @@
...
+const { ethers, upgrades } = require('hardhat');
+
+test.before(async t => {
+  t.context.Greeter = await ethers.getContractFactory('GreeterProxiable');
+});
+
+test('uups with initialOwner option', async t => {
+  const { Greeter } = t.context;
+
+  const initialOwner = (await ethers.getSigners())[1];
```
ernestognw, line 12:
```suggestion
  const initialOwner = await ethers.getSigner(1);
```

packages/plugin-hardhat/test/transparent-transfer-admin-ownership-wrong-signer.js
```diff
@@ -11,11 +11,14 @@ test.before(async t => {
...
   const signer = (await ethers.getSigners())[1];
 
-  await t.throwsAsync(() => upgrades.admin.transferProxyAdminOwnership(testAddress, signer), {
-    message: /(caller is not the owner)/,
-  });
+  const proxyAddress = await greeter.getAddress();
+
+  await t.throwsAsync(
+    () => upgrades.admin.transferProxyAdminOwnership(proxyAddress, testAddress, signer),
+    { message: /0x118cdaa7/ }, // abi encoding of OwnableUnauthorizedAccount(address)
```
ernestognw, line 22:
```suggestion
    { message: /0x118cdaa7/ }, // bytes4(keccak256('OwnableUnauthorizedAccount(address)'))
```

packages/plugin-hardhat/test/transparent-admin-initial-owner.js
```diff
@@ -0,0 +1,62 @@
+const test = require('ava');
+
+const { ethers, upgrades } = require('hardhat');
+const hre = require('hardhat');
+
+test.before(async t => {
+  t.context.Greeter = await ethers.getContractFactory('Greeter');
+});
+
```
ernestognw, line 9: What about adding this variable and replacing every occurrence?

```suggestion
const OWNABLE_ABI = ['function owner() view returns (address)']
```

packages/plugin-hardhat/test/transparent-admin-initial-owner.js
```diff
@@ -0,0 +1,62 @@
...
+test('initial owner - no signer in ContractFactory', async t => {
+  const defaultProvider = ethers.getDefaultProvider();
+
+  const Greeter = await ethers.getContractFactory('Greeter', defaultProvider);
+
+  await t.throwsAsync(upgrades.deployProxy(Greeter, ['hello']), {
+    message: /Initial owner must be specified/,
+  });
+
+  const initialOwner = (await ethers.getSigners())[2];
```
ernestognw, line 55:
```suggestion
  const initialOwner = await ethers.getSigner(2);
```

packages/plugin-hardhat/test/transparent-admin-initial-owner.js
```diff
@@ -0,0 +1,62 @@
...
+  const adminAddress = await upgrades.erc1967.getAdminAddress(await proxy.getAddress());
+  const admin = await hre.ethers.getContractAt(['function owner() view returns (address)'], adminAddress);
+
+  t.is(await admin.owner(), customSigner.address);
+});
+
+test('initial owner using initialOwner option', async t => {
+  const { Greeter } = t.context;
+
+  const initialOwner = (await ethers.getSigners())[2];
```
ernestognw, line 37:
```suggestion
  const initialOwner = await ethers.getSigner(2);
```

packages/plugin-hardhat/test/transparent-admin-initial-owner.js
```diff
@@ -0,0 +1,62 @@
...
+});
+
+test('initial owner using default signer', async t => {
+  const { Greeter } = t.context;
+
+  const proxy = await upgrades.deployProxy(Greeter, ['hello']);
+  const adminAddress = await upgrades.erc1967.getAdminAddress(await proxy.getAddress());
+  const admin = await hre.ethers.getContractAt(['function owner() view returns (address)'], adminAddress);
+
+  const defaultSigner = (await ethers.getSigners())[0];
```
ernestognw, line 17:
```suggestion
  const defaultSigner = await ethers.getSigner(0);
```

packages/plugin-hardhat/test/import-with-deploy-v5.js
```diff
@@ -0,0 +1,145 @@
...
+  t.is(
+    await upgrades.erc1967.getImplementationAddress(await greeterImportedAgain.getAddress()),
+    await upgrades.erc1967.getImplementationAddress(await greeterImported.getAddress()),
+  );
+});
+
+test('import then deploy transparent (with deployProxy) with different admin', async t => {
+  const { Greeter, GreeterV2, TransparentUpgradableProxy } = t.context;
+
+  const owner = (await ethers.getSigners())[0];
```
ernestognw, line 121:
```suggestion
  const owner = await ethers.getSigners(0);
```

ernestognw (commented):
On a side note, I think we should change all of the `@openzeppelin/contracts-v5` imports and use `@openzeppelin/contracts-v4` for v4 instead.

Should be added before the final 3.0 release, right?

ernestognw:
> Since `upgrades-core` already packages `/artifacts/@openzeppelin/contracts/proxy` for v4 proxies, we need to define a different naming scheme for v5 proxies.

Right, thanks for clarifying. I'd suggest creating an issue to track this if we ever bump upgrades/core so we can consider a fix, but you decide.

rsimonton:
You all are legends!  🫡

rsimonton:
I don't have time to open an issue atm but it looks like the `deployProxy` function doesn't support/respect explicitly typed data.

In my case I have contracts A and B, where B extends A.  Both are upgradeable.  B's initializer adds an additional argument.  When calling B's initializer (prior to my incorporating the upgrades plugin) in order for ethers to know which version of `initialize` to call, I had to explicitly type B's additional argument as described, for example, [here](https://github.com/ethers-io/ethers.js/discussions/4304#discussioncomment-6687043).

Trying to do this when calling `deployProxy`, like so:

```
     const proxy = await upgrades.deployProxy(
            contractFactory,
            [
                // .. other initializer args here
                ethers.Typed.uint16(foo)
            ],
            {
                kind: "uups",
            }
        )
```

...the deployment fails with `TypeError: ambiguous function description (i.e. matches "initialize(...`

Is it possible for the upgrades plugin to support the ethers v6 typed data API?  Or is there some other way to work around?  Thanks!

ericglau:
@rsimonton You can try specifying a fragment using the `initializer` option.  See [Specifying Fragments](https://docs.ethers.org/v5/api/utils/abi/interface/#Interface--specifying-fragments) (that's from ethers v5 doc but it should be the same for v6).

For example, you can provide the initializer function's signature like:
```
     const proxy = await upgrades.deployProxy(
            contractFactory,
            [
                // .. other initializer args here
                foo
            ],
            {
                kind: "uups",
                initializer: 'initialize(uint256)',
            }
        )
```
If this is still a problem for you, please open a separate issue, thanks!

PR #926, opened 2023-11-17
Ignore UPGRADE_INTERFACE_VERSION if selector returns anything other than "5.0.0"

### LINKED ISSUE #925: Ignore UPGRADE_INTERFACE_VERSION if selector returns anything other than "5.0.0"
To support OpenZeppelin Contracts 5.0, the upgrades plugin checks for the existence of a selector corresponding to `UPGRADE_INTERFACE_VERSION` in the contract.  If this is "5.0.0", it treats it as a v5 upgrade interface.  Currently, if this selector returns a non-string value or anything other than "5.0.0", the plugin throws an error since this is not a recognized value.

However, if the contract has a fallback function, this selector could exist and return anything (including a non-string value).  The upgrade should not fail in this case, but it should treat this as a pre-v5 upgrade interface.

Code to reproduce:
Have the following in the implementation contract
```
    fallback() external payable {}
```
Then run `upgradeProxy`.

ericglau:
Fixes #925 

Fixes upgrading a proxy from an implementation that has a fallback function and is not using OpenZeppelin Contracts 5.0.  Also handles the case where a custom proxy admin may have a similar fallback function.

The fallback function causes the UPGRADE_INTERFACE_VERSION selector to return a value which can be anything, so if it returns anything other than "5.0.0", we should treat it as a pre-5.0 interface.


packages/plugin-hardhat/test/transparent-admin-unknown-interface.js
```diff
@@ -0,0 +1,78 @@
...
+  t.context.TransparentUpgradableProxy = await ethers.getContractFactory(
+    TransparentUpgradableProxy.abi,
+    TransparentUpgradableProxy.bytecode,
+  );
+});
+
+function getInitializerData(contractInterface, args) {
+  const initializer = 'initialize';
+  const fragment = contractInterface.getFunction(initializer);
+  return contractInterface.encodeFunctionData(fragment, args);
```
Amxx, line 27:
```suggestion
 return contractInterface.encodeFunctionData('initialize', args);
```

packages/plugin-hardhat/contracts/GreeterProxiable40FallbackString.sol
```diff
@@ -0,0 +1,30 @@
...
+    function greet() public view returns (string memory) {
+        return greeting;
+    }
+
+    function setGreeting(string memory _greeting) public {
+        greeting = _greeting;
+    }
+
+    fallback(bytes calldata) external returns (bytes memory) {
+        return abi.encode(greeting);
```
Amxx, line 22: Does it not work with a direct cast 
```suggestion
        return bytes(greeting);
```
?

Using `abi.encode` does an extra encapsulation. Not sure that is needed.

packages/plugin-hardhat/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Fix `upgradeProxy` from an implementation that has a fallback function and is not using OpenZeppelin Contracts 5.0. ([#925](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/925))
```
ernestognw, line 5: Same as in core's changelog

packages/core/src/upgrade-interface-version.ts
```diff
@@ -13,7 +14,10 @@ export async function getUpgradeInterfaceVersion(
     // The first 32 bytes represent the offset, which should be 32 for a string
     const offset = parseInt(buf.slice(0, 32).toString('hex'), 16);
     if (offset !== 32) {
-      throw new Error(`Unexpected type for UPGRADE_INTERFACE_VERSION at address ${address}. Expected a string`);
+      // Log as debug and return undefined if the interface version is not a string.
+      // Do not throw an error because this could be caused by a fallback function.
+      debug(`Unexpected type for UPGRADE_INTERFACE_VERSION at address ${address}. Expected a string`);
```
ernestognw, line 19: I'd suggest adding a test to make sure this is logged if possible although it's just debug data. Surely is something you'd like to see if you're doing a production deployment.

ernestognw (approved):
LGTM, minor nits

packages/plugin-hardhat/contracts/GreeterProxiable40FallbackString.sol
ericglau: It actually was needed to force the return type to look like a string (added a test in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/926/commits/521f1acde17da7f075dcf0819518c6cb30f5e886)

packages/core/src/upgrade-interface-version.ts
ericglau: Thanks, this revealed that the tests were testing the wrong scenario, which is fixed in the latest two commits.

Added tests for the debug logging by using optional parameters for the relevant functions. It's not my preferred approach but couldn't get mocks to work well with this.

PR #939, opened 2023-12-08
Avoid address clash when redeploying implementation

### LINKED ISSUE #938: Deployment clashes with existing when `redeployImplementation: 'always'` and using Defender with salt
Scenario:
1. Run deployProxy, which tries to deploy the implementation, but assume gas limit is too low and it runs out of gas.
2. Run deployProxy using `useDefenderDeploy: true`, `salt` set, and `redeployImplementation: 'always'` because a deterministic address is preferred.
3. `Error: The following deployment clashes with an existing one at ...`

Consider whether `redeployImplementation: 'always'` should ignore clashes in the network file (perhaps as long as the rest of the deployment information is identical, including storage layout, bytecode version hash and address), because the intention as the option states is to redeploy.

ericglau:
Avoids checking for address clash when using `forceImport` or `redeployImplementation: 'always'`.

Address clashes should not occur in general, because it indicates a deployment is being attempted to an address which was already in the manifest, and such deployment should not occur if the bytecode version hash was the same before (unless using `forceImport` or `redeployImplementation: 'always'`, which triggers a deployment regardless of an existing bytecode version hash).  

Fixes #938

### DIFF
--- a/packages/core/src/impl-store.ts
+++ b/packages/core/src/impl-store.ts
@@ -229,17 +229,19 @@ async function checkForAddressClash(
   merge: boolean,
 ): Promise<void> {
   let clash;
+  let findClash = () => lookupDeployment(data, updated.address, !merge);
+
   if (await isDevelopmentNetwork(provider)) {
     // Look for clashes so that we can delete deployments from older runs.
     // `merge` only checks primary addresses for clashes, since the address could already exist in an allAddresses field
     // but the updated and stored objects are different instances representing the same entry.
-    clash = lookupDeployment(data, updated.address, !merge);
+    clash = findClash();
     if (clash !== undefined) {
       debug('deleting a previous deployment at address', updated.address);
       clash.set(undefined);
     }
   } else if (!merge) {
-    clash = lookupDeployment(data, updated.address, true);
+    clash = findClash();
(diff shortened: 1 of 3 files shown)


packages/core/src/impl-store.ts
```diff
@@ -223,28 +226,37 @@ async function checkForAddressClash(
...
         existing !== undefined &&
         (existing.remoteDeploymentId === undefined || existing.remoteDeploymentId !== updated.remoteDeploymentId)
       ) {
         throw new Error(
-          `The following deployment clashes with an existing one at ${updated.address}\n\n` +
-            JSON.stringify(updated, null, 2) +
-            `\n\n`,
+          `The deployment clashes with an existing one at ${updated.address}.\n\n` +
+            `Existing deployment: ${JSON.stringify(existing, null, 2)}\n\n` +
+            `New deployment: ${JSON.stringify(updated, null, 2)}\n\n`,
```
ernestognw, line 255: I am not sure printing both JSON's is ok, it's difficult to read in the CLI, what if?

```suggestion
            `Deployments: ${JSON.stringify({ existing, updated }, null, 2)}\n\n`,
```

PR #948, opened 2023-12-19
Update unidici dependency

ericglau:


ernestognw (approved):
LGTM with some comments:

I'm not convinced about the `as <Type>` casts. Following [these recommendations](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36868), we should create functions for checking types, for example:

```ts
function isEtherscanResponseBody(input: any): json is EtherscanResponseBody {
  return 'status' in input;
} 
```

However, I also don't think we should accumulate this type-checking functions all over the place. The current state is good to me but mentioning in case you agree with "better" type-checking @ericglau

ericglau:
@ernestognw Agree it would be good to improve type checking, but perhaps that can be left until/unless there are changes in Etherscan's API or issues with the current approach.  I don't a large benefit in adding checking at the moment, other than to give a better error message, but if errors occur then the issue might be quite apparent.

Note that Hardhat currently does a [cast](https://github.com/NomicFoundation/hardhat/blob/3f4680f21f32e0165bee266ea3b08315db15f92f/packages/hardhat-verify/src/internal/etherscan.ts#L168) as well.

Issue #943, opened 2023-12-13
Cannot read properties of null (reading 'chainId')


adrienthiery:
I'm getting this error out of the blue with the latest versions when running: `npx hardhat run --network private scripts/deploy.js`
 
```TypeError: Cannot read properties of null (reading 'chainId')
    at new Manifest (/Users/adrien/Code/my-project/my-project-smart-contracts/node_modules/@openzeppelin/upgrades-core/src/manifest.ts:133:65)
    at Function.forNetwork (/Users/adrien/Code/my-project/my-project-smart-contracts/node_modules/@openzeppelin/upgrades-core/src/manifest.ts:104:14)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Proxy.deployProxy (/Users/adrien/Code/my-project/my-project-smart-contracts/node_modules/@openzeppelin/hardhat-upgrades/src/deploy-proxy.ts:46:22)
    at deployMyProjectNFT (/Users/adrien/Code/my-project/my-project-smart-contracts/scripts/deploy.js:106:21)
    at main (/Users/adrien/Code/my-project/my-project-smart-contracts/scripts/deploy.js:29:5)
    at /Users/adrien/Code/my-project/my-project-smart-contracts/scripts/deploy.js:128:7
```

When logging this is the node modules, I find this is on [this line](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/6a89da4ee700aa9dcae96e669d93b7f8c38b8728/packages/core/src/manifest.ts#L133) and if I log `devInstanceMetadata`, I get: 
```
{
  networkName: 'hardhat',
  instanceId: '0xb7a74718b581ff4054338676a6cda012eb69bdccaab5f739f66eb1a8f6e5362e',
  forkedNetwork: null
}
```

<details>
  <summary>My `hardhat.config.js` is as follows</summary> 
  
```
require("hardhat-contract-sizer");
require("@nomicfoundation/hardhat-toolbox");
require("@openzeppelin/hardhat-upgrades");
require("hardhat-gas-reporter");
require("hardhat-tracer");

const path = require("path");

let BUILD_DIR = path.join(__dirname, "../my-project-frontend/public/hardhat/");
if (process.env.BUILD_DIR) {
  BUILD_DIR = process.env.BUILD_DIR;
}

let COIN_MARKET_CAP_API_KEY = "some-key";
if (process.env.COIN_MARKET_CAP_API_KEY) {
[...]

PR #950, opened 2023-12-20
Fix manifest error when connecting to an Anvil dev network

ericglau:
When handling network manifest files, we use the [RPC method `hardhat_metadata`](https://hardhat.org/hardhat-network/docs/reference#hardhat_metadata) to determine if the connected network is a Hardhat development network, in which case we write the manifest to the temp dir instead of `.openzeppelin`.

Furthermore, if this method has a `forkedNetwork` field in its response, we treat this as a network fork and make a copy of the origin networks's manifest file, so that we can write to it for the fork without affecting the original version.

In recent versions of Anvil, Anvil also supports the `hardhat_metadata` RPC method, but has `"forkedNetwork": null` in its response as seen [here](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/943#issuecomment-1864052578).

This PR does two things:
1. Try calling `anvil_metadata` to determine if Anvil, and if so, use `anvil` for the manifest file prefix to more accurately reflect the network type.
2. Check if `forkedNetwork` is null or undefined, in which case we determine that it is not a network fork.

Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/943

### DIFF
--- a/packages/core/src/manifest.ts
+++ b/packages/core/src/manifest.ts
@@ -58,7 +58,6 @@ async function getDevInstanceMetadata(
   chainId: number,
 ): Promise<DevInstanceMetadata | undefined> {
   let networkMetadata: HardhatMetadata;
-
   let networkType: DevNetworkType;
   try {
     networkMetadata = await getAnvilMetadata(provider);
--- a/packages/core/src/manifest.ts
+++ b/packages/core/src/manifest.ts
@@ -48,25 +48,23 @@ async function getDevInstanceMetadata(
   provider: EthereumProvider,
   chainId: number,
 ): Promise<DevInstanceMetadata | undefined> {
-  let hardhatMetadata;
-
   try {
-    hardhatMetadata = await getHardhatMetadata(provider);
+    const hardhatMetadata = await getHardhatMetadata(provider);
+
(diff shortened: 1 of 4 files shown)


packages/core/src/manifest.ts
```diff
@@ -63,7 +63,7 @@ async function getDevInstanceMetadata(
   }
 
   return {
-    networkName: 'hardhat',
+    networkName: hardhatMetadata.clientVersion.startsWith('anvil') ? 'anvil' : 'hardhat',
```
ernestognw, line 66: My only concern is that this value is depending on `clientVersion` to always start with `anvil`, and may suddenly not be true again.

Nonblocking, we already chatted about this and it's not an issue. Just for the record.

ericglau: Changed this to try calling `anvil_metadata` to determine if Anvil, otherwise try `hardhat_metadata` to determine if Hardhat, otherwise conclude that it is neither.

AaronNan24:
Hi guys, thanks for ur effort!

wondering that have we updated to NPM hosting? It still looks old to me
CC @ericglau @ernestognw 

image

PR #954, opened 2024-01-05
Use query instead of body for Etherscan API

mikkel-arturo:
When attempting to verify contracts on Snowtrace, I received the following error:
`Etherscan returned with message: NOTOK, reason: Invalid querystring request`

Looking at the Etherscan API I noticed that all endpoints used query parameters instead of a form-encoded body like in the code. With this change my verification calls are now working.

It's not clear to me why the original implementation was created or how it's working on Etherscan - I assume that Snowtrace conforms closer to the API spec and that is why the change is required.

### DIFF
--- a/packages/plugin-hardhat/CHANGELOG.md
+++ b/packages/plugin-hardhat/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


ericglau (approved):
Thanks for the PR!

PR #958, opened 2024-01-15
CLI: Improve checks for build info file settings

ericglau:
This PR addresses 2 issues:

1. When running validations from CLI, we previously looked for the following `outputSelection` from the build info file, and gave an error if that was missing:
```
      outputSelection: {
        '*': {
          '*': ['storageLayout']
        }
      }
```

However, in recent versions of Foundry (reported in https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades/issues/4), running `forge script` no longer uses the above format.  Instead, it has settings enabled for each individual contract, like:
```
      outputSelection: {
        'mypath/MyContract.sol': {
          '': ['ast'],
          '*': ['abi', 'evm.bytecode', 'evm.deployedBytecode', 'evm.methodIdentifiers', 'metadata', 'storageLayout']
        }
     }
```

This PR checks each contract's settings and gives an error if any of them do not have `storageLayout`.

2. Additionally, if users modify a Solidity file and run Foundry compilation, Foundry does a partial compilation by default.  This causes the updated build info file to contain AST nodes that reference node IDs from older build-info files, and we currently do not support this. Therefore, we throw an error telling users to do a full recompilation.

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- CLI: Improve checks for build info file settings. ([#958](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/958))
+
 ## 1.32.2 (2023-12-21)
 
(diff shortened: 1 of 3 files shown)


ernestognw (approved):
LGTM, thanks @ericglau !

Issue #959, opened 2024-01-18
500 error when I try to deploy a non-upgradable contract with a linked library and CREATE2 salt


cedoor:
I'm using the `defender.deployContract` function to deploy (deterministically) a contract ([Semaphore.sol](https://github.com/semaphore-protocol/semaphore/blob/chore/oz-deployer/packages/contracts/contracts/Semaphore.sol)) where its constructor parameter is a simple address and it also needs a linked library ([PoseidonT3](https://github.com/vimwitch/poseidon-solidity/blob/main/contracts/PoseidonT3.sol)).

```ts
const SemaphoreFactory = await ethers.getContractFactory("Semaphore", {
  libraries: {
      PoseidonT3: poseidonAddress
  }
})

const semaphore = await defender.deployContract(SemaphoreFactory, [verifierAddress], {
    salt: process.env.CREATE2_SALT,
    unsafeAllow: ["external-library-linking", "constructor"],
    unsafeAllowDeployContract: true
})

await semaphore.waitForDeployment()
```

Ref: https://github.com/semaphore-protocol/semaphore/blob/f0be86671278f6ab610ac36da9196b8d74d757c6/packages/contracts/tasks/deploy-semaphore.ts#L62-L66.

I've also tried to use some unsafeAllow options but I get the same error:

DefenderApiResponseError: Request failed with status code 500.
```js
{
  request: { path: '/v2/deployments', method: 'POST' },
  response: {
      status: 500,
      statusText: 'Internal Server Error',
      data: {
        message: 'invalid arrayify value (argument="value", value="0x60806040523480156200001157600080fd5b50... ~17k more characters',
        ...
      }
  }
}
```

I have no problems with the other contracts, which have no constructors or linked libraries (e.g. [SemaphoreVerifier.sol](https://github.com/semaphore-protocol/semaphore/blob/chore/oz-deployer/packages/contracts/contracts/base/SemaphoreVerifier.sol)).

ericglau:
Hi @cedoor, thanks for reporting this. We've identified an internal Defender error that occurs in this scenario and are working on a fix. I'll let you know when the fix is available.

cedoor:
> Hi @cedoor, thanks for reporting this. We've identified an internal Defender error that occurs in this scenario and are working on a fix. I'll let you know when the fix is available.

Hi @ericglau, thank you very much!

ericglau:
[...]

PR #960, opened 2024-01-19
Support externally linked libraries for Defender deployments

ericglau:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/959

---

PR summary:

If a contract has externally linked libraries, Hardhat requires that users provide a `libraries` parameter when creating their Ethers contract factory, for example using library name and address:
```
const MyContractFactory = await ethers.getContractFactory("MyContract", {
  libraries: {
    MyLibrary: "0x..."
  }
});
```
or fully qualified library name and address:
```
const MyContractFactory = await ethers.getContractFactory("MyContract", {
  libraries: {
    "contracts/MyLibrary.sol:MyLibrary": "0x..."
  }
});
```

When perform deployments using Defender, we need to extract this information and pass it through to the Defender SDK using the fully qualified library name and address.  We cannot get this directly from the Ethers contract factory, because Hardhat discards the `libraries` object after linking the bytecode.

We solve this by getting the linked library information that we cached for validations, finding the corresponding addresses in the Ethers contract factory's linked bytecode, and building a new `libraries` parameter that always has fully qualified library names. 

---

Example usage:
```
const MyContractFactory = await ethers.getContractFactory("MyContract", {
  libraries: {
    MyLibrary: "0x..."
  }
});

const myContract = await defender.deployContract(MyContractFactory, [args],  { unsafeAllowLinkedLibraries: true }})
```

---


packages/plugin-hardhat/src/defender/deploy.ts
```diff
@@ -132,12 +139,31 @@ async function getContractInfo(
...
+
+      const start = linkReferences[ref].start * 2;
+      const length = linkReferences[ref].length * 2;
+
+      const linkFullyQualifiedName = getFullyQualifiedName(linkReferences[ref].src, linkReferences[ref].name);
+      const linkAddress = `0x${linkedBytes.substring(start, start + length)}`;
+
+      if (libraries === undefined) {
+        libraries = {};
+      }
```
ernestognw, line 163: Why not? 

```suggestion
      libraries ??= {}
```

ernestognw (approved):
LGTM

PR #963, opened 2024-01-30
Add op-sepolia to manifest file names

cruzdanilo:


packages/core/src/provider.ts
```diff
@@ -143,6 +143,7 @@ export const networkNames: { [chainId in number]?: string } = Object.freeze({
   43114: 'avalanche',
   42220: 'celo',
   44787: 'celo-alfajores',
+  11155420: 'op-sepolia',
   11155111: 'sepolia',
```
Amxx, line 147: IMO they should be sorted by chainId
```suggestion
  11155111: 'sepolia',
  11155420: 'op-sepolia',
```

ericglau (commented):
Thanks for the PR.

Can you please a [changelog entry](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/master/packages/core/CHANGELOG.md) at the top with heading `## Unreleased`?

See https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/master/packages/core/CHANGELOG.md#1250-2023-04-26 for example.

Amxx (commented):
420 is marked as "optimism-goerli"

Should this be "optimism-sepolia" ? Is there a strong convension to use "op-" vs "optimism-" ?

cruzdanilo:
i understand there's no way to easily rename current networks (the "Optimism" network is now called "OP Mainnet"), but i feel like new networks should use the proper names.

ericglau (approved):

PR #967, opened 2024-02-01
Support Defender deployments using EOA or Safe

### LINKED ISSUE #923: Update use of Defender SDK to handle optional address in deployment response
Deployments using Defender SDK has a recent [change in its return types](https://github.com/OpenZeppelin/defender-sdk/pull/120) that makes the `address` and `txHash` fields optional in deployment responses.  This occurs to support using an EOA or Safe for deployments.

The current Hardhat Upgrades plugin particularly relies on the `address` field to be returned, since this is saved to the network files and is used to wrap the deployment response into an Ethers-compatible contract instance (which requires an address).

Hardhat Upgrades should be updated to handle the optional `address` field above.  Possible approaches could be relying only on the deployment id (although this slightly breaks Ethers compatibility), or always wait for deployments to finish so that it can retrieve the address.

ericglau:
Fixes #923 

The Hardhat Upgrades plugin typically expects address and txHash to be available immediately after submitting a deployment.
For Defender deployments using Relayers, these are available.
For Defender deployments using EOA or Safe, these are only available when the deployment completes.  This PR checks if these are not available, in which case we prompt to tell the user to submit the deployment in Defender and waits until these are available.

Also updates `salt` option documentation based on [this information](https://docs.openzeppelin.com/defender/v2/tutorial/deploy#deploy-caveat)

### DIFF
--- a/docs/modules/ROOT/pages/api-hardhat-upgrades.adoc
+++ b/docs/modules/ROOT/pages/api-hardhat-upgrades.adoc
@@ -31,7 +31,7 @@ The following options are common to some functions.
 * `useDefenderDeploy`: (`boolean`) Deploy contracts using OpenZeppelin Defender instead of ethers.js. See xref:defender-deploy.adoc[Using with OpenZeppelin Defender]. **Note**: OpenZeppelin Defender deployments is in beta and functionality related to it is subject to change.
 * `verifySourceCode`: (`boolean`) When using OpenZeppelin Defender deployments, whether to verify source code on block explorers. Defaults to `true`.
 * `relayerId`: (`string`) When using OpenZeppelin Defender deployments, the ID of the relayer to use for the deployment. Defaults to the relayer configured for your deployment environment on Defender.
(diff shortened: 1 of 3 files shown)


ernestognw (approved):
Alternatively to query Defender's status, getting the transaction hash is possible before confirmation, so that we can check for the status without waiting for confirmations and speeding up the process.

~~I'll suggest this to the Defender team~~ Figured out the tx hash depends on the executor for safes and the nonce if it's an EOA

ericglau:
Note that we are only waiting for the address and txHash to be present, but not strictly checking for a "completed" status at this point (even though we describe it as such, for ease of explanation and because that is the most likely scenario).  We proceed as long as those are present.

Defender does allow for transactions to be replaced, so the txHash could possibly change at a latter step.  For example, tx replacements are tested in [this testcase](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/96b62bdd7bfeaf31e23be6b2a86d13c16d7d6b02/packages/core/src/deployment.test.ts#L130).

PR #972, opened 2024-02-14
Remove devDependency lerna

ericglau:
Closes https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/929


ernestognw:
Seems like there are also [breaking changes for 7.0](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#breaking-changes-1).

Have you run `lerna repair`?

ericglau:
After further consideration, I think we can remove the dependency altogether.  We mainly use lerna for publishing, but since we only have 2 packages at this point (core and hardhat-upgrades), we can publish them individually.

ernestognw:
> I think we can remove the dependency altogether. We mainly use lerna for publishing, but since we only have 2 packages at this point (core and hardhat-upgrades), we can publish them individually.

Sounds good to me for now. We might reconsider if we ever release something for the incoming Stylus library (EVM compatible)

ernestognw (approved):

PR #973, opened 2024-02-16
Remove pack workflow

ericglau:
I don't think we use the pack workflow for anything. Seems it was originally added for testing purposes, but we don't make use of its results. Removing this workflow, since it is now failing because we removed lerna in https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/972.


ernestognw (approved):
This is weird but yeah, it looks unnecessary. LGTM

PR #974, opened 2024-02-20
Add holesky network

islishude:


Amxx (approved):

ericglau (approved):
Reverted the change from `migrate-oz-cli-project.ts` since that is only for migrating from old CLI projects, which did not support Holesky in the first place.

Thanks for the PR!

PR #978, opened 2024-02-28
Remove console logging for transferProxyAdminOwnership

ericglau:
Fixes issue reported in https://forum.openzeppelin.com/t/openzeppelin-hardhat-upgrades-s-admin-transferproxyadminownership-always-logs-success/39686

transferProxyAdminOwnership previously iterated through all proxies from the manifest file, and displayed either:
```
console.log(SUCCESS_CHECK + `${address} (${kind}) proxy ownership transferred through proxy admin`);
```
or
```
console.log(FAILURE_CROSS + `${address} (${kind}) proxy ownership not affected by proxy admin`);
```
to help the user see which proxies were using that proxy admin.  This was useful with proxies from OpenZeppelin Contracts 4.x because the plugin only tracked 1 proxy admin and all transparent proxies were using the same admin.

But with OpenZeppelin Contracts 5.x where each proxy deploys its own proxy admin, this is less relevant.  The proxy address is a parameter of the `transferProxyAdminOwnership` function, and this function affects its proxy admin which is typically for only that specific proxy.

Therefore, this PR just removes this logging altogether.

### DIFF
--- a/packages/plugin-hardhat/src/admin.ts
+++ b/packages/plugin-hardhat/src/admin.ts
@@ -68,7 +68,14 @@ export function makeTransferProxyAdminOwnership(
       const { proxies } = await manifest.read();
       const adminAddress = await admin.getAddress();
 
-      const affected = proxies.filter(async proxy => (await getAdminAddress(provider, proxy.address)) === adminAddress);
+      const affected = [];
+      for (const proxy of proxies) {
+        const controller = await getAdminAddress(provider, proxy.address);
+        if (controller === adminAddress) {
+          affected.push(proxy);
+        }
+      }
+
       if (affected.length > 0) {
         console.log(SUCCESS_CHECK + `${affected.length} proxies ownership transferred through proxy admin`);
         affected.forEach(proxy => console.log(`    - ${proxy.address} (${proxy.kind})`));
--- a/packages/plugin-hardhat/src/admin.ts
+++ b/packages/plugin-hardhat/src/admin.ts
@@ -1,12 +1,8 @@
-import chalk from 'chalk';
 import type { HardhatRuntimeEnvironment } from 'hardhat/types';
-import { Manifest, getAdminAddress } from '@openzeppelin/upgrades-core';
+import { getAdminAddress } from '@openzeppelin/upgrades-core';
 import { Contract, Signer } from 'ethers';
 import { EthersDeployOptions, attachProxyAdminV4 } from './utils';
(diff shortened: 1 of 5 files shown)


Amxx:
If a dev has a manifest that contains the old proxy admin and proxies, and if that dev transfers the ownership of the proxy admin, don't we want to confirm that some known proxy are affected?

I would definitelly remove the "failure" side, because it not an error for a new proxy to not be affected, but I'd still keep some logs. Maybe doing something like:

```typescript
    const adminAddress = await admin.getAddress();
    const affected = await Promise.all(
        manifest.read().then(({ proxies }) => proxies.map(proxy => getAdminAddress(provider, address).then(a => a === adminAddress && proxy))
    ).then(list => list.filter(Boolean));
    
    console.log(SUCCESS_CHECK + `${affected.length} proxies ownership transferred through proxy admin`);
    affected.forEach(({ address, kind }) => console.log(`    - ${address} (${kind})`));
```

ericglau:
Added @Amxx's suggestion, and also added a `silent` option for the `transferProxyAdminOwnership` function if the user really wants to silence it.

packages/plugin-hardhat/src/admin.ts
```diff
@@ -59,14 +62,16 @@ export function makeTransferProxyAdminOwnership(
...
-        console.log(SUCCESS_CHECK + `${address} (${kind}) proxy ownership transferred through proxy admin`);
-      } else {
-        console.log(FAILURE_CROSS + `${address} (${kind}) proxy ownership not affected by proxy admin`);
+    if (!opts.silent) {
+      const { provider } = hre.network;
+      const manifest = await Manifest.forNetwork(provider);
+      const { proxies } = await manifest.read();
+      const adminAddress = await admin.getAddress();
+
+      const affected = proxies.filter(async proxy => (await getAdminAddress(provider, proxy.address)) === adminAddress);
```
Amxx, line 71: Can you filter using an async function without some `await Promise.all` ?

Amxx:
```suggestion
      const affected = await Promise.all(proxies.map(
          proxy => getAdminAddress(provider, proxy.address).then(controller => controller === adminAddress && proxy)
      )).then(array => array.filter(Boolean));
```

ericglau: I find this syntax quite confusing and have changed it to use a simple for-loop instead.

PR #989, opened 2024-03-07
Support private networks and forked networks with Defender

### LINKED ISSUE #988: Support private networks for Defender deployments
If using a private network with a chainId that is not known to Defender SDK, currently the plugin gives the error:
```
Network ${chainId} is not supported by OpenZeppelin Defender
```
Change this to lookup the private network name and pass that in to the deployment request.

ericglau:
Fixes #988

### DIFF
--- a/packages/plugin-hardhat/src/defender/utils.ts
+++ b/packages/plugin-hardhat/src/defender/utils.ts
@@ -10,6 +10,7 @@ import {
 
 import { Network, fromChainId } from '@openzeppelin/defender-sdk-base-client';
 import { DeployClient, TxOverrides } from '@openzeppelin/defender-sdk-deploy-client';
+import { NetworkClient } from '@openzeppelin/defender-sdk-network-client';
 
 import { HardhatDefenderConfig } from '../type-extensions';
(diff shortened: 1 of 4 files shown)


packages/plugin-hardhat/src/defender/utils.ts
```diff
@@ -35,6 +36,20 @@ export async function getNetwork(hre: HardhatRuntimeEnvironment): Promise<Networ
   const chainId = hre.network.config.chainId ?? (await getChainId(provider));
   const network = fromChainId(chainId);
   if (network === undefined) {
```
ernestognw, line 38: Same as [here](https://github.com/OpenZeppelin/defender-deploy-client-cli/pull/7#discussion_r1516524208)

ericglau: Updated to support a `network` field in `defender` section of Hardhat config. If set, the network name must match, and this is used to disambiguate if there are multiple networks with the same chainId.

ernestognw (approved):
LGTM 💯

PR #992, opened 2024-03-15
Fix typos

xiaoxianBoy:
Fix `packages/core/src/storage/report-rename-retype.test.ts.md` misspell


ericglau (changes requested):
Thank you for the PR. However, these are snapshots that are generated from tests. Could you please do the following instead?

1. Edit the test file e.g. https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/e125afd90c6f70a6510b754cf22a9c24ce3642ee/packages/core/src/storage/report-rename-retype.test.ts#L62
2. Run this from the `packages/core` directory to regenerate snapshots: `yarn tsc -b && yarn ava --update-snapshots`
3. Commit the changes

ericglau:
Resolved by https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/1010

Issue #672, opened 2022-10-06
Support configuration of .openzeppelin directory path


AboldUSER:
Our organization uses Hardhat for deployment across multiple private chains, and configures the paths of Hardhat's inputs and outputs (https://hardhat.org/hardhat-runner/docs/config#path-configuration) into folders not necessarily in the project's directory.

Currently the .openzeppelin directory is created in the root of where the oz-upgrades scripts are called from, at least when using the Hardhat plugin. This is due to the openZeppelin-upgrades directory being hardcoded as '.openzeppelin' in the core package's file [./packages/core/src/manifest.ts](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/master/packages/core/src/manifest.ts#L41), and that string is defined as the .openzeppelin path in the [manifest.ts constructor](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/master/packages/core/src/manifest.ts#L58).

I'd like to have optional configuration of the manifest directory path, with a default to the current manifestDir string, '.openzeppelin', similar to how Hardhat handles configuration of its paths.

A possible solution would be to pass an optional parameter into the manifest.ts constructor that defines the manifest path for the this.fallbackFile and this.file variables, and falls back to the default path if the parameter is not passed/null. This passing of path data can be similar to how the manifest.ts constructor currently takes the chainId, which comes from the deployData object defined in the plugin packages. In the Hardhat plugin, when creating the deployData object in [getDeployData function](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/master/packages/plugin-hardhat/src/utils/deploy-impl.ts#L41), another const can be added that reads the hre object for a path defined in the hardhat.config file, such as 'hre.config.ozUpgrades.manifestPath'. I am assuming something similar can be done with the Truffle plugin.

frangio:
You included a lot of details about how this would be implemented, but we should focus on what this is going to look like for the end-user.

We currently don't have any global configuration (in hardhat.config.js) so I'd try to avoid that. Would it be sufficient to include it as an option in functions like `deployProxy`? `deployProxy(..., { ..., manifest: 'path/to/manifest.json' })`

AboldUSER:
My initial thought on the hardhat.config file was because that's how I'm use to using other hardhat plugins. I guess I let my tool bias drive the above 😅

[...]

PR #1011, opened 2024-04-17
chore: add possibility of changing default MANIFEST_DEFAULT_DIR with env variable

3h4x:
For a repo that uses multiple environments deploying to the same networks, static MANIFEST_DEFAULT_DIR set to .openzeppelin is not enough.

I know it's simple way of handling it via environment variable but I haven't seen in the repo best way to make such changes.
If this is not preferred way, please let me know and I will update this PR with changes

Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/672

❤️

### DIFF
--- a/docs/modules/ROOT/pages/network-files.adoc
+++ b/docs/modules/ROOT/pages/network-files.adoc
@@ -62,12 +62,12 @@ You can determine the location of a temporary network file by doing the followin
 . Run your Hardhat test or script.
 . Find the log message containing the text `development manifest file:`.
 
-[[custom-files-location]]
-== Custom Files Location
+[[custom-network-files-location]]
+== Custom Network Files Location
 
-To customize the location of the `.openzeppelin` folder, you can set the `MANIFEST_DEFAULT_DIR` environment variable. This variable allows you to specify different directories for storing configuration files, enabling you to deploy same contracts, to the same networks for different environments without conflicts.
+To customize the location of the `.openzeppelin` folder, set the `MANIFEST_DEFAULT_DIR` environment variable to an absolute path or a path relative to the root of your project. This variable allows you to specify different directories for storing network files, enabling you to deploy the same contracts to the same networks for different environments without conflicts. This can be used with private networks to avoid chain ID conflicts.
 
 For example:
 
-* Setting `export MANIFEST_DEFAULT_DIR=.openzeppelin/testnet` will configure OpenZeppelin Hardhat Upgrades to use `.openzeppelin/testnet/<network_name>.json` for testnet contracts.
-* Setting `export MANIFEST_DEFAULT_DIR=.openzeppelin/mainnet` will configure OpenZeppelin mainnet Upgrades to use `.openzeppelin/mainnet/<network_name>.json` for mainnet contracts.
+* Run `export MANIFEST_DEFAULT_DIR=.openzeppelin/tests` to configure OpenZeppelin Hardhat Upgrades to use `.openzeppelin/tests/<network_name>.json` for test deployments.
+* Run `export MANIFEST_DEFAULT_DIR=.openzeppelin/production` to configure OpenZeppelin Hardhat Upgrades to use `.openzeppelin/production/<network_name>.json` for production deployments.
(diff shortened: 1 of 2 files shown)


packages/core/src/manifest.ts
```diff
@@ -48,7 +48,7 @@ function defaultManifest(): ManifestData {
   };
 }
 
-const MANIFEST_DEFAULT_DIR = '.openzeppelin';
+const MANIFEST_DEFAULT_DIR = process.env.MANIFEST_DEFAULT_DIR || '.openzeppelin';
```
ericglau, line 51: I think it would be better to use an environment variable name such as `OPENZEPPELIN_NETWORK_FILES_DIR` to more closely match with the ["Network Files"](https://docs.openzeppelin.com/upgrades-plugins/1.x/network-files) term used in the documentation.  It may also be useful to make it load from a `.env` file.  For example:
```suggestion
require('dotenv').config();
const MANIFEST_DEFAULT_DIR = process.env.OPENZEPPELIN_NETWORK_FILES_DIR || '.openzeppelin';
```

Note this would also need a `"dotenv"` dependency in `packages/core/package.json`

ericglau (commented):
Thanks for the PR, this looks like a good approach!  Just left one comment.

Would you also be able to add something in the documentation in `docs/modules/ROOT/pages/network-files.adoc` to mention this environment variable?

packages/core/src/manifest.ts
3h4x: Hey, thanks for quick review @ericglau 
I did add `dotenv` to the dependencies but as there is no `.env` file I started to have second thoughts if adding library just for sake of one env variable is reasonable. I'm very reluctant to add new libraries to the codebase and always try to keep it vanilla.
Also, my rule of thumb is to refactor for third repeating element. We haven't reached that point. 

I'm open to hear, what you think about this. If you think this is good idea then no problem in implementing it but I don't think I should be making decision about it. ✌️

3h4x: I added section of documentation that explains this feature. Let me know if this is up to `openzeppelin` standards

docs/modules/ROOT/pages/network-files.adoc
```diff
@@ -61,4 +60,14 @@ You can determine the location of a temporary network file by doing the followin
...
+
+[[custom-files-location]]
+== Custom Files Location
+
+To customize the location of the `.openzeppelin` folder, you can set the `MANIFEST_DEFAULT_DIR` environment variable. This variable allows you to specify different directories for storing configuration files, enabling you to deploy same contracts, to the same networks for different environments without conflicts.
+
+For example:
+
+* Setting `export MANIFEST_DEFAULT_DIR=.openzeppelin/testnet` will configure OpenZeppelin Hardhat Upgrades to use `.openzeppelin/testnet/<network_name>.json` for testnet contracts.
+* Setting `export MANIFEST_DEFAULT_DIR=.openzeppelin/mainnet` will configure OpenZeppelin mainnet Upgrades to use `.openzeppelin/mainnet/<network_name>.json` for mainnet contracts.
```
ericglau, line 73: Edited the wording and examples to make it clearer:
```suggestion
[[custom-network-files-location]]
== Custom Network Files Location

To customize the location of the `.openzeppelin` folder, set the `MANIFEST_DEFAULT_DIR` environment variable to an absolute path or a path relative to the root of your project. This variable allows you to specify different directories for storing network files, enabling you to deploy the same contracts to the same networks for different environments without conflicts. This can be used with private networks to avoid chain ID conflicts.

For example:

* Run `export MANIFEST_DEFAULT_DIR=.openzeppelin/tests` to configure OpenZeppelin Hardhat Upgrades to use `.openzeppelin/tests/<network_name>.json` for test deployments.
* Run `export MANIFEST_DEFAULT_DIR=.openzeppelin/production` to configure OpenZeppelin Hardhat Upgrades to use `.openzeppelin/production/<network_name>.json` for production deployments.
```

packages/core/src/manifest.ts
ericglau: Thanks for the comments, I think the env var itself is fine as is.  I left a separate suggestion for the docs.

docs/modules/ROOT/pages/network-files.adoc
3h4x: Looks good to me!

ericglau (approved):
Thanks!

PierreJeanjacquot:
:100: I can't wait for this to be available in `@openzeppelin/hardhat-upgrades`

ericglau:
@PierreJeanjacquot This is available in [@openzeppelin/upgrades-core@1.33.1](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fupgrades-core%401.33.1) which is a dependency of `@openzeppelin/hardhat-upgrades`.

You can get this by updating your lock file, or by directly installing `@openzeppelin/upgrades-core@latest`

PR #1013, opened 2024-04-18
Defender: Improve license type handling

ericglau:
For Hardhat, fixes the same issue identified in https://github.com/OpenZeppelin/defender-deploy-client-cli/issues/8
Equivalent to https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades/pull/43
Updates submodule for latest docs from https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades/pull/43

Contracts have SPDX license identifiers, while block explorer verification (e.g. Etherscan source code verification) through Defender SDK requires a format that matches with https://etherscan.io/contract-license-types.

This PR does the following:
- Infers `licenseType` based on SPDX license identifier, instead of using the SPDX license identifier directly.
- Adds a `licenseType` option which can override the SPDX license identifier.
- Adds a `skipLicenseType` option which can skip setting a license type.

**Breaking change:**
- If the contract does not have an SPDX identifier, previously the verification sets the license type to be "None" on Etherscan.  With this PR, this case would result in Etherscan not showing the license type field at all.
  - If you want the license type to appear as "None", either set your contract to have `// SPDX-License-Identifier: UNLICENSED` according to [Solidity docs](https://docs.soliditylang.org/en/latest/layout-of-source-files.html#spdx-license-identifier), or set the `licenseType` option to `"None"`.


ernestognw (approved):
Looks good!

PR #1029, opened 2024-05-29
Defender: Fix proxy deployments when using `constructorArgs`

### LINKED ISSUE #1026: Defender DeployProxy not working when using `constructorArgs`
Trying to deploy an UUPS upgradeable proxy through Defender where the implementation contract has a constructor identified with `@custom:oz-upgrades-unsafe-allow`, throws the following error:

image

I tried to do this deployment without Defender and it works fine, but as I want to include this deployment behind an approval process mechanism, I'm trying to deploy it using Defender. Bear in mind that on my constructor I've a param requirement, so I have to execute the deployment using the following code:

```typescript
const proxy = await defender
    .deployProxy(factory, [upgradeApprovalProcess.address], {
        initializer: "initialize",
        kind: "uups",
        constructorArgs: [FORWARDER_ADDRESS],
        salt,
      })
    .then(async (f) => await f.waitForDeployment());
```

Behind the scenes it seems that [defenderDeploy](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/master/packages/plugin-hardhat/src/utils/deploy.ts#L18C1-L24C49) and [ethersDeploy](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/master/packages/plugin-hardhat/src/utils/deploy.ts#L24) have a difference that makes me to have a successful work when using the second one. Apparently, [defenderDeploy is overriding the `constructorArgs` required by the proxy deployment with the `constructorArgs`](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/master/packages/plugin-hardhat/src/defender/deploy.ts#L66C60-L66C88) that were used for the implementation deploy and causing the issue.

Just in case, on my constructor I'm sending the address of an [upgradeable forwarder](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/metatx/ERC2771ForwarderUpgradeable.sol) that I use to relay transactions.

**Any workaround for this scenario?**

ericglau:
Fixes #1026 

Defender deployments are run using the internal `defenderDeploy` function, which takes constructor arguments directly using the `...args` argument.  This function can deploy proxies or implementations.

However, users may provide `constructorArgs` to specify constructor args for the implementation contract.  These should not apply to the proxy deployment itself.

Therefore, the internal `defenderDeploy` function should use its `args` to override the ones in `opts.constructorArgs`, so that the correct arguments are used for each specific deployment.

### DIFF
--- a/packages/plugin-hardhat/src/defender/deploy.ts
+++ b/packages/plugin-hardhat/src/defender/deploy.ts
@@ -67,7 +67,6 @@ export async function defenderDeploy(
   // The ones in the options are for implementation contracts only, while this function
   // can be used to deploy proxies as well.
   const contractInfo = await getContractInfo(hre, factory, { ...opts, constructorArgs: args });
-  const constructorBytecode = contractInfo.constructorBytecode;
   const network = await getNetwork(hre);
   debug(`Network ${network}`);
 
@@ -107,7 +106,7 @@ export async function defenderDeploy(
     network: network,
     artifactPayload: JSON.stringify(contractInfo.buildInfo),
     licenseType: licenseType,
-    constructorBytecode,
+    constructorBytecode: contractInfo.constructorBytecode,
     verifySourceCode: verifySourceCode,
     relayerId: opts.relayerId,
     salt: opts.salt,
--- a/packages/plugin-hardhat/src/defender/deploy.ts
+++ b/packages/plugin-hardhat/src/defender/deploy.ts
@@ -66,6 +66,7 @@ export async function defenderDeploy(
   // The ones in the options are for implementation contracts only, while this function
   // can be used to deploy proxies as well.
   const contractInfo = await getContractInfo(hre, factory, { ...opts, constructorArgs: args });
+  const constructorBytecode = factory.interface.encodeDeploy(args);
   const network = await getNetwork(hre);
   debug(`Network ${network}`);
 
@@ -105,7 +106,7 @@ export async function defenderDeploy(
     network: network,
     artifactPayload: JSON.stringify(contractInfo.buildInfo),
     licenseType: licenseType,
-    constructorInputs: args as any[],
+    constructorBytecode,
     verifySourceCode: verifySourceCode,
     relayerId: opts.relayerId,
     salt: opts.salt,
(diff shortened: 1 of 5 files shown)


packages/plugin-hardhat/src/defender/deploy.ts
```diff
@@ -103,7 +105,7 @@ export async function defenderDeploy(
     network: network,
     artifactPayload: JSON.stringify(contractInfo.buildInfo),
     licenseType: licenseType,
-    constructorInputs: constructorArgs,
+    constructorInputs: args as (string | number | boolean)[],
```
Amxx, line 108: It looks like this that doesn't support constructor argument that are arrays or structs

For example
```solidity
constructor(address[] memory admins) {
    for (uint256 i = 0; i < admins.length; ++i) _grantRole(DEFAULT_ADMIN_ROLE, admins[i]);
}
```

Imo this should be 
```suggestion
    constructorInputs: args as any[],
```

ericglau: Defender SDK requires the former type, but this does look more correct and I tested that your constructor example works through Defender.

MCarlomagno: Hi, just adding context on this, I remember we had some conversations with @ericglau about how to handle higher order types as constructor args in defender deployments.
The main issue was that there was no easy way to serialize arrays or structs objects, so we decided to support `constructorBytecode` in our API for those cases, meaning that in cases of arrays or higher order types, we delegate the conversion to bytecode in the client side
Here an example of that:
https://github.com/OpenZeppelin/defender-sdk/blob/main/examples/deploy-contract/index.js#L37

ericglau: @MCarlomagno Thanks for the context. Based on this, we should switch the Hardhat plugin to encode the calldata first, and pass it through using `constructorBytecode` to avoid this limitation.  We also use `constructorBytecode` [in Foundry](https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades/blob/main/src/internal/DefenderDeploy.sol#L84).

Amxx (approved):

packages/plugin-hardhat/src/defender/deploy.ts
MCarlomagno: Yes! Correct

ericglau:
Updated to pre-encode constructor data and use `constructorBytecode` according to comments above.

MCarlomagno (approved):
LGTM

PR #1032, opened 2024-06-05
Fix storage layout comparison for function types, disallow internal functions in storage

### LINKED ISSUE #555: Disallow internal function types in storage
Internal functions are code pointers which will no longer be valid after an upgrade, because the code will move around and the pointer would change. External functions are still ok.

See https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/529#issuecomment-1090538356

### LINKED ISSUE #1031: An error due to function defined in struct in library
This is my library 

    struct Array {
        Elem[] _elems;
        function(Elem memory, Elem memory) internal pure returns (ebool) _Compare;
        function(ebool, Elem storage, Elem storage) internal _SwapIf;
    }

This is an error

```
Error: An unexpected condition occurred. Please report this at https://zpl.in/upgrades/report
      at assert (node_modules/.pnpm/@openzeppelin+upgrades-core@1.33.1/node_modules/@openzeppelin/upgrades-core/src/utils/assert.ts:7:11)
      at StorageLayoutComparator.getVisibilityChange (node_modules/.pnpm/@openzeppelin+upgrades-core@1.33.1/node_modules/@openzeppelin/upgrades-core/src/storage/compare.ts:194:11)
      at StorageLayoutComparator.uncachedGetTypeChange (node_modules/.pnpm/@openzeppelin+upgrades-core@1.33.1/node_modules/@openzeppelin/upgrades-core/src/storage/compare.ts:287:19)
      at StorageLayoutComparator.getTypeChange (node_modules/.pnpm/@openzeppelin+upgrades-core@1.33.1/node_modules/@openzeppelin/upgrades-core/src/storage/compare.ts:273:27)
      at StorageLayoutComparator.getFieldChange (node_modules/.pnpm/@openzeppelin+upgrades-core@1.33.1/node_modules/@openzeppelin/upgrades-core/src/storage/compare.ts:208:12)
      at /Users/lhp/git-project/CommodityTrading/node_modules/.pnpm/@openzeppelin+upgrades-core@1.33.1/node_modules/@openzeppelin/upgrades-core/src/storage/compare.ts:152:63
      at change (node_modules/.pnpm/@openzeppelin+upgrades-core@1.33.1/node_modules/@openzeppelin/upgrades-core/src/levenshtein.ts:86:20)
      at buildMatrix (node_modules/.pnpm/@openzeppelin+upgrades-core@1.33.1/node_modules/@openzeppelin/upgrades-core/src/levenshtein.ts:55:29)
      at levenshtein (node_modules/.pnpm/@openzeppelin+upgrades-core@1.33.1/node_modules/@openzeppelin/upgrades-core/src/levenshtein.ts:25:18)
      at StorageLayoutComparator.layoutLevenshtein (node_modules/.pnpm/@openzeppelin+upgrades-core@1.33.1/node_modules/@openzeppelin/upgrades-core/src/storage/compare.ts:152:28)
```

ericglau:
This fixes 2 things:

1. Fixes #1031.  In storage layout comparisons, type changes involving function pointers caused an error because of an incorrect assertion.  The original and updated types could be unrelated during the comparison, but `getVisibilityChange` was asserting that both are functions.  We should only call `getVisibilityChange` if both the original and updated types are indeed functions.

2. Fixes #555.  Internal functions are code pointers which will no longer be valid after an upgrade, so they should not be allowed.  Added this as a validation error, and allow it to be skipped with the `unsafeAllow.internal-function-storage` option (if skipping, users need to make sure they always reassign the internal functions in storage variables during each upgrade).


aungbu17 (approved):

cairoeth (commented):
mostly looks good, but shouldn't it also check for internal functions as a storage variable? for example:

```solidity
pragma solidity ^0.8.24;

contract StorageUpgrade_FunctionType_Visibility_V2_Bad {
    function () internal m;
}
```

function `m` will break after an upgrade as it gets moved around storage, no? similar to internal functions in structs

ernestognw (approved):
Very interesting finding!

I wonder if we should document how to reset the internal function pointers in the docs. It's true we now have a clear instruction within the CLI but we're not providing a concrete example other than the one shared in #1031.

What do you think?

ernestognw:
> mostly looks good, but shouldn't it also check for internal functions as a storage variable? for example:
> 
> ```solidity
> // SPDX-License-Identifier: MIT
> pragma solidity ^0.8.24;
> 
> contract StorageUpgrade_FunctionType_Visibility_V2_Bad {
>     function () internal m;
> }
> ```
> 
> function `m` will break after an upgrade as it gets moved around storage, no? similar to internal functions in structs

Functions do not live in storage, so `m` will not change. Instead, its pointer will change.
It's fine that it changes given that an upgrade changes the implementation (i.e. the runtime code and its internal function pointers).

The issue with storing internal function pointers in storage is that after upgrading to another implementation, the pointer in storage is a pointer to code that isn't used anymore. That's why it needs reinitialization.

aungbu17 (approved):

ericglau:
Agree we should check storage variables that have function types even outside of structs, and will add a section in the docs.

cairoeth:
> > mostly looks good, but shouldn't it also check for internal functions as a storage variable? for example:
> > ```solidity
> > // SPDX-License-Identifier: MIT
> > pragma solidity ^0.8.24;
> > 
> > contract StorageUpgrade_FunctionType_Visibility_V2_Bad {
> >     function () internal m;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > function `m` will break after an upgrade as it gets moved around storage, no? similar to internal functions in structs
> 
> Functions do not live in storage, so `m` will not change. Instead, its pointer will change. It's fine that it changes given that an upgrade changes the implementation (i.e. the runtime code and its internal function pointers).
> 
> The issue with storing internal function pointers in storage is that after upgrading to another implementation, the pointer in storage is a pointer to code that isn't used anymore. That's why it needs reinitialization.

Yup, was referring to the pointer changing when upgrading to a different implementation.

aungbu17 (approved):

packages/core/src/validate/run.ts
```diff
@@ -557,3 +560,32 @@ function* getLinkingErrors(
...
+      yield {
+        kind: 'internal-function-storage',
+        name: variableDef.name,
+        src: decodeSrc(variableDef),
+      };
+    }
+  }
+
+  // Note: Solidity currently does not include struct annotations in the AST, so this cannot be skipped with annotations
+  for (const structDef of findAll('StructDefinition', contractDef)) {
```
frangio, line 580: Pretty sure this loop is redundant. `StructDefinition.members` is `VariableDeclaration[]` so each member should be covered by the `findAll` loop above.

ericglau: This makes me realize the `findAll` usage above would miss structs that are declared outside of the contract.

We may need to look for all storage variables which use structs, and recursively check their referenced structs for function types.

I think we should still continue to look for struct declarations within contracts though, because for example namespaced storage structs could define these and not be referenced by regular storage variables.

frangio: > We may need to look for all storage variables which use structs, and recursively check their referenced structs for function types.

Yes this sounds right.

ericglau:
Code now handles https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/1032#discussion_r1631864200. 
Still need to add documentation.

docs/modules/ROOT/pages/faq.adoc
```diff
@@ -218,26 +217,90 @@ Renaming a variable is disallowed by default because there is a chance that a re
...
 }
-```
+----
 
 Docstring comments don't yet work for struct members, due to a current Solidity limitation.
+
+[[internal-function-storage]]
+== How can I use internal functions in storage variables?
+
+Internal functions in storage variables are code pointers which will no longer be valid after an upgrade, because the code will move around and the pointer would change. To avoid this issue, you can either declare those functions as external, or reassign internal functions during upgrades.
```
frangio, line 249: Honestly this would probably be better practice:

```suggestion
Internal functions in storage variables are code pointers which will no longer be valid after an upgrade, because the code will move around and the pointer would change. To avoid this issue, you can either declare those functions as external, or reassign internal functions during upgrades. Alternatively, avoid code pointers in storage altogether, and define an `enum` that you will use to select with a dispatcher function to select from the list of available functions.
```

Note sure if the description is clear enough like this.

packages/core/src/validate/run.ts
```diff
@@ -562,17 +572,25 @@ function* getLinkingErrors(
...
-  for (const variableDec of findAll('VariableDeclaration', contractDef)) {
+  for (const variableDec of findAll('VariableDeclaration', contractOrStructDef)) {
     if (variableDec.typeName?.nodeType === 'FunctionTypeName' && variableDec.typeName.visibility === 'internal') {
+      // Find internal function types directly in this node's scope
       yield {
         kind: 'internal-function-storage',
         name: variableDec.name,
         src: decodeSrc(variableDec),
       };
+    } else if (variableDec.typeName?.nodeType === 'UserDefinedTypeName') {
```
frangio, line 589: If a variable is a mapping we would want to inspect the contents of the mapping value. The same for arrays and so on. I think we need a recursive traversal through the type.

docs/modules/ROOT/pages/faq.adoc
ericglau: Is there any reason for the specific recommendation to use an `enum`, or would any implementation of a dispatcher function be suitable for this scenario?

frangio: An enum is just the classic way you would implement this and I think it's clean, but any type would work!

packages/core/src/validate/run.ts
```diff
@@ -586,12 +593,33 @@ function* getInternalFunctionStorageErrors(
...
+function findUserDefinedType(typeName: TypeName): UserDefinedTypeName | undefined {
+  switch (typeName.nodeType) {
+    case 'ArrayTypeName':
+      return findUserDefinedType(typeName.baseType);
+    case 'Mapping':
+      // only mapping values can possibly refer to other array, mapping, or user-defined types
+      return findUserDefinedType(typeName.valueType);
+    case 'UserDefinedTypeName':
+      return typeName;
+    default:
```
frangio, line 622: To make this more robust to future language changes:

```suggestion
    case 'ElementaryTypeName':
    case 'FunctionTypeName':
```

frangio (approved):
LGTM! Minor suggestion.

Issue #1037, opened 2024-06-17
Contract in `@openzeppelin/contracts-upgradeable` raises an `internal-function-storage` error


clemsos:
After upgrading to the latest version of the upgrades plugin, a `internal-function-storage` error is thrown with contract from the `@openzeppelin/contracts-upgradeable` lib. Here is the line:

https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/e49c574f71d402bbd5905175ec2205af02e87dda/contracts/governance/utils/VotesUpgradeable.sol#L259-L262

The error 

```
@openzeppelin/contracts-upgradeable5/governance/utils/VotesUpgradeable.sol:261: Variable `op` is an internal function
    Use external functions or avoid functions in storage.
     If you must use internal functions, skip this check with the `unsafeAllow.internal-function-storage`
     flag and ensure you always reassign internal functions in storage during upgrades
    https://zpl.in/upgrades/error-009
      at assertUpgradeSafe (/home/unlock/node_modules/@openzeppelin/upgrades-core/src/validate/query.ts:20:11)
      at validateImpl (node_modules/@openzeppelin/hardhat-upgrades/src/utils/validate-impl.ts:42:20)
      at validateProxyImpl (node_modules/@openzeppelin/hardhat-upgrades/src/utils/validate-impl.ts:63:10)
      at deployProxyImpl (node_modules/@openzeppelin/hardhat-upgrades/src/utils/deploy-impl.ts:72:3)
      at Proxy.deployProxy (node_modules/@openzeppelin/hardhat-upgrades/src/deploy-proxy.ts:48:28)
      at Context.<anonymous> (test/votes.js:36:10)
```

What is the best approach to silence the warning ? I would prefer to not disable the check for the entire upgrade if possible.

Many thanks !

ericglau:
This is a false positive, as the check was intended to give the error if there is an internal function pointer in storage variables.  In this case, internal function pointers are used directly from the implementation code and not from storage variables.

We are working on a fix for the validations, but for now you can disable the `internal-function-storage` check when calling the Upgrades plugin's functions according to
https://docs.openzeppelin.com/upgrades-plugins/1.x/faq#how-can-i-disable-checks

ericglau:
[...]

PR #1038, opened 2024-06-18
Avoid validation error when function parameter has internal function pointer

ericglau:
**Problem:**
The following function 
https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/e49c574f71d402bbd5905175ec2205af02e87dda/contracts/governance/utils/VotesUpgradeable.sol#L259-L262 causes the Upgrades plugin to give a validation error: Variable `op` is an internal function

This appears to be a false positive, since the internal function pointer comes from the same implementation code and is not kept in storage.

**Background:**
In https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/1032, we added validation errors for internal function pointers in storage variables.

This was done by using solidity-ast's `findAll` to look for all transitive nodes of type `VariableDeclaration` within a `ContractDefinition`.

The goal was to check all direct variables of a contract, or all variables within structs.  But the contract's functions' parameters can also have `VariableDeclaration` type (which may be function pointers), and these are also transitive children of the contract -- these should be ignored.

**Fix:**
Instead of checking all transitive `VariableDeclaration` nodes from a contract, only check those from its direct variables or from its structs.

Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1037


packages/core/src/validate/run.ts
```diff
@@ -608,6 +609,33 @@ function* getInternalFunctionStorageErrors(
...
+): VariableDeclaration[] {
+  const results: VariableDeclaration[] = [];
+  if (contractOrStructDef.nodeType === 'ContractDefinition') {
+    for (const node of contractOrStructDef.nodes) {
+      if (node.nodeType === 'VariableDeclaration') {
+        results.push(node);
+      } else if (node.nodeType === 'StructDefinition' && !visitedNodeIds.has(node.id)) {
+        visitedNodeIds.add(node.id);
+        results.push(...getVariableDeclarations(node, visitedNodeIds));
+      }
```
frangio, line 627: Is this necessary? It should be handled automatically by `getInternalFunctionStorageErrors`. Structs are not necessarily in storage, and those that are via EIP-7201 are already handled differently.

ericglau: I found this was necessary, otherwise this test case
```
contract StructInternalFunctionPointerUsed {
    struct S {
        function(bool) internal foo;
    }
    S s;
}
```
would give two identical errors: once for the struct def, and once for the struct variable referencing the struct def.

frangio: Right, I think this is consistent with what I'm saying. I would remove the lines that I commented on here. As far as I can tell that would only visit the struct via dereferencing the type of `s`.

ericglau: My above comment was for the need to track `visitedNodeIds`.  I now see what you mean by removing these lines, but I don't think that would work.

`getInternalFunctionStorageErrors` first calls `getVariableDeclarations` on a contract.
- For contracts with EIP-7201 structs, the lines that you commented on are needed.  If we removed these lines, the struct definitions would not be accessed (since there are no references)
- For contracts with variables that reference regular structs, this returns those variables and `getInternalFunctionStorageErrors` will handle dereferencing those structs.

Perhaps the code layout is somewhat confusing with the multiple stages of generators (`getVariableDeclarations` and `getInternalFunctionStorageErrors`) and the overloading of their argument `contractOrStructDef: ContractDefinition | StructDefinition`

frangio: Oh or is that only for layout purposes? I see.

frangio (approved):

PR #1046, opened 2024-07-16
Keep constants for namespaced compilation

### LINKED ISSUE #1045: [FATAL] Incorrect assumption of constant address as enum prevents compilation
With the following implementation, the package incorrectly assumes `address constant WETH_ADDRESS` as `type(enum WETH_ADDRESS)`. This issue only happens if the constant is used to instantiate a storage slot in a contract definition. Usage in other situations, like casting to IWETH or using the raw address inside functions work as intended.

Re-defining the constant inside the contract by using `address constant _WETH_ADDRESS = WETH_ADDRESS;` yields the same error.

Might be related to https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1016.

`./types/Constants.sol`
```
pragma solidity 0.8.25;

address constant WETH_ADDRESS = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
```

`./Gateway.sol`
```
pragma solidity 0.8.25;

import {WETH_ADDRESS} from "./types/Constants.sol";

contract Gateway is IGateway, UUPSUpgradeable, AccessControlUpgradeable {
  // ...

  IWETH public constant WETH = IWETH(WETH_ADDRESS);

  // ...
}
```

**Output**

```
An unexpected error occurred:

TypeError: Explicit type conversion not allowed from "type(enum WETH_ADDRESS)" to "contract IWETH".
  --> contracts/Gateway.sol:52:32:
   |
52 |   IWETH public constant WETH = IWETH(WETH_ADDRESS);
   |                                ^^^^^^^^^^^^^^^^^^^
```

**Package versions**

```
solc: 0.8.25
hardhat: 2.22.6
@openzeppelin/hardhat-upgrades: 3.2.0
```

ericglau:
When creating modified solc input for extracting namespaced storage layout, we remove variable declarations where possible to prevent them from referencing other nodes (such as functions which we have deleted).

By "removing", we don't really delete the actual variables, but only convert them to dummy enums to avoid orphaning any preceding NatSpec comments.

However, we want to avoid converting `constant`s to enums in case the constants are used in a struct, for example to allow this example to work:
```
int8 constant MAX_SIZE_C = 2;

contract StructArrayUsesConstant {
  uint16 private constant MAX_SIZE = 10;

  /// @custom:storage-location erc7201:uses.constant
  struct MainStorage {
    uint256 x;
    uint256[MAX_SIZE] y;
    uint256[MAX_SIZE_C] c;
  }
}
```

This PR changes the modifed solc input to keep all constants, and replaces functions with empty functions in case function signatures are referenced from constants.

Fixes #1045


packages/core/src/utils/make-namespaced.ts
```diff
@@ -61,8 +61,8 @@ export function makeNamespacedInput(input: SolcInput, output: SolcOutput): SolcI
           for (const contractNode of contractNodes) {
             switch (contractNode.nodeType) {
               case 'VariableDeclaration': {
-                // If variable is a constant, keep it since it may be referenced in a struct
-                if (contractNode.constant) {
+                // If variable is a constant int or uint, keep it since it may be referenced in a struct
+                if (contractNode.constant && contractNode.typeName?.nodeType === 'ElementaryTypeName' && (contractNode.typeName.name.startsWith('int') || contractNode.typeName.name.startsWith('uint'))) {
```
frangio, line 65: Let's define a helper function and use it below too?

frangio:
Here is some code that will present an issue with this logic:
```solidity
address constant a = address(0);
uint constant c = uint160(a);
```

This would be fixed by removing the restriction to int or uint types so that all constants are passed through. Are we aware of issues that that could cause?

For constant variables Solidity seems to accept the subset of expressions that only reference other constant variables and that don't invoke user functions (not even `pure` ones). So it seems to me that it should be fine to let them through. That said, I wouldn't be surprised if this restriction is relaxed at some point and then the strategy would need to be revised, but I don't see that happening soon.

ericglau:
@frangio Constants can also reference function selectors, such as:
```
bytes4 constant CONSTANT_USING_SELECTOR = MyContract.myFunction.selector;
````
And since we delete functions, this reference would be broken.  Perhaps instead of deleting functions, we can consider replacing them with dummy functions so that their selectors can still be referenced.

**Comment by frangio** [comment]
> Perhaps instead of deleting functions, we can consider replacing them with dummy functions

Or just erasing the body, is that what you had in mind? Yes I think this can work. I don't recall if we had tried this.

**Comment by ericglau** [comment]
Updated to reconstruct an empty function and keep all constants, but reconstructing the function does not look like the right approach.  Will simplify it further to just keep the existing function and remove body and modifiers as suggested above.

**Comment by frangio** [comment]
> keep the existing function and remove body and modifiers

I think there are two options. 1) Remove the bodies entirely so as to leave functions unimplemented, add `virtual` to all functions and `abstract` to all contracts. 2) Leave empty bodies, but to avoid compiler errors from uninitialized return values also remove all return types (`returns (...)`).

**Comment by frangio** [comment]
Here's an example that fails to compile with empty body:
```solidity
contract A {
    struct S { uint x; }
    function foo(S calldata s) internal pure returns (S calldata) {
        return s;
    }
}
```
The key issue is that a calldata pointer can't be default-initialized.

ericglau:
@frangio Ready for re-review. We can't use virtual since free functions can't be virtual. Also can't completely remove return types since the `returns` statement in `returns (...)` does not appear to be a node in Solidity's AST. So we replace return types with `(bool)` which can be default-initialized.

frangio (approved):
Looks good! Would be good to stress test this new approach on OZ Contracts.

PR #1048, opened 2024-07-19
Fix modified namespaced compilation when multiple return parameters are documented

### LINKED ISSUE #1047: `DocstringParsingError: Documentation tag exceeds the number of return parameters.`
In this commit https://github.com/pcaversaccio/escrow-contract/commit/f4aba711276840d3a3134d9457242fd6e30a687d I updated my play-around repo to the latest dependency versions and the Hardhat tests started to fail (see [here](https://github.com/pcaversaccio/escrow-contract/actions/runs/10005945231/job/27657634191#step:9:30)):

```console
An unexpected error occurred:

DocstringParsingError: Documentation tag "@return redeployProtectionFlag The 8-byte enum for the selection of a cross-chain redeploy protection." exceeds the number of return parameters.
   --> xdeployer/src/contracts/CreateX.sol:575:5:
    |
575 |     /**
    |     ^ (Relevant source part starts here and spans across multiple lines).



 ELIFECYCLE  Test failed. See above for more details.
Error: Process completed with exit code 1.
```

After investigation, I realised that via the `@tenderly/hardhat-tenderly` dependency, I also use `@openzeppelin/upgrades-core` which got upgraded as well to the latest `1.34.2` version. So the compilation issue is via my [`xdeployer`](https://github.com/pcaversaccio/xdeployer) Hardhat plugin, which also ships the [`CreateX`](https://github.com/pcaversaccio/createx) contract if you want to test local deployments.

Now the line that throws is [here](https://github.com/pcaversaccio/xdeployer/blob/300ff9f15b106d5dfeb76d0c71b01dc61186b54c/src/contracts/CreateX.sol#L914-L924):

```solidity
    /**
     * @dev Returns the enum for the selection of a permissioned deploy protection as well as a
     * cross-chain redeploy protection.
     * @param salt The 32-byte random value used to create the contract address.
     * @return senderBytes The 8-byte enum for the selection of a permissioned deploy protection.
     * @return redeployProtectionFlag The 8-byte enum for the selection of a cross-chain redeploy
     * protection.
     */
    function _parseSalt(
        bytes32 salt
    ) internal view returns (SenderBytes senderBytes, RedeployProtectionFlag redeployProtectionFlag) {
```

This was not an issue since the latest release. I don't see any DocString issue here since for each return value you use one `@return` tag.

ericglau:
When creating a modified copy of the solc input to extract namespaced storage layout, we replaced function return parameters with a single return parameter `(bool)`.  However, functions may have NatSpec that document multiple return parameters from the original function.  We should replace the return parameters with a list of `bool`s that corresponds to the same number of original return parameters. 

Fixes #1047


frangio:
I would add a test case like this:

```solidity
    /**
     * @return a first
     * @return b second
     */
    function bar() external returns (uint a, uint b) {}
```

It shouldn't cause the logic to fail because we replace the return parameters with unnamed ones. But you should see that when all names are present in the code the compiler does check that the first word after `@return` is the real parameter name and I'd want to guard against this getting stricter in future compiler versions.

frangio (approved):
Other than the above suggestion looks good to me though.

packages/core/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.34.3 (2024-07-19)
+
+- Fix Hardhat compile error when multiple return parameters are documented. ([#1048](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/1048))
```
ernestognw, line 5: Is it just Hardhat? I get #1047 got it through @tenderly/hardhat-tenderly because it depends on `@openzeppelin/upgrades-core`. I think compilation may be failing in Foundry too

ernestognw (approved):
Nit in changelog, but LGTM otherwise

packages/core/CHANGELOG.md
ericglau: This code path is currently only active when compiling in Hardhat.

ericglau: I think this happens because the above project enables `@tenderly/hardhat-tenderly` in hardhat.config.ts, and `@tenderly/hardhat-tenderly` imports `@openzeppelin/hardhat-upgrades` [here](https://github.com/Tenderly/hardhat-tenderly/blob/d97dc377fa9659942e8fe6b9e9add64d9fd4eb1f/packages/tenderly-hardhat/src/tenderly/extender.ts#L1).

So this means `@tenderly/hardhat-tenderly` effectively inherits Hardhat Upgrade's compilation subtask override, which adds the namespaced compilation step whenever Hardhat compile is run.

Issue #1053, opened 2024-07-25
Failed to compile modified contracts for namespaced storage, missing implementation


BladeRunner713:
Hardhat compilation error with OpenZeppelin upgrades plugin: Mismatch in function return type

## Description
When compiling my Solidity contracts using Hardhat, I encounter a compilation error that only occurs when the OpenZeppelin upgrades plugin is imported in the `hardhat.config.ts` file. The error suggests a mismatch in the return type of a function between the interface and its implementation.

## Steps to Reproduce
1. Create two Solidity files:
   - `IVersionableResolver.sol` (interface)
   - `ResolverAbs.sol` (abstract contract implementing the interface)
2. In `hardhat.config.ts`, import the OpenZeppelin upgrades plugin:
   ```typescript
   import '@openzeppelin/hardhat-upgrades';
3. Run `npx hardhat compile`

## Error message
`Note: Missing implementation:
 --> contracts/interfaces/IVersionableResolver.sol:7:5:
  |
7 |     function recordVersions(bytes32 node) external view returns (bool);`

Seems like the compiler recognizes the return value as bool, but the actual return value is uint64 

## Additional Information

- The compilation succeeds if the OpenZeppelin upgrades plugin import is removed from hardhat.config.ts.
- Solidity version: ^0.8.24
- Hardhat version: 2.22.6
- openzeppelin/hardhat-upgrades version: 3.2.0
- openzeppelin/contracts-upgradeable version: 5.0.2
- openzeppelin/upgrades-core: 1.34.4

   
**IVersionableResolver.sol**
  ```solidity
  // SPDX-License-Identifier: MIT
  pragma solidity ^0.8.24;
  
  interface IVersionableResolver {
      event VersionChanged(bytes32 indexed node, uint64 newVersion);
  
      function recordVersions(bytes32 node) external view returns (uint64);
  }
```

**ResolverAbs.sol**
  ```solidity
  // SPDX-License-Identifier: MIT
pragma solidity ^0.8.24;

import "@openzeppelin/contracts-upgradeable/utils/introspection/ERC165Upgradeable.sol";
import "../interfaces/IVersionableResolver.sol";

abstract contract ResolverAbs is ERC165Upgradeable, IVersionableResolver {
    // node -> node version
    mapping(bytes32 => uint64) public recordVersions;

    function isAuthorized(bytes32 node) internal view virtual returns (bool);

    modifier authorized(bytes32 node) {
[...]

PR #1055, opened 2024-07-30
Fix Hardhat compile error when public variables are used to implement interface functions

ericglau:
When modifying solc input to extract namespaced storage layouts, interface functions are kept (because constants might refer to their function selectors), but we effectively remove state variables (by converting them to dummy enums).

If an interface has a function, which is implemented using a state variable to provide a public getter, then this causes a compile error because the modified solc input (with state variable converted to dummy enum) would no longer have the public getter.

We can resolve this by converting all contracts in the modified solc input to `abstract`.

Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1053


ericglau:
Immutables are also effectively removed.  Since we are converting contracts to `abstract` now,
```

**Comment by ericglau** [comment]
Immutables are also effectively removed.  Since we are converting all contracts to `abstract` now, removing immutables also avoids the following from being an issue:
```
contract DeploysContractToImmutable {
    MyContract public immutable myContract = new MyContract();
}
```

Issue #1057, opened 2024-08-07
Enabling plugin causes valid contract to stop compiling


n0thingness:
Our investigations have lead us to conclude that recent releases of @openzeppelin/upgrades-core have introduced a bug when compiling contracts using hardhat

The contract in question

```
// SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.26;

contract Test {
    fallback(bytes calldata data) external returns (bytes memory) {
        return data;
    }
}
```

and the resulting compilation error

```
An unexpected error occurred:

Error: Failed to compile modified contracts for namespaced storage:

TypeError: Fallback function either has to have the signature "fallback()" or "fallback(bytes calldata) returns (bytes memory)".
   --> contracts/coverage/PanTest.sol:476:52:
    |
476 |     fallback(bytes calldata data) external returns (bool) {}
    |                                                    ^^^^^^
```

### Reproduction

1. `git clone git@github.com:n0thingness/hardhat-upgrades-bug-mvp.git`
2. `npm install`
3. `npx hardhat clean; npx hardhat compile;` There should be no error in compilation
4. `npm install --save-dev @openzeppelin/hardhat-upgrades`
5. Go to `hardhat.config.js` and uncomment line 2 to enable `require("@openzeppelin/hardhat-upgrades");`
6. `npx hardhat clean; npx hardhat compile;` Compilation fails with the above error

ericglau:
Thanks for reporting. This is fixed in [@openzeppelin/upgrades-core@1.35.1](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fupgrades-core%401.35.1)

PR #1060, opened 2024-08-12
Fix Hardhat compile error when `fallback` or `receive` functions are present

ericglau:
When making modified source code for namespaced storage layout extraction, we previously deleted constructors and kept functions (but removed the function bodies and changed the return parameters) in case the function selectors were referenced in other parts of the source code.

For `fallback` and `receive` functions, these must have specific signatures and they do not have selectors.  Therefore, we should delete them similar to constructors.

Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1057


packages/core/src/utils/make-namespaced.ts
```diff
@@ -228,28 +228,39 @@ function toDummyEnumWithAstId(astId: number) {
...
+      }
 
-    if (node.body) {
-      modifications.push(makeReplace(node.body, orig, '{}'));
+      break;
+    }
+    case 'constructor':
+    case 'fallback':
+    case 'receive': {
+      modifications.push(makeDelete(node, orig));
```
frangio, line 259: Should we be concerned about orphaned NatSpec? Or do we consider that a solved issue now with #1051?

packages/core/src/utils/make-namespaced.ts
frangio, line 1: I think the list of modifications in the docs for `makeNamespacedInput` needs to be updated.

packages/core/src/utils/make-namespaced.ts
ericglau: Changed to replace with dummy enums.  Orphaned NatSpec is indeed possible on platforms that don't support Slang.

frangio (approved):

Issue #1034, opened 2024-06-11
CLI: Support validations with references to other build info dirs


ericglau:
Users may want to validate storage layout compatibility between contracts in different versions (branches or tags) of a repo.  Each distinct version may have its own build-info dir.

Consider supporting multiple build-info dirs, and allow the `--reference` parameter or `@custom:oz-upgrades-from` annotation to specify the specific build-info dir as a further qualifier. 

Originally requested in https://forum.openzeppelin.com/t/storage-layout-verification-on-complex-projects/40773

PR #1062, opened 2024-08-16
CLI: Support `--referenceBuildInfoDirs` option

ericglau:
In some project setups, users of the [Upgrades Core CLI](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-core) may not want to change the names of upgradeable contracts for every new version that they deploy.  Instead, they just change the source code and want to perform storage layout comparisons with a previous version of their repository (e.g. from a different branch).

This PR provides a `--referenceBuildInfoDirs` option that takes a comma-separated list of additional build info directories that can be referenced for storage layout comparisons.  These build info directories can be built from previous versions of the project repository, containing the build artifacts (including storage layouts) of previously deployed versions of contracts.

When using this option, refer to one of these directories using prefix `<dirName>:` before the contract name or fully qualified name in the `--reference` option or `@custom:oz-upgrades-from` annotation, where `<dirName>` is the directory short name. Each directory short name must be unique.

Example usage:
1. From an older branch containing previously deployed versions of contracts, [compile the contracts with storage layouts](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-core#compile_contracts_with_storage_layouts) and save a copy of the resulting build-info directory (defaults are `artifacts/build-info` for Hardhat projects or `out/build-info` for Foundry projects), renaming it to a unique name.  For example, `build-info-v1`.
2. From the current branch containing the new versions of contracts that you want to upgrade to, [compile the contracts with storage layouts](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-core#compile_contracts_with_storage_layouts) for the new versions.
3. If you want to compare the storage layout of a contract called `MyContract` with its previous version, run the CLI command like:
```
npx @openzeppelin/upgrades-core validate --referenceBuildInfoDirs /old-builds/build-info-v1 --contract MyContract --reference build-info-v1:MyContract
```

Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1034


ericglau:
Here is a local development build that I packaged locally, in case anyone wants to try this out.
[openzeppelin-upgrades-core-v1.36.0.tgz](https://github.com/user-attachments/files/16641754/openzeppelin-upgrades-core-v1.36.0.tgz)
For testing purposes, you can install this in your project dependencies using a file url, e.g.
```
"@openzeppelin/upgrades-core": "file:///my-dir/openzeppelin-upgrades-core-v1.36.0.tgz",
```
Then follow the example in the description.

packages/core/src/cli/validate.ts
```diff
@@ -151,6 +155,10 @@ function getUnsafeAllowKinds(unsafeAllow: string | undefined): ValidationError['
   return unsafeAllowTokens as errorKindsType[];
 }
 
+function getReferenceBuildInfoDirsArray(referenceBuildInfoDirs: string | undefined): string[] | undefined {
+  return referenceBuildInfoDirs !== undefined ? referenceBuildInfoDirs.split(/,+/) : undefined;
```
Amxx, line 159: Does that work with typescript type system ?

```suggestion
  return referenceBuildInfoDirs?.split(/,+/);
```

Amxx (approved):
Lot to look at, and difficult to keep track when you are not familiar with the CLI's code, but it looks good to me !

PR #1065, opened 2024-08-22
CLI: Support `--exclude` option

### LINKED ISSUE #849: CLI: UpgradeableBeacon incorrectly detected as upgradeable implementation
UpgradeableBeacon is incorrectly detected as an upgradable implementation because it has a function with signature `upgradeTo(address)`, which causes it to "look" like a UUPS implementation, even though this function is for the beacon itself and not related to UUPS.

Steps to reproduce:
1. Install https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/ using Foundry
2. Run upgrades core CLI
3. The following error occurs:
```
lib/openzeppelin-contracts/contracts/proxy/beacon/UpgradeableBeacon.sol:UpgradeableBeacon

      lib/openzeppelin-contracts/contracts/proxy/beacon/UpgradeableBeacon.sol:31: Contract `UpgradeableBeacon` has a constructor
          Define an initializer instead
          https://zpl.in/upgrades/error-001
      
      lib/openzeppelin-contracts/contracts/access/Ownable.sol:38: Contract `Ownable` has a constructor
          Define an initializer instead
          https://zpl.in/upgrades/error-001
```

### LINKED ISSUE #1035: CLI: Provide a way to skip certain contracts from validation
CLI performs validation on contracts that look upgradeable, according to the rules described in https://docs.openzeppelin.com/upgrades-plugins/1.x/api-core#define_upgradeable_contracts.

Some contracts may be incorrectly detected as upgradeable, or users may want to omit certain contracts from validation.

Provide a way to skip certain contracts from validation, perhaps using a list, glob, or regex style pattern.

Related to https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/849
Originally requested in https://forum.openzeppelin.com/t/skipping-upgrades-validation-for-node-modules/40789

ericglau:
Adds an `--exclude "<GLOB_PATTERN>" [--exclude "<GLOB_PATTERN>"...]` option to the CLI, which allows glob patterns of source file paths that should be excluded from the validation report.  If passing in multiple patterns, call the option multiple times, once for each pattern.

This supports any of the user's source files or their dependencies.  Note this only supports file paths (e.g. the paths of `.sol` files), not the names of specific contracts in those files.

Additional notes:
- If the `--contract <CONTRACT_NAME>` option was specified and the file is excluded via the `--exclude` option, an error will be thrown.
- The pattern `**/contracts/proxy/beacon/UpgradeableBeacon.sol` is always added to the `exclude` list automatically, since `UpgradeableBeacon` contains an `upgradeTo(address)` function and is incorrectly detected as an implementation contract.
- This PR also updates the `--referenceBuildInfoDirs` and `--unsafeAllow` options to support those options being called either with a comma (as before) or multiple times (introduced in this PR).
  - However, `--exclude` does not treat comma as a delimiter, because globs can have commas within them e.g. `--exclude "**/{A,B}.sol"`

Fixes #1035
Fixes #849


packages/core/src/cli/validate/contract-report.ts
```diff
@@ -91,20 +96,43 @@ export function getContractReports(
...
+ * Gets a report for an upgradeable contract.
+ * Returns undefined if the contract is excluded or is abstract.
+ */
 function getUpgradeableContractReport(
   contract: SourceContract,
   referenceContract: SourceContract | undefined,
   opts: ValidationOptions,
+  exclude?: string[],
 ): UpgradeableContractReport | undefined {
+  const excludeWithDefaults = exclude !== undefined ? exclude.concat(defaultExclude) : defaultExclude;
```
Amxx, line 129: small suggestion
```suggestion
  const excludeWithDefaults = defaultExclude.concat(exclude ?? []);
```

Amxx (approved):
Left a very small suggestion. Looks good to me.

Issue #1076, opened 2024-09-11
Support metadata field for Defender deployments


ericglau:
Defender recently added support for a metadata field for deployments, as documented in https://docs.openzeppelin.com/defender/module/deploy#metadata

We should add this field as an option in the Hardhat and Foundry Upgrades plugins.

ericglau:
Completed in the above PRs

PR #1079, opened 2024-09-13
Defender: Add `metadata` option

ericglau:
Adds `metadata` option for Defender deployments in Hardhat Upgrades.

Part of #1076


MCarlomagno (approved):
LGTM, thanks!

PR #1083, opened 2024-09-20
Add check to ensure `initialOwner` is not a ProxyAdmin contract when deploying a transparent proxy

ericglau:
With OpenZeppelin Contracts v5, the `initialOwner` when deploying a transparent proxy must not be a predeployed ProxyAdmin contract.  The transparent proxy deploys its own ProxyAdmin using this `initialOwner` as the owner, therefore the `initialOwner` must be an EOA or a contract that is able to call functions on a ProxyAdmin.

This PR adds the following check: if the `initialOwner` when deploying a transparent proxy is a contract that has an `owner()` getter which returns something that looks like an address, it might be a ProxyAdmin contract.  In this case, throw an error.

It may have false positives if the `initialOwner` is some other contract (such as a smart account) which has an owner.  If this false positive occurs and the user is sure that the `initialOwner` is a contract that can invoke functions on a ProxyAdmin but is not a ProxyAdmin itself, this check can be skipped with the `unsafeSkipProxyAdminCheck` option.

Hardhat equivalent of https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades/pull/76


packages/core/src/infer-proxy-admin.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  } as EthereumProvider;
+}
+
+test('inferProxyAdmin returns true when owner looks like an address', async t => {
+  // abi encoding of address 0x1000000000000000000000000000000000000123
+  const provider = makeProviderReturning('0x0000000000000000000000001000000000000000000000000000000000000123');
+  t.true(await inferProxyAdmin(provider, addr));
+});
+
+test('inferProxyAdmin returns false when owner is more than 20 bytes', async t => {
```
Amxx, line 25:
```suggestion
test('inferProxyAdmin returns false when owner is more than 32 bytes', async t => {
```

ericglau: This does return false if the result is more than 20 bytes though.  When [parsing to determine whether it is an address](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/1083/files#diff-3cea14ff2284ad223d8822435a111d3b7b7ed158001adfdc652a51df94ca1bd1R11), we check the left part of the hex string to ensure the higher order bytes (beyond 20 bytes) are 0.

ericglau: Clarified this in commit https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/1083/commits/caf7553715191c75e481b34fc1f030d7ab966a69

ericglau: This test case is when some of the leftmost 12 bytes are non-zero, so is clearly not a padded address.  Updated as follows to make this clearer:
```js
test('inferProxyAdmin returns false when returned value is 32 bytes but clearly not an address', async t => {
  // some higher order bytes beyond 20 bytes are non-zero (the 'abc' in the below)
  const provider = makeProviderReturning('0x000000000000000000000abc1000000000000000000000000000000000000123');
  t.false(await inferProxyAdmin(provider, addr));
});
```

PR #1086, opened 2024-09-27
Fix Hardhat compile error when library or interface has namespace struct, print warning for library

### LINKED ISSUE #1066: hardhat-upgrades: Failed to compile a library that contains storage variable when using hardhat-upgrades plugin
When I try to compile a Hardhat project that contains a Solidity library, I get an unexpected error. The library seems to be valid solidity code, but the compile fails when the @openzeppelin/hardhat-upgrades plugin is enabled.

### Steps to Reproduce

1. Install the hardhat(v2.22.9) and @openzeppelin/hardhat-upgrades(v3.2.1) plugin.
2. Include the following solidity library and hardhat config in the project:
```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.20;

library Storage {
    // keccak256(abi.encode(uint256(keccak256("library.storageerror")) - 1)) & ~bytes32(uint256(0xff))
    bytes32 internal constant STORAGE_LOCATION =
        0x55d0c499112f0d58436b580033c012e14f7cc466c2e114bde6231acb85553000;

    /// @custom:storage-location erc7201:library.storageerror
    struct MainStorage {
        uint256 a;
    }

    function layout() internal pure returns (MainStorage storage $) {
        assembly {
            $.slot := STORAGE_LOCATION
        }
    }
}
```

```js
require("@openzeppelin/hardhat-upgrades");

/** @type import('hardhat/config').HardhatUserConfig */
module.exports = {
  solidity: "0.8.24",
};
```
3. `npx hardhat compile`

### Error message

```
$ npx hardhat compile
An unexpected error occurred:

TypeError: Library cannot have non-constant state variables
  --> contracts/Storage.sol:12:7:
   |
12 |     } MainStorage $MainStorage_751401;
   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^



```

ericglau:
This PR addresses two issues:

---

1. Fixes compilation error if library or interface contains namespace.

When making modified solc input to extract storage layouts, we normally detect namespace annotated structs in contracts and insert corresponding storage variables in order to get the detailed storage layouts of those structs.

However, libraries and interfaces do not support storage variables, so we should skip this insertion for libraries and interfaces.

(Note that if contracts inherit interfaces, namespaces in those interfaces can still be validated but they won't have detailed type information such as slots or lengths)

Fixes #1066 

---

2. Warn if library contains namespace.

ERC-7201 states, "Structs with this annotation found outside of contracts are not considered to be namespaces for any contract in the source code."

Storage layout validations can only occur on namespaces that we can detect in contracts and their inheritance (including inherited interfaces).  Particularly, namespaces in libraries are NOT validated, although we can see from [this GitHub query](https://github.com/search?q=%28path%3A*.sol%29+AND+%22library%22+AND+%22%40custom%3Astorage-location%22&type=code&ref=advsearch) that it is quite common.

This PR also adds a warning (not an error, to avoid breaking changes) if namespaces are defined in libraries.


frangio:
> (Note that if contracts inherit interfaces, namespaces in those interfaces can still be validated but they won't have detailed type information such as slots or lengths)

This is unfortunate. Have you considered any workarounds?

Can the interface be converted into a contract?

Amxx (approved):

PR #1090, opened 2024-10-04
Support ignoring Hardhat compile errors when extracting detailed namespaced storage layouts for validations

### LINKED ISSUE #1088: Allow namespaced compilation failures to be ignored
**Background**

When using Hardhat to compile, we create a modified copy of the solc input for namespaced storage layouts.  We then compile this modified input and extract detailed type information for namespaced struct members (such as slots and lengths) to use later for storage layout comparisons when the user attempts to upgrade a contract.

The detailed type information primarily allows for more complex modifications to pass validations.  For example, this allows upgrades to insert variables in unused space between existing variables (making use of tight storage packing), such as:
```
{
  uint128 x;
  uint256 y;
}
==>
{
  uint128 x;
  uint128 x1;
  uint256 y;
}
```
Detailed type information also allows for the consumption of storage gaps, but we don't expect storage gaps to be used inside namespaced structs.

Without detailed type information, only simpler modifications are allowed, for example adding variables to the end of the struct.

Note that currently only Hardhat Upgrades does the above.  Foundry Upgrades currently does not extract detailed type information for namespaces, but only has detailed type information for conventional layouts.

**Problem**

There have been a number of issues with Hardhat compilation failures when performing the above, due to various edge cases in users' contracts.  We have resolved them as they are reported, but future issues may still occur in additional edge cases or as we make further changes to how we modify the solc input.

**Proposed solution**

Since detailed type information for namespaces should not be a strict requirement but is a feature that allows for more complex modifications during upgrades, we should allow a way to ignore compilation errors from the modified solc input.

Any compilation errors typically indicate a bug in the modifications, rather than in the user's code.

It should be safe for users who are not going to make complex modifications in namespaces or are not even using namespaces to ignore these compilation errors.  We should still encourage them to open issues though.

Consider one of the following:
- Convert namespaced compilation errors into warnings by default, or
- Allow a flag to ignore namespaced compilation errors.

ericglau:
Fixes #1088

Support ignoring Hardhat compile errors when extracting detailed namespaced storage layouts for validations.

This can be done by setting the new option `namespacedCompileErrors` in the hardhat config, which supports `error` (default), `warn`, or `ignore`.

For example:
```
module.exports = {
  namespacedCompileErrors: 'warn',
};
```
or
```
module.exports = {
  namespacedCompileErrors: 'ignore',
};
```


ernestognw (approved):
Thanks @ericglau!

PR #1095, opened 2024-10-25
Check for initializer errors

### LINKED ISSUE #160: Detect issues in parent initializer calls
From [Slither's upgradeability checks](https://github.com/crytic/slither/wiki/Upgradeability-Checks#initialize-functions-are-not-called).

We can try to detect issues with parent initializer calls, concretely whether they are 1) missing, 2) duplicate, or 3) incorrectly linearized.

The main question is how we detect what is an initializer function. We could look for an `initializer` modifier, or for functions with name `initialize`, or for the naming convention used by Contracts Ethereum Package (soon Contracts Upgrade Safe) which is something like `__ContractName_init`.

ericglau:
Draft PR to detect issues in parent initializer calls.

This infers functions as initializers if they have a modifier that is `initializer`, `reinitializer`, or `onlyInitializing`, or the function itself is named as `initialize`, `initializer`, `reinitialize`, or `reinitializer`.

This reports an error if any parent contract has an inferred initializer and any of the following are true for the (child) contract being validated:
 1. Missing initializer: This contract does not appear to have an initializer.
 2. Missing initializer call: This contract's initializer is missing a call to a parent initializer.
 3. Duplicate initializer call: This contract has duplicate calls to the same parent initializer function.
 4. Incorrect initializer order: This contract does not call parent initializers in the correct order.

Remaining:
- [ ] Add a way to ignore all initializer errors from options
- [ ] Documentation
- [ ] Improve error messages to include more context, such as what was the expected linearization order and what was found instead, or what the duplicate call was
- [ ] Improve error messages to have more detailed suggestions for user action.
- [ ] Update changelog and bump version

Limitations:
- This does not recursively check the initializer function, where its called functions may have calls to the parent initializer.  This could result in a false positive, but users could workaround this by using unsafe-allow `missing-initializer-call`

Fixes #160

### DIFF
--- a/packages/core/src/validate-initializers.test.ts
+++ b/packages/core/src/validate-initializers.test.ts
@@ -1,0 +1,106 @@
+import _test, { TestFn } from 'ava';
+import { artifacts } from 'hardhat';
+
+import {
+  validate,
+  getContractVersion,
+  assertUpgradeSafe,
+  ValidationOptions,
+  RunValidation,
+  ValidationErrors,
+} from './validate';
+import { solcInputOutputDecoder } from './src-decoder';
+
+interface Context {
+  validation: RunValidation;
+}
+
+const test = _test as TestFn<Context>;
+
+test.before(async t => {
+  const contracts = ['contracts/test/ValidationsInitializer.sol:Parent_NoInitializer'];
+
+  t.context.validation = {} as RunValidation;
+  for (const contract of contracts) {
+    const buildInfo = await artifacts.getBuildInfo(contract);
+    if (buildInfo === undefined) {
+      throw new Error(`Build info not found for contract ${contract}`);
+    }
+    const solcOutput = buildInfo.output;
+    const solcInput = buildInfo.input;
+    const decodeSrc = solcInputOutputDecoder(solcInput, solcOutput);
+    Object.assign(t.context.validation, validate(solcOutput, decodeSrc));
+  }
+});
+
+function testValid(name: string, kind: ValidationOptions['kind'], valid: boolean, numExpectedErrors?: number) {
+  testOverride(name, kind, {}, valid, numExpectedErrors);
+}
+
+function testOverride(
+  name: string,
+  kind: ValidationOptions['kind'],
+  opts: ValidationOptions,
+  valid: boolean,
+  numExpectedErrors?: number,
+) {
+  if (numExpectedErrors !== undefined && numExpectedErrors > 0 && valid) {
+    throw new Error('Cannot expect errors for a valid contract');
+  }
+
+  const optKeys = Object.keys(opts);
+  const describeOpts = optKeys.length > 0 ? '(' + optKeys.join(', ') + ')' : '';
+  const testName = [valid ? 'accepts' : 'rejects', kind, name, describeOpts].join(' ');
+  test(testName, t => {
+    const version = getContractVersion(t.context.validation, name);
+    const assertUpgSafe = () => assertUpgradeSafe([t.context.validation], version, { kind, ...opts });
+    if (valid) {
+      t.notThrows(assertUpgSafe);
+    } else {
+      const error = t.throws(assertUpgSafe) as ValidationErrors;
+      if (numExpectedErrors !== undefined) {
+        t.is(error.errors.length, numExpectedErrors);
+      }
+    }
+  });
+}
+
+testValid('Child_Of_NoInitializer_Ok', 'transparent', true);
+
+testValid('Child_Of_InitializerModifier_Ok', 'transparent', true);
+testValid('Child_Of_InitializerModifier_Bad', 'transparent', false, 1);
+testValid('Child_Of_InitializerModifier_UsesSuper_Ok', 'transparent', true);
+
+testValid('Child_Of_ReinitializerModifier_Ok', 'transparent', true);
+testValid('Child_Of_ReinitializerModifier_Bad', 'transparent', false, 1);
+
+testValid('Child_Of_OnlyInitializingModifier_Ok', 'transparent', true);
+testValid('Child_Of_OnlyInitializingModifier_Bad', 'transparent', false, 1);
+
+testValid('MissingInitializer_Bad', 'transparent', false, 1);
+testValid('MissingInitializer_UnsafeAllow_Contract', 'transparent', true);
+testOverride('MissingInitializer_Bad', 'transparent', { unsafeAllow: ['missing-initializer'] }, true);
+
+testValid('InitializationOrder_Ok', 'transparent', true);
+testValid('InitializationOrder_Ok_2', 'transparent', true);
+
+testValid('InitializationOrder_WrongOrder_Bad', 'transparent', false, 1);
+testValid('InitializationOrder_WrongOrder_UnsafeAllow_Contract', 'transparent', true);
+testValid('InitializationOrder_WrongOrder_UnsafeAllow_Function', 'transparent', true);
+testOverride(
+  'InitializationOrder_WrongOrder_Bad',
+  'transparent',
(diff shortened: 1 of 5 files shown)


packages/core/src/validate-initializers.test.ts
```diff
@@ -0,0 +1,140 @@
...
+function testOverride(
+  name: string,
+  kind: ValidationOptions['kind'],
+  opts: ValidationOptions,
+  valid: boolean,
+  expectedErrorContains?: string,
+) {
+  if (expectedErrorContains !== undefined && valid) {
+    throw new Error('Cannot expect errors for a valid contract');
+  }
```
Amxx, line 49: I guess you could remove the `valid` boolean, and only use the expected error. It would be undefined for valid execution.

packages/core/src/validate/run.ts
```diff
@@ -629,6 +661,134 @@ function* getInternalFunctionStorageErrors(
...
+): Generator<ValidationErrorInitializer> {
+  if (contractDef.baseContracts.length > 0) {
+    const baseContractDefs = contractDef.baseContracts.map(base =>
+      deref('ContractDefinition', base.baseName.referencedDeclaration),
+    );
+    const baseContractsInitializersMap = new Map(
+      baseContractDefs.map(base => [base.name, getPossibleInitializers(base, true)]),
+    );
+    const baseContractsWithInitializers = baseContractDefs
+      .filter(base => hasInitializers(base.name, baseContractsInitializersMap))
```
Amxx, line 684: This should work, and would remove the need for `hasInitializers`
```suggestion
      .filter(base => baseContractsInitializersMap.get(base.name)?.length)
```

packages/core/src/validate/run.ts
```diff
@@ -629,6 +661,129 @@ function* getInternalFunctionStorageErrors(
...
+}
+
+function getPossibleInitializers(contractDef: ContractDefinition, isParentContract: boolean) {
+  const fns = [...findAll('FunctionDefinition', contractDef)];
+  return fns.filter(
+    fnDef =>
+      (fnDef.modifiers.some(modifier =>
+        ['initializer', 'reinitializer', 'onlyInitializing'].includes(modifier.modifierName.name),
+      ) ||
+        ['initialize', 'initializer', 'reinitialize', 'reinitializer'].includes(fnDef.name)) &&
```
ernestognw, line 778: For another PR, I think it may be useful if we annotate initializers with `@custom:initializer` or similar.
The idea is that there are contracts with initializers that don't follow the `initialize` or `initializer` convention in its name.

A concrete example is any of the Signers we created for ERC4337 accounts:

```solidity
abstract contract SignerRSA is AbstractSigner {
    error SignerRSAUninitializedSigner(bytes e, bytes n);

    bytes private _e;
    bytes private _n;

    /// @custom:initializer true
    function _initializeSigner(bytes memory e, bytes memory n) internal {
        if (_e.length != 0 || _n.length != 0) revert SignerRSAUninitializedSigner(e, n);
        _e = e;
        _n = n;
    }

...
}
```

packages/core/src/validate/run.ts
```diff
@@ -629,6 +661,129 @@ function* getInternalFunctionStorageErrors(
...
+                    parentInitializer: foundParentInitializer.name,
+                    parentContract: baseName,
+                  };
+                }
+                calledInitializerIds.push(referencedFn);
+
+                const index = uninitializedBaseContracts.indexOf(baseName);
+                if (
+                  !duplicate &&
+                  index !== 0 &&
```
ernestognw, line 736: If a userz disables the `duplicate-initializer-call` and there's a duplicate, we won't be yielding the `incorrect-initializer-order` check, is this correct?

Perhaps defining `duplicate` as follows would work:

```solidity 
const duplicate = calledInitializerIds.includes(referencedFn) &&
                  !skipCheck('duplicate-initializer-call', contractDef) &&
                  !skipCheck('duplicate-initializer-call', contractInitializer) &&
                  !skipCheck('duplicate-initializer-call', stmt)
```

ericglau: This is intentional, because `incorrect-initializer-order` should only be for things that are not duplicate (regardless of whether the duplicate was ignored).

For example if parent contracts are `A, B, C`, and the initializer calls were in the order of `A, B, B, C`, the third call could be seen as both a duplicate and incorrect order, because it does not match the expected order.  But we want to ignore it from being considered incorrect order in this case.

Even if the user disables `duplicate-initializer-call`, I think this case should still not be considered as incorrect order.

Added some comments and a separate testcase to clarify this.

ernestognw: I see what you mean, basically having a duplicate will already be considered out of order. So if a user disables `duplicate-initializer-call`, those would be out of order anyways.

Makes sense. Thanks for documenting further. Solving

ernestognw (approved):

ericglau:
The previous approach did not work for one of the main use cases that we want to detect: missing initializer calls to transitive parent initializers.

Here is an illustrating example: https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/330057ef1067d96e0bf3fee78016b6824db9be6d/packages/core/contracts/test/ValidationsInitializer.sol#L406 , which represents the issue in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4690.  Since `Ownable2StepUpgradeable` inherits `OwnableUpgradeable` but does not call its initializer, the user's contract must call it. 
 The previous approach would not detect this issue because it only checks the direct parent initializers.

Instead, I changed this to get the entire linearized base contracts list, remove the ones that are already initialized by some other parent in the list, and then check if the user's contract has an initializer that calls the remaining ones.

---

Remaining considerations:
- Duplicate initializer calls are currently not detected if they occur from different contracts (e.g. a grandparent initializer is called by both a parent and a child).  This is not necessarily unsafe.  Initializer usage is not a strict as constructors, and this cannot be reliably detected because users could have multiple initializers and/or reinitializers.  Perhaps we should avoid checking for duplicates altogether, or just leave the duplicate checks for be the same contract.
- Incorrect order checks might also be too strict.  Incorrect order of linearized contracts is not necessarily unsafe.  For example, in [this testcase](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/330057ef1067d96e0bf3fee78016b6824db9be6d/packages/core/contracts/test/ValidationsInitializer.sol#L360) where the parent is initialized before the grandparent, that is incorrect linearization but does not pose an actual issue.  Perhaps we should check linearization order only for initializer calls to direct parents.
- Simplify/cleanup code.

ericglau:
Remaining considerations:

- Should we check for duplicate initializer calls if they occur from different contracts?
  - For comparison, if similar constructor calls are duplicated, Solidity gives an error `Base constructor arguments given twice.`
  - Scenario:
```
abstract contract A {
  function __A_init(uint x) {...}
}

abstract contract B is A {
  function __B_init() {
     __A_init(1);
  }
}

contract C is B {
   function initialize() {
      __A_init(2); // Duplicate call
      __B_init();
   }
}
```
- Should we check for initializer order?
  - For comparison, Solidity always calls constructors in linearized order [regardless of the argument order](https://docs.soliditylang.org/en/latest/contracts.html#multiple-inheritance-and-linearization).
  - Scenario:
```
abstract contract A {
  function __A_init() {...}
}

abstract contract B is A {
  function __B_init() {
    // does NOT initialize A
  }
}

contract C is B {
   function initialize() {
      __B_init();
      __A_init(); // Wrong order
   }
}
```

ericglau:
Added recursive checks in the initializer function, to detect cases where its called functions (e.g. helper functions or parent initializer calls) may have duplicate calls or incorrect order of calls to parent initializers.

ericglau:
Ready for re-re-review! @Amxx @ernestognw

ernestognw (approved):

Issue #1082, opened 2024-09-17
deployProxy function does not return correct contract type


ryanc-bs:
I've encountered an issue using the deployProxy method with TypeScript.

When deployed without a proxy using `ethers.deployContract`, TypeScript is able to infer the correct type for the contract being deployed. For example, if I have a contract named `MyContract` with method `foo` then calling `ethers.deployContract` returns a `MyContract` instance with the required `foo` method.

This breaks when using `deployProxy` - instead of returning the correct contract type according to the factory type, it returns a generic `Contract` object which needs to be explicitly cast into the correct contract type from typechain-types.

Would it be possible to update `deployProxy` so that the correct type can be automatically inferred from the factory to avoid this ugly casting?

ericglau:
Related to https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/535. As that PR hasn't had recent activity, this may be open to contribution if anyone is interested in picking it up.

robertmagier:
I will do it

ericglau:
Fixed by #1099 and available in [@openzeppelin/hardhat-upgrades@3.8.0](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fhardhat-upgrades%403.8.0)

PR #1099, opened 2024-11-11
Support TypeChain in deploy and upgrade proxy functions - continuation of work done on #535

robertmagier:
Continuation of work done on #535 by frimoldi. 
See #325 and #1082.

Returns type safe contract types with: `deployProxy`, `deployContract`, `deployBeaconProxy`, and `upgradeProxy` functions.

This does not apply to `deployBeacon`, `upgradeBeacon`, and `forceImport` since those can return `UpgradeableBeacon` rather than the implementation contract type.


ericglau (approved):
Thanks a lot for picking this up and sorry for the late response.  This looks great!

I made a few changes as follows:
- `deployBeacon` and `upgradeBeacon` should return an `UpgradeableBeacon` instance, rather than the implementation contract's type.  We don't generate detailed TypeChain information for `UpgradeableBeacon` for now, so I've reverted those to just return `Contract`.
- `forceImport` could also return an UpgradeableBeacon if the given address is a beacon.  For simplicity, I've reverted that as well so it just returns `Contract`.
- changed the type definition of `ContractTypeOfFactory` to be `ReturnType<F['attach']> & ReturnType<F['deploy']>` since with Ethers v6, `attach` just returns a BaseContract which does not have the contract's type.  But TypeChain provides an override of `deploy` which gives us the expected type -- this seems to work. 

For reference, since our tests in this project uses JS, I created a sample TS project and tested it with these testcases: https://github.com/ericglau/hardhat-upgrades-typechain-tests/blob/main/test/test.ts

PR #1102, opened 2024-11-20
Upgrade slang to 0.18.3

ericglau:
Continuation of https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/1100

The `upgrades-core` package has a `makeNamespacedInput` function, where near the end of its processing, it uses Slang to try to sanitize the source code further to remove non-struct NatSpec.  This previously used Slang 0.17.

Slang 0.18.x supports WASM and must be imported using `await import()`.

To avoid converting the `makeNamespacedInput` function to async which would result in breaking changes, this PR instead removes the sanitization step from `makeNamespacedInput` and exposes an async `trySanitizeNatSpec` function that can be called separately.  We then call this new function in the `hardhat-upgrades` plugin.


packages/plugin-hardhat/src/index.ts
```diff
@@ -69,7 +69,9 @@ interface RunCompilerArgs {
 }
 
 subtask(TASK_COMPILE_SOLIDITY, async (args: { force: boolean }, hre, runSuper) => {
-  const { readValidations, ValidationsCacheOutdated, ValidationsCacheNotFound } = await import('./utils/validations');
+  const { readValidations, ValidationsCacheOutdated, ValidationsCacheNotFound } = await import(
+    './utils/validations.js'
```
ernestognw, line 73: Why are these `.js` required now?

packages/plugin-hardhat/src/index.ts
```diff
@@ -85,21 +87,21 @@ subtask(TASK_COMPILE_SOLIDITY, async (args: { force: boolean }, hre, runSuper) =
...
 
-  const { isFullSolcOutput } = await import('./utils/is-full-solc-output');
+  const { isFullSolcOutput } = await import('./utils/is-full-solc-output.js');
   if (isFullSolcOutput(output)) {
     const decodeSrc = solcInputOutputDecoder(args.input, output);
 
     let namespacedOutput = undefined;
     if (isNamespaceSupported(args.solcVersion)) {
       const namespacedInput = makeNamespacedInput(args.input, output, args.solcVersion);
+      await trySanitizeNatSpec(namespacedInput, args.solcVersion);
```
ernestognw, line 104: For consistency, I think `trySanitizeNatSpec` should return the modified input.

packages/core/src/utils/make-namespaced.ts
```diff
@@ -163,74 +163,75 @@ export function makeNamespacedInput(input: SolcInput, output: SolcOutput, solcVe
...
+ * @param solcInput Solc input.
+ * @param solcVersion The version of the solc compiler that was originally used to compile the input.
+ */
+export async function trySanitizeNatSpec(solcInput: SolcInput, solcVersion: string) {
+  for (const [sourcePath, source] of Object.entries(solcInput.sources)) {
+    if (source.content !== undefined) {
+      solcInput.sources[sourcePath].content = await tryRemoveNonStructNatSpec(source.content, solcVersion);
+    }
+  }
+}
```
ernestognw, line 187: Accordingly

```suggestion
export async function trySanitizeNatSpec(solcInput: SolcInput, solcVersion: string): Record<string, { content?: string }> {
  for (const [sourcePath, source] of Object.entries(solcInput.sources)) {
    if (source.content !== undefined) {
      solcInput.sources[sourcePath].content = await tryRemoveNonStructNatSpec(source.content, solcVersion);
    }
  }
  
  return solcInput;
}
```

I may be wrong with the types

packages/core/src/utils/make-namespaced.ts
```diff
@@ -163,74 +163,75 @@ export function makeNamespacedInput(input: SolcInput, output: SolcOutput, solcVe
...
 
-  if (solcVersion !== undefined && tryRequire('@nomicfoundation/slang') && slangSupportsVersion(solcVersion)) {
-    /* eslint-disable @typescript-eslint/no-var-requires */
-    const { Language } = require('@nomicfoundation/slang/language');
-    const { NonterminalKind, TerminalKind } = require('@nomicfoundation/slang/kinds');
-    const { TerminalNode } = require('@nomicfoundation/slang/cst');
-    const { isTrivia } = require('./slang/trivia');
-    /* eslint-enable @typescript-eslint/no-var-requires */
+  const { Parser } = await import('@nomicfoundation/slang/parser');
+  if (!Parser.supportedVersions().includes(solcVersion)) {
```
ernestognw, line 201: Is this breaking compatibility with older versions of Solidity? Perhaps we should mention it in the Changelog

ernestognw (commented):
I'll keep reviewing

ernestognw:
Should this be released under a minor?

packages/plugin-hardhat/src/index.ts
ericglau: This is because `moduleResolution` was changed to `nodenext` in tsconfig.base.json, in order to support importing Slang.

Omitting the `.js` extension here causes error `Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'.`

packages/core/src/utils/make-namespaced.ts
ericglau: On the contrary, this adds compatibility for a newer version of Solidity.  So I agree this can be in the changelog and can have a minor version bump.

packages/core/src/utils/make-namespaced.ts
ericglau: Reimplemented along with using the returned input in the callers.

packages/core/src/utils/make-namespaced.ts
ernestognw: Yes, let's expand a bit in the changelog entry to make it clear we're extending compatibility

Issue #231, opened 2020-11-19
Add support for custom proxy factory


thegostep:
I've been playing around with various types of proxies and there is a fair bit of friction to putting them in prod due to poor testing infra.

Proxy examples:
https://github.com/onboardmoney/solidity-version-proxy
https://github.com/0age/Spawner

It would be great to be able to use the upgrades package and pass in a custom proxy factory

frangio:
We will be adding a beacon proxy like Dharma's in OpenZeppelin Contracts soon, so this is something we're interested in as well.

kiliw:
Are there any updates so far? I'm playing around with the beacon proxy but I could not find any equivalents for `deployProxy` and `upgradeProxy` functions from the truffle upgrades plugin, which works with the beacon proxies.

frangio:
No updates on this yet. Some thoughts on how the implementation should proceed:

We will need to parameterize `deployProxy` and `upgradeProxy` with a structure describing how to deploy and interact with the custom proxy. We seem to need these operations:

- `deploy(admin: address, impl: address, data: bytes): Promise<address>`
- `getCurrentImplementation(proxy: address): address`
- `upgrade(proxy: address, newImpl: address)`

There may be another operation needed to check that an upgrade will be possible with the current provider signer (i.e. if it is the right admin for the proxy), to avoid deploying an implementation if the upgrade will not be able to proceed.

For beacon proxies it's kind of weird that upgrading a single one will end up upgrading multiple proxies, I wish there was some way to make this more explicit in the API...

gnarvaja:
I'm pushing for this change. I submitted PR #1104 with tests and the use case that this change will unlock for me.

ericglau:
Fixed by #1104 and available in [@openzeppelin/hardhat-upgrades@3.7.0](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fhardhat-upgrades%403.7.0)

PR #1104, opened 2024-11-28
Options to change deploy function and proxy contract factory

gnarvaja:
Implementation for #231 

Adds two new options that allow override of the deploy function and contract factory for the proxy.

Kind is still used to identify the kind of proxy used, but this
customization can manage variations of the proxy implementation and its
constructor parameters.

As a test case, and to show a concrete application of this change, I
implemented a AccessManagedProxy. This proxy works like a UUPS (ERC1967)
proxy, but before delegating the calls, checks with the access manager
if the call is allowed.


ericglau (commented):
Thank you @gnarvaja for this PR!  This looks like a good approach for allowing the proxy deployment to be overridden.

I made a few minor changes to the example proxy's comments and simplified the types for the new options.

The following would also be needed before we can merge this.  Can you please let me know if you would be interested in adding these?  Thanks!
- A basic test for deploying a custom beacon proxy
- Add the new options in the documentation in `docs/modules/ROOT/pages/api-hardhat-upgrades.adoc`

ericglau:
@gnarvaja I changed the `proxyFactory` option to directly take a ContractFactory instead of a function that returns the ContractFactory.  I think this makes it simpler to use.  Let me know if that works for you, then I can go ahead and release this. 
 Thanks again for your contributions here!

gnarvaja:
> @gnarvaja I changed the `proxyFactory` option to directly take a ContractFactory instead of a function that returns the ContractFactory. I think this makes it simpler to use. Let me know if that works for you, then I can go ahead and release this. Thanks again for your contributions here!

Yes, I agree with the change. Thanks!

ericglau (approved):

PR #1130, opened 2025-02-20
fix: update broken link foundry-upgrades.adoc

dramarereg:
The file `foundry-upgrades.adoc` contained a broken link to the Foundry documentation. The old link pointed to `tutorials/solidity-scripting`, which is deprecated. Updated it to the correct and current link: `guides/scripting-with-solidity`.


ericglau:
Thank you for the PR!

These files are copied from a submodule using the repository https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades, so should be updated there instead.

I've opened issue https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades/issues/96 to point out the two places which need to be updated, and will close this issue.  Feel free to open an equivalent PR in https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades , thanks.

Issue #1138, opened 2025-03-16
Correctly process Solidity custom storage layout


frangio:
Solidity 0.8.29 released a few days ago adds ["custom storage layout"](https://docs.soliditylang.org/en/latest/contracts.html#custom-storage-layout) via `layout at` annotations in contracts.

This affects the correctness of storage analysis done by OpenZeppelin Upgrades, so it should eventually be supported, but until that's implemented it should be detected and result in an error.

This is now available in `solidity-ast` [0.4.60](https://github.com/OpenZeppelin/solidity-ast/releases/tag/v0.4.60).

ericglau:
- If storageLayout is included in Solidity compiler output, which is expected to be the case with both Hardhat and Foundry, using a custom storage layout affects the slots of variables.  Changing the storage layout in the implementation during an upgrade causes validations errors due to slot changes.
- If validations are run directly without storageLayout, detect the presence of `ContractDefinition.storageLayout` and give an error for now, or give an error if we can determine that a different layout root is being used.
- When extracting type information for namespaced storage layout, we need to remove the layout annotation from the modified copy of the source code, otherwise the following error occurs:
```
TypeError: Storage layout cannot be specified for abstract contracts.
  --> contracts/MyTokenV2.sol:10:29:
   |
10 | abstract contract MyTokenV2 layout at 0x1 is Initializable, ERC20Upgradeable, OwnableUpgradeable, UUPSUpgradeable {
   |                             ^^^^^^^^^^^^^
```

ericglau:
- Also check for conflicts between a custom layout and any of the namespaces in use (!)
  - Ideally, use the namespace id to calculate the offset of each slot for namespaced variables, so they can be compared to normal variables that are using custom storage layout.
  - But for now, we check for a clash in the storage root location, similar to the current check for duplicate namespace ids.

frangio:
Regarding namespaces note that Solidity plans to support an `erc7201` function for use in `layout at` annotations (https://github.com/ethereum/solidity/issues/15727).

PR #1144, opened 2025-03-27
Validate Solidity custom storage layout, use OpenZeppelin Contracts 5.3.0

ericglau:
Solidity 0.8.29 adds ["custom storage layout"](https://docs.soliditylang.org/en/latest/contracts.html#custom-storage-layout) via `layout at` annotations in contracts.

Existing behavior (before this PR):
- If detailed storage layout information is included in Solidity compiler output, which is expected to be the case with both Hardhat and Foundry, using a custom storage layout affects the slots of variables in the detailed layout.  Changing the custom storage layout in the implementation during an upgrade causes validation errors due to slot changes.

This PR adds the following additional features/fixes:
- Reports an error if Solidity custom storage layout has the same base slot as an ERC-7201 namespace that is in use.
- Reports an error if detailed storage layout information (e.g. slot numbers for every variable) is not available in the Solidity output and the contract's custom storage layout's base slot changes during an upgrade.
- Warns that the checks between custom storage layout and namespaces only apply to the base slots.
- Fixes compile error when extracting type information for namespaced storage layout while `layout at` is included in the contract.
- Updates proxies to use OpenZeppelin Contracts 5.3.0 and Solidity version 0.8.29 to match with the latest project settings.

Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1138


packages/core/src/utils/integer-literals.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  let calculatedValue: bigint;
+  if (isScientific) {
+    const [base, exponent] = lowercaseNoUnderscores.split('e');
+    calculatedValue = BigInt(base) * BigInt(10) ** BigInt(exponent);
+  } else {
+    // works for hex and decimal
+    calculatedValue = BigInt(lowercaseNoUnderscores);
+  }
+
```
CoveMB, line 29: could literal be negative, should they be converted into 256 bit 2 components?

packages/core/src/storage/namespace.ts
```diff
@@ -140,6 +160,52 @@ function addContractNamespacesWithSrc(
...
+function checkCustomLayoutClashWithNamespace(
+  contractDef: ContractDefinition,
+  storageLocation: string,
+  contractName: string,
+  src: string,
+) {
+  if (contractDef.storageLayout !== undefined && storageLocation.startsWith('erc7201:')) {
+    const namespaceId = storageLocation.split(':')[1];
+    const storageLocationHash = calculateERC7201StorageLocation(namespaceId);
+    if (contractDef.storageLayout.baseSlotExpression.nodeType === 'Literal') {
```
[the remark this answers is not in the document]
ericglau: According to the solidity-ast types, `contractDef.storageLayout.baseSlotExpression` won't be undefined here.

packages/core/src/storage/namespace.ts
```diff
@@ -140,6 +160,52 @@ function addContractNamespacesWithSrc(
...
+ * @throws CustomLayoutClashWithNamespaceError if the custom layout clashes with the namespace.
+ */
+function checkCustomLayoutClashWithNamespace(
+  contractDef: ContractDefinition,
+  storageLocation: string,
+  contractName: string,
+  src: string,
+) {
+  if (contractDef.storageLayout !== undefined && storageLocation.startsWith('erc7201:')) {
+    const namespaceId = storageLocation.split(':')[1];
```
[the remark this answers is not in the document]
ericglau: Good catch, using split isn't the right approach since it could result in more than 2 segments if there are more colons.  Changed it to use substring.

packages/core/src/utils/integer-literals.ts
```diff
@@ -0,0 +1,33 @@
...
+    return '0x0000000000000000000000000000000000000000000000000000000000000000';
+  }
+  const lowercaseNoUnderscores = literal.replace(/_/g, '').toLowerCase();
+
+  const isHex = lowercaseNoUnderscores.startsWith('0x');
+  const isScientific = !isHex && lowercaseNoUnderscores.includes('e');
+
+  let calculatedValue: bigint;
+  if (isScientific) {
+    const [base, exponent] = lowercaseNoUnderscores.split('e');
```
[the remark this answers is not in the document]
ericglau: Good point, it could indeed be a decimal, and integer literal just means the final represented value must be integer.
Changed to use a library to handle parsing and added testcases.

packages/core/src/utils/integer-literals.ts
ericglau: The input would only be a valid unsigned integer that passed Solidity compilation.  Renamed the function to make this clear.

packages/core/contracts/test/CustomLayout.sol
```diff
@@ -0,0 +1,178 @@
...
+    uint256 public y;
+    uint256 public z;
+}
+
+contract CustomLayout_Same_Root_Decimal_Ok layout at 1 is Base {
+    uint256 public y;
+    uint256 public z;
+}
+
+contract CustomLayout_Same_Root_ScientificNotation_Ok layout at 1000.0e-3 is Base {
```
ernestognw, line 22: looool you can do this

packages/core/src/utils/make-namespaced.test.ts.md
```diff
@@ -688,3 +688,53 @@ Generated by [AVA](https://avajs.dev).
...
+      },
+      sources: {
+        'contracts/test/NamespacedToModifyCustomLayout.sol': {
+          content: `// SPDX-License-Identifier: MIT␊
+          pragma solidity ^0.8.29;␊
+          ␊
+          abstract contract Base {␊
+          }␊
+          ␊
+          abstract contract ExampleCustomLayout  is Base {␊
```
ernestognw, line 723: There seems to be a space left when the layout is removed. Not a big deal but pointing out if we want to change it

packages/core/src/storage/namespace.ts
```diff
@@ -140,6 +160,52 @@ function addContractNamespacesWithSrc(
...
+ */
+function checkCustomLayoutClashWithNamespace(
+  contractDef: ContractDefinition,
+  storageLocation: string,
+  contractName: string,
+  src: string,
+) {
+  if (contractDef.storageLayout !== undefined && storageLocation.startsWith(ERC7201_FORMULA_PREFIX)) {
+    const namespaceId = storageLocation.substring(ERC7201_FORMULA_PREFIX.length);
+    const storageLocationHash = calculateERC7201StorageLocation(namespaceId);
```
ernestognw, line 180: nit: this is not a hash (i.e. is `hash() & `~bytes32(uint256(0xff)`), so I would just call it `calculatedStorageLocation` or similar

ernestognw (approved):

packages/core/src/utils/make-namespaced.test.ts.md
ericglau: The modified contract is only used internally, and whitespace isn't part of the AST so it's not simple to remove and attempting to do so could be more prone to errors.

Issue #1122, opened 2025-02-10
Support custom annotation to assert that a function is an initializer


ericglau:
When validating errors in parent initializer calls, initializers are detected based on the function name or modifier.  

In cases where the user's initializer is named differently or has differently named modifiers, allow users to annotate that a function is an initializer. See https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/1095#discussion_r1890445928

Issue #1125, opened 2025-02-11
Error: `Define an initializer function and use it to call the initializers of parent contracts`


vladimir-trifonov:
I get 
```
      test/contracts/xxxV2.sol:20: Missing initializer
          Define an initializer function and use it to call the initializers of parent contracts
          https://zpl.in/upgrades/error-001
```
when I try to upgrade from v1 to v2 of the smart contract. This error happens when 
```
npx @openzeppelin/upgrades-core@^1.42.0 validate out/build-info --contract test/contracts/xxxV2.sol:xxxV2 --requireReference
```
is executed.
I didn't have this error with @openzeppelin/upgrades-core v1.41.0. 

My v2 initializer:
```
function initializeV2() public reinitializer(2) {}
```

vladimir-trifonov:
> hey [@vladimir-trifonov](https://github.com/vladimir-trifonov) I am getting a similar issue [#1124](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1124), do you remember which version of the `@openzeppelin/hardhat-upgrades` package did `@openzeppelin/upgrades-core v1.41.0` pointed too ?

I was having the same issue as you before, but I removed the 
```
function initialize
```
in the upgraded contract, because either way, I don't need it. But this current issue is braking the upgrade altogether.

viraj124:
hmm I had `initialize` defined before and it was working fine but for some reason now it doesn't since last week or so which is super weird

ericglau:
We added checks to detect issues in parent initializer calls in [@openzeppelin/upgrades-core@1.42.0](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fupgrades-core%401.42.0).  In your scenario, since you are upgrading and if you don't need an initializer in your upgraded contract, you can skip the `missing-initializer` check.

From Foundry:
```
Options memory opts;
opts.unsafeAllow = "missing-initializer";
Upgrades.upgradeProxy(proxy, "MyContractV2.sol", data, opts);
```
From Hardhat:
```
await upgrades.upgradeProxy(proxyAddress, MyContractV2, {
  unsafeAllow: ['missing-initializer'],
});
```

We may consider fixing the plugin to skip this by default when upgrading.

vladimir-trifonov:
I saw this `missing-initializer` check, but I didn't want to use it since I have initializer in my upgraded contract e.g.:

```
[...]

PR #1148, opened 2025-04-10
Support annotation to validate arbitrary functions as initializers

ericglau:
- Adds a section in docs about the initializer validations that are performed by Upgrades Plugins.
- Logs a note when reinitializers are detected, to tell users that reinitializers are not included in validations by default.
- Support a custom NatSpec annotation `@custom:oz-upgrades-validate-as-initializer` to mark arbitrary functions to be treated as initializers for validations.
  - This can be used for any function that is not normally detected as an initializer (for example, custom initializer functions that do not use `onlyInitializing` or `initializer` modifiers), and can also be used on reinitializers to include them in validations.

Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1122
Related to https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1125

### DIFF
--- a/packages/core/src/validate/run/initializer.ts
+++ b/packages/core/src/validate/run/initializer.ts
@@ -331,28 +331,28 @@ function getPossibleInitializers(
 ): FunctionDefinition[] {
   const fns = [...findAll('FunctionDefinition', contractDef)];
   return fns.filter((fnDef: FunctionDefinition) => {
-    const assumeInitializer = hasAssumeInitializerAnnotation(fnDef, decodeSrc);
-    if (!assumeInitializer && fnDef.modifiers.some(modifier => 'reinitializer' === modifier.modifierName.name)) {
+    const validateAsInitializer = hasValidateAsInitializerAnnotation(fnDef, decodeSrc);
+    if (!validateAsInitializer && fnDef.modifiers.some(modifier => 'reinitializer' === modifier.modifierName.name)) {
       logNote(`Reinitializers are not included in validations by default`, [
-        `${decodeSrc(fnDef)}: If you want to validate this function as an initializer, annotate it with '@custom:oz-upgrades-assume-initializer'`,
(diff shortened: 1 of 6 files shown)


packages/core/src/validate/run/initializer.ts
```diff
@@ -318,22 +321,55 @@ function getRecursiveFunctionIds(
...
+    if (annotationArgs.length !== 0) {
+      throw new Error(`${decodeSrc(node)}: @custom:${tag} annotation must not have any arguments`);
+    }
+  }
+  return validateAsInitializer;
+}
+
+/**
+ * Infers whether a function could be an initializer. Does not include private functions.
+ * For parent contracts, only internal and public functions which contain statements are included.
```
ernestognw, line 360: I would expand on why internal functions can be considered a valid initializer. I would expect the initializer is by definition, an external function called into the contract to set it up.

ernestognw (approved):
Not blocking, LGTM

packages/core/src/validate/run/initializer.ts
ericglau: There are two different things that we are referring to as initializers.  For the child contract, this is the external function that should be called to set up.  But for parent contracts, these can be internal functions like `__ERC20_init()` that need to be inherited and called by the child initializer.

I think this new annotation needs some rework to account for this, for example if the annotation is on a private function on a parent contract, that should not be allowed.  Otherwise no child contract can call it.

ernestognw: In Line 369 we're filtering private functions, so those are already not considered, right?

I get the point of running validation in internal functions as in the case o `__ERC20_init()`

ericglau: That logic was previously short circuited on line 341 if the annotation was present.  Changed it to ignore those cases (e.g. private) even if the annotation is present, and added documentation to clarify this behavior.

ernestognw: Right, I see the annotation was preceding those conditions. Solving

PR #1154, opened 2025-05-07
Use changesets for releases, use Node 20

ericglau:
Uses changesets for versioning and publishing.
Updates build environment to Node.js v20.

Equivalent to https://github.com/OpenZeppelin/contracts-wizard/pull/528


CoveMB (approved):

.github/workflows/version.yml
```diff
@@ -0,0 +1,34 @@
...
+          ref: ${{ github.ref }}
+      - name: Set up environment
+        uses: ./.github/actions/setup
+      - name: Create Prepare Release PR
+        uses: changesets/action@001cd79f0a536e733315164543a727bdf2d70aff # v1.5.1
+        with:
+          title: Prepare Release
+          commit: Prepare Release
+          version: npm run version
+          commitMode: github-api
```
tirumerla, line 32: Is this specifically for signing commits or any other advantages over cli?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: changesets/action@001cd79f0a536e733315164543a727bdf2d70aff # v1.5.1
+        with:
+          title: Prepare Release
+          commit: Prepare Release
+          version: npm run version
+          publish: npm run publish
+          commitMode: github-api
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
ernestognw, line 34: Same comment as in the other PR

ernestognw (approved):

.github/workflows/publish.yml
tirumerla, line 34: For future reference making comment to enable config provenance when pushing

tirumerla (approved):
lgtm

.github/workflows/publish.yml
```diff
@@ -0,0 +1,44 @@
...
+concurrency: version-or-publish-${{ github.ref }}
+
+jobs:
+  publish:
+    name: Publish Packages
+    permissions:
+      contents: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    environment: publish
```
tirumerla, line 15: ❤️

.github/workflows/version.yml
ericglau: Yes, according to the Changesets changelog,
> When used with "github-api" value all tags and commits will be attributed to the user whose GITHUB_TOKEN is used, and also signed using GitHub's internal GPG key.

.github/workflows/publish.yml
ericglau: Adding link to reference comment for tracking: https://github.com/OpenZeppelin/contracts-wizard/pull/528#discussion_r2072026221

Issue #1150, opened 2025-04-23
Potential false positive "Missing initializer calls for one or more parent contracts"


heueristik:
I am cross-linking an issue I've already raised in https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades/issues/104:

I think I've handled the initializer calls for my upgradeable ERC20 token contract correctly.

```solidity
contract Token is ERC20Upgradeable, UUPSUpgradeable {
    /// @custom:oz-upgrades-unsafe-allow constructor
    constructor() {
        _disableInitializers();
    }

    function initialize(address initialOwner) external initializer {
        __Token_init(initialOwner);
    }

    function __Token_init(address initialOwner) internal onlyInitializing {
        __Context_init_unchained();
        __ERC20_init_unchained("Token", "TOK");
        __UUPSUpgradeable_init_unchained();

        __Token_init_unchained(initialOwner);
    }

    function __Token_init_unchained(address initialOwner) internal onlyInitializing {
        _mint(initialOwner, 100);
    }

    function _authorizeUpgrade(address newImplementation) internal pure override {
        {
            newImplementation;
        }
        // NOTE: In this MWE, anyone can upgrade. DO NOT USE IN PRODUCTION.
    }
}
```

However, deploying it with `Upgrades.deployUUPSProxy` in the test below

```solidity
contract TokenTest is Test {
    address internal _defaultSender;
    Token internal _tokenProxy;

    function setUp() public {
        (, _defaultSender,) = vm.readCallers();

        // Deploy proxy and mint tokens for the `_defaultSender`.
        vm.prank(_defaultSender);
        _tokenProxy = Token(
            Upgrades.deployUUPSProxy({
                contractName: "Token.sol:Token",
                initializerData: abi.encodeCall(Token.initialize, _defaultSender)
            })
        );
    }

    function test_balance() public view {
        assertEq(_tokenProxy.balanceOf(_defaultSender), 100);
    }
}
```

fails with the following error

```bash
[FAIL: revert: Upgrade safety validation failed:
✘  src/Token.sol:Token

    src/Token.sol:29: Missing initializer calls for one or more parent contracts: `ERC20Upgradeable`
        Call the parent initializers in your initializer function
        https://zpl.in/upgrades/error-001

FAILED] setUp() (gas: 0)
```

I've provided a minimal example to reproduce the error in this repo https://github.com/heueristik/token.
To reproduce the problem, run

```bash
[...]

PR #1161, opened 2025-05-16
Omit missing initializer calls error for initializers named `*_unchained`

ericglau:
This PR omits the "missing initializer calls" error when the child contract's initializer name is `*_unchained`.  These by design should not need to call parent initializers, as described in https://docs.openzeppelin.com/contracts/5.x/upgradeable#multiple-inheritance.

Note we still need to check that other child initializers properly call parent initializers, and some of those parent initializers can also be named `*_unchained`.  Tests have been added for these cases.

Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1150


CoveMB (approved):
Looks good

ernestognw (approved):

PR #1166, opened 2025-06-10
Updates to support contract verification via etherscan V2 API

### LINKED ISSUE #1165: With the updated version of etherscan V1 -> V2 the chainId is missing
When I verify a transparent upgradeable contract and after upgrading to version 2 of Etherscan https://docs.etherscan.io/etherscan-v2, you need to inject the chain id into the query parameters, otherwise you get the following error:

Etherscan returned with message: NOTOK, reason: Missing chainid parameter (required for v2 api), please see https://api.etherscan.io/v2/chainlist for the list of supported chainids

I think modifying line 17 of packages/plugin-hardhat/src/utils/etherscan-api.ts to
`const parameters = { ...params, chainid: etherscan.chainId, apikey: etherscan.apiKey };`
should fix the problem.

So i don't know if you already have a fix somewhere or we have missed something in the configuration

rokso:
This PR adds support for contract verification via etherscan V2 API. 
- `hardhat-verify` dependency is updated to latest 
- `chainid` param is added in etherscan API call.

Fixes #1165

### DIFF
--- a/.changeset/witty-zebras-fold.md
+++ b/.changeset/witty-zebras-fold.md
@@ -2,9 +2,8 @@
 '@openzeppelin/hardhat-upgrades': patch
 ---
 
-Support contract verification via etherscan V2 API. Update peer dependencies to include Pectra support.
+Support contract verification via etherscan V2 API.
 - **Potentially breaking changes**: Changes peer dependencies to the following:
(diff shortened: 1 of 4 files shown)


CoveMB:
Thanks 🌻

CoveMB (approved):

PR #1190, opened 2025-09-17
Support PolkaVM for Hardhat Upgrades Plugin

### LINKED ISSUE #1188: Support upgrades plugin for PolkaVM
Currently `@openzeppelin/hardhat-upgrades` deploys contracts that are shipped with `@openzeppelin/upgrades-core`. They are built with a regular `solc` which is incompatible with PolkaVM. To support PolkaVM for `@openzeppelin/hardhat-upgrades` for PolkaVM, we need:
* Ship a PolkaVM version of contracts (and all needed data like build info) in the adjacent directory, for example `artifacts-pvm` in `@openzeppelin/upgrades-core`
* In `@openzeppelin/hardhat-upgrades` add helpers that will select a correct versions of the contracts based on the field `networks.hardhat.polkavm` being set to `true`. This helpers should be used whenever a contract is deployed.

KitHat:
Support precompiled contracts for PolkaVM deployments.

What should be done to move this PR to ready state:
- [x] https://github.com/paritytech/revive/issues/98 merged and released in `resolc`
- [x] version of Hardhat Polkadot plugin with this compiler released
- [x] [ignored tests](https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/1190/files#diff-cfc36810de882e2fda42a590b9f97042e913425c43aceb7d63c749f0a83c7694R98) moved back in

**NB**: NodeJS version was pushed to 22.x, because Hardhat Polkadot plugin requires built-in WebSocket, that is present only in v22 and higher

Closes #1188


ericglau (commented):
Looking good!

This compiles the core package for both EVM and PVM, but I don't see a way for it to actually run any test cases with the PVM code path.

We should try to find a way to run all of the tests (in both core and plugin-hardhat) using the PVM code path, if feasible, perhaps using a custom GitHub workflow to modify the way it runs.  Even if that is not feasible, we should aim to add at least one mainline test scenario (e.g. deploy and upgrade a proxy) using PVM.

(If that has been tested manually already, I think this would be ok as-is, and the automated tests can be added later)

PR #1192, opened 2025-09-19
Contract Verification - Graceful failure on isBeacon check when using blockscout RPC.

quintessenial:
When verifying a non-beacon contract, the `verify` plugin will check if the `implementation()` call fails to determine if it's not a beacon. Unfortunately with a Blockscout eth-rpc the error message is different than the specified list, causing HH to assume there's something wrong with the RPC.

rpc-error

Adding this additional error message check supports the response coming back from `eth-rpc` found [here](https://github.com/blockscout/blockscout/blob/fab90607cb0c87c1695f0585640de82848b12196/apps/explorer/lib/explorer/eth_rpc.ex#L1304)

### DIFF
--- a/packages/core/src/call-optional-signature.ts
+++ b/packages/core/src/call-optional-signature.ts
@@ -10,6 +10,7 @@ export async function callOptionalSignature(provider: EthereumProvider, address:
       e.message.includes('function selector was not recognized') ||
       e.message.includes('invalid opcode') ||
       e.message.includes('revert') ||
+      e.message.includes('Action not found') ||
       e.message.includes('execution error')
     ) {
       return undefined;


ericglau:
Hi @quintessenial, thanks for the PR and sorry for the late response.  However, this doesn't look like the root cause.  The plugin is calling the `eth_call` RPC method on a function name that possibly does not exist on a contract (and it should catch this case), whereas the "Action not found" error occurs when the RPC method itself does not exist.

For example, if you call `eth_call` on a Blockscout RPC with an address and a non-existent function:
```
curl https://eth.blockscout.com/api/eth-rpc \
  -X POST \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc": "2.0", "method": "eth_call", "params": [{"from": null, "to": "0x6b175474e89094c44da98b954eedeac495271d0f", "data": "0xc2985578"}, "latest"], "id": 1}'
```
the RPC returns:
```
{"jsonrpc":"2.0","error": {"code":-32000,"data":"0x","message":"execution reverted"},"id": 1}
```
which should be caught by the plugin.

However, if you call an invalid RPC method such as "eth_invalid":
```
curl https://eth.blockscout.com/api/eth-rpc \
  -X POST \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc": "2.0", "method": "eth_invalid", "params": [{"from": null, "to": "0x6b175474e89094c44da98b954eedeac495271d0f", "data": "0xc2985578"}, "latest"], "id": 1}'
```
then it returns "Action not found":
```
{"jsonrpc":"2.0","error": "Action not found.","id": 1}
```

If the issue still occurs, can you please provide the following info:
- Which specific RPC URL are you using?  
- Can you use a different RPC provider?  If using Blockscout, is there a different Blockscout RPC URL that you can try?
- Please provide the debug logs and other information described in https://forum.openzeppelin.com/t/required-information-when-requesting-support-for-upgrades-plugins/30530

quintessenial:
Thanks for the response @ericglau 

This error was received using the explorer & RPC over at https://evm.taostats.io/ (RPC: https://evm.taostats.io/api/eth-rpc)

It may not be the root cause, but the only workaround I was able to find was temporarily removing the import for `openzeppelin-upgrades` which then no longer ran the `implementation()` check while verifying a contract.

afa7789:
Hey __@quintessenial__, hope you're doing well!

I reviewed the Elixir link you shared before, so the `{:error, "Action not found."}` error only occurs when an unknown RPC method is called (when `get_action(method)` fails) - not from contract execution failures.

In the case you mentioned, a VM exception is the expected behavior when calling non-existent functions on contracts.
If there were any errors in the validation logic, we would be seeing VM reverts - which is what you get with non-beacon contracts:

```
{"jsonrpc":"2.0","error": {"code":-32603,"data":"0x","message":"VM Exception while processing transaction: revert"},"id": 1}
```

You can make the curl calls, as Eric mentioned, to verify it.

However, I can still debug this a bit more. I need the following:
* The proxy and implementation contract addresses you deployed
* What type of proxy you are using (e.g. beacon or UUPS or transparent)

Note that our plugin only supports source code verification on Etherscan-based explorers.  If you’re using Blockscout, please see this thread on how to verify your proxy and implementation: https://forum.openzeppelin.com/t/how-to-verify-a-contract-on-etherscan-bscscan-polygonscan/14225

If you’d like us to help verify the proxy contract itself, please also provide:
* The proxy address
* Which versions of `@openzeppelin/hardhat-upgrades` and `@openzeppelin/upgrades-core` you used to deploy the proxy

Thanks!

PR #1208, opened 2026-02-02
feat: Add validation for Foundry and Hardhat 3 build-info formats

afa7789:
We are having an issue with this PR https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/1194 CI/CD that depends on a deployed core package; this could be the fix for it.

still investigating, not finished yet.

## Release Notes

* **Bug Fixes**
  * Improved build-info file validation with enhanced error messages for Foundry and Hardhat 3 formats.
  * Added format-specific guidance to direct users to appropriate remediation commands based on their build system.

* **Tests**
  * Added validation tests for missing build-info outputs across different build formats.

### DIFF
--- a/packages/core/src/cli/validate/build-info-file.ts
+++ b/packages/core/src/cli/validate/build-info-file.ts
@@ -151,6 +151,19 @@ async function readBuildInfo(buildInfoFilePaths: string[], dirShortName: string)
   return buildInfoFiles;
 }
 
+const FOUNDRY_BUILD_INFO_HELP = `\
+Foundry build-info files must contain Solidity compiler input, output, and solcVersion.
+Ensure foundry.toml has:
+  [profile.default]
+  build_info = true
+  extra_output = ["storageLayout"]
+Then run: ${FOUNDRY_COMPILE_COMMAND}`;
+
+const HH3_BUILD_INFO_HELP = `\
+Hardhat 3 (HH3) build-info uses a main .json file (input, solcVersion) and a separate .output.json (output).
+Ensure you have compiled with Hardhat 3 so that artifacts/build-info/ contains both the main files and the .output.json files.
+Then run: ${HARDHAT_COMPILE_COMMAND}`;
+
 async function loadBuildInfo(buildInfoFilePath: string): Promise<{
   input: SolcInput;
   output: SolcOutput;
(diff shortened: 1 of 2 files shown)


packages/core/src/cli/validate/build-info-file.ts
```diff
@@ -151,6 +151,19 @@ async function readBuildInfo(buildInfoFilePaths: string[], dirShortName: string)
...
+Ensure foundry.toml has:
+  [profile.default]
+  build_info = true
+  extra_output = ["storageLayout"]
+Then run: ${FOUNDRY_COMPILE_COMMAND}`;
+
+const HH3_BUILD_INFO_HELP = `\
+Hardhat 3 (HH3) build-info uses a main .json file (input, solcVersion) and a separate .output.json (output).
+Ensure you have compiled with Hardhat 3 so that artifacts/build-info/ contains both the main files and the .output.json files.
+Then run: ${HARDHAT_COMPILE_COMMAND}`;
```
ericglau, line 165: This sounds like it's telling the user to compile twice which is confusing.
Perhaps combine them:

```suggestion
Ensure you compile with Hardhat 3 so that artifacts/build-info/ contains both .json files and .output.json files: ${HARDHAT_COMPILE_COMMAND}`;
```

ericglau: Is there anything that the user needs to do if build-info does not have both types of files?  Or does that perhaps suggest they are using Hardhat 2 (which is not supported with this version)?

packages/core/src/cli/validate/build-info-file.ts
```diff
@@ -151,6 +151,19 @@ async function readBuildInfo(buildInfoFilePaths: string[], dirShortName: string)
...
+const FOUNDRY_BUILD_INFO_HELP = `\
+Foundry build-info files must contain Solidity compiler input, output, and solcVersion.
+Ensure foundry.toml has:
+  [profile.default]
+  build_info = true
+  extra_output = ["storageLayout"]
+Then run: ${FOUNDRY_COMPILE_COMMAND}`;
+
+const HH3_BUILD_INFO_HELP = `\
+Hardhat 3 (HH3) build-info uses a main .json file (input, solcVersion) and a separate .output.json (output).
```
ericglau, line 163: Let's not call it HH3 in messages (the acronym is not widely used).
```suggestion
Hardhat 3's build-info uses a main .json file (input, solcVersion) and a separate .output.json (output).
```

ericglau:
Closing since this will be part of https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/1194 instead.

PR #1250, opened 2026-05-28
Publish via npm OIDC trusted publishing, avoid submodule checkout in changeset workflows

ericglau:
## Summary

Hardens the release workflows and switches publishing to npm OIDC trusted
publishing. Originally scoped to dropping the recursive submodule checkout;
now also covers the OIDC migration and workflow-hardening review feedback.

### Don't checkout submodules in release workflows (`version.yml`, `publish.yml`, `changeset.yml`)
These jobs only run changeset version/publish/check and build the TypeScript
packages — they don't need submodules. The recursive checkout was a temporary
workaround for a `yarn install` crash back when `package.json` referenced
foundry-upgrades via the submodule; it now uses the published registry version.
It was also breaking releases: `commitMode: github-api` rejects executable files
anywhere in the working tree, and the nested forge-std submodule ships one
(`scripts/vm.py`). `checks.yml` keeps its submodule checkout (Solidity-test
integration + docs generation) and is untouched.

### OIDC trusted publishing (`publish.yml`)
Publishes authenticate via OIDC for existing packages, with an npm token used
only as a fallback for new packages; provenance enabled. The publish job runs
on Node 24 (required for OIDC / npm ≥ 11.5.1).

### Workflow hardening (review feedback)
- Disable dependency and compiler caching in the publish job.
- Pin `setup-node` and `cache` to commit SHAs instead of mutable `v4` tags.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -68,9 +68,18 @@ jobs:
       - name: Configure npm auth
         run: |
           if [ "$HAS_NEW_PACKAGES" = "true" ]; then
+            # New packages aren't registered for OIDC trusted publishing yet, so
+            # authenticate with the npm token. setup-node's .npmrc reads it from
+            # NODE_AUTH_TOKEN.
             echo "NODE_AUTH_TOKEN=${NPM_TOKEN}" >> "$GITHUB_ENV"
           else
-            echo "NODE_AUTH_TOKEN=" >> "$GITHUB_ENV"
+            # Existing packages publish via OIDC trusted publishing. npm only
+            # falls through to OIDC when no _authToken is configured, so strip the
+            # line setup-node wrote to the npm userconfig. Removing it is enough:
+            # nothing downstream then reads NODE_AUTH_TOKEN.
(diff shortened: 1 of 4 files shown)


ericglau:
@SocketSecurity ignore-all
Looks like false positives, and unrelated to this PR.

.github/workflows/publish.yml
```diff
@@ -18,7 +18,9 @@ jobs:
         with:
           fetch-depth: 0 # To get all tags
           ref: ${{ github.ref }}
-          submodules: recursive
+          # Submodules are intentionally not checked out: publishing only builds
+          # the TypeScript packages, and commitMode: github-api rejects executable
+          # files anywhere in the working tree (e.g. forge-std scripts).
       - name: Set up environment
         uses: ./.github/actions/setup
```
tirumerla, line 26: Lets disable cache for publish workflows and also pin the actions to specific commit sha

tirumerla (changes requested):

tirumerla (approved):
lgtm

PR #1253, opened 2026-06-01
Support Solidity 0.8.35 erc7201 comptime builtin and fix storage location calc

ericglau:
## Summary

- Support the Solidity 0.8.35 `erc7201` comptime builtin as a `layout at` base slot.
  - Accept both `"..."` and `unicode"..."` string literals as the argument.
  - Extend the clash check between a custom base slot and ERC-7201 namespaces to handle `layout at erc7201("<id>")` (previously only numeric-literal base slots were resolved).
  - When the base slot is an `erc7201(...)` builtin call, suppress the warning about possible overlap between the custom layout and namespaced storage. The base slot is itself computed via the ERC-7201 formula, so it relies on the same collision resistance that namespaces rely on against each other; the tool likewise does not warn when multiple namespaces coexist. An exact clash is still reported as an error.
- Fix a pre-existing bug in `calculateERC7201StorageLocation`: the intermediate value `keccak256(id) - 1` was converted to minimal hex via `toString(16)` and passed to `Buffer.from(hex, 'hex')`, which silently truncated when the value had a leading zero nibble or byte — so fewer than 32 bytes were hashed in the second step. Zero-padded to 32 bytes to match `abi.encode(uint256)`. Only affected the plugin's own clash detection; on-chain slots are computed by Solidity and were unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added support for Solidity 0.8.35 `erc7201` comptime builtin when used as a `layout at` base slot.
  * Enhanced storage layout validation to detect namespace/base-slot clashes and suppress warnings for compatible configurations.

* **Bug Fixes**
  * Fixed ERC-7201 storage location calculation when intermediate values include leading zero nibbles or bytes.

* **Chores**
  * Updated Hardhat dependency version.


ericglau:
@SocketSecurity ignore-all
Unrelated to PR.

packages/core/src/utils/erc7201-builtin.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const id = getErc7201BuiltinNamespaceId(node);
+  if (id !== undefined) {
+    return calculateERC7201StorageLocation(id);
+  }
+
+  return undefined;
+}
+
+function getStringLiteralValue(literal: { value?: string | null }): string {
```
ernestognw: Nit: `getStringLiteralValue` takes `{ value?: string | null }` rather than `Literal`. The only caller is the `arg.kind === 'string' | 'unicodeString'` branch above, so tightening to `Literal` (or even the narrower string-literal variant) would make the contract self-documenting without losing anything. Not a blocker.

ernestognw (approved):
Non-blocking

